PR #32, opened 2025-12-04
[Spec 0013] Document OS dependencies and add codev-doctor

waleedkadous:
## Summary

- Add comprehensive Prerequisites section to README.md with dependency tables for core deps (Node.js, tmux, ttyd, git, Python) and AI CLIs
- Create `agent-farm/src/utils/deps.ts` for programmatic dependency checking with version validation
- Update `af start` to check all core dependencies (not just ttyd) with helpful error messages
- Add `codev/bin/codev-doctor` shell script for full environment verification
- Update INSTALL.md to use `codev-doctor` as primary verification step

## Test plan

- [x] Run `./codev/bin/codev-doctor` - verifies all dependencies with versions
- [x] Build agent-farm TypeScript (`npm run build`) - compiles without errors
- [x] Verify deps.ts is properly exported from utils/index.ts
- [x] Check version comparison logic handles edge cases (e.g., "3.3a" -> "3.3")

### DIFF
--- a/agent-farm/src/utils/deps.ts
+++ b/agent-farm/src/utils/deps.ts
@@ -113,6 +113,7 @@ export interface DependencyCheckResult {
   version: string | null;
   minVersion: string;
   versionOk: boolean;
+  versionUnknown: boolean;
   installHint: string;
 }
 
@@ -131,6 +132,7 @@ export async function checkDependency(dep: Dependency): Promise<DependencyCheckR
       version: null,
       minVersion: dep.minVersion,
       versionOk: false,
+      versionUnknown: false,
       installHint,
     };
   }
@@ -138,6 +140,7 @@ export async function checkDependency(dep: Dependency): Promise<DependencyCheckR
   // Get version
   let version: string | null = null;
   let versionOk = false;
+  let versionUnknown = false;
 
   try {
     const { stdout, stderr } = await run(dep.versionCmd);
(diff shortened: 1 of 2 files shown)


waleedkadous:
## 3-Agent Review (Claude, Gemini, Codex)

### Overall: APPROVE with revisions

The implementation is solid and comprehensive. Address these issues before merge:

### High Priority
**Version warnings should be errors for Node**
- `codev-doctor` exits 0 when Node is below 18.0 (just warns)
- But `agent-farm` declares `engines.node: >=18.0.0` and won't run
- Users get false sense of readiness
- **Fix**: Change from `warn` to `fail` for Node version check (and arguably tmux/ttyd too)

### Medium Priority  
**Unparseable versions default to OK**
- In `deps.ts`, if `versionParser` returns `null`, `versionOk` is forced to `true`
- Could accept too-old versions silently (e.g., `ttyd version 1.6` without patch)
- **Fix**: Consider treating parse failures as warnings instead of silent pass

### Low Priority
**OS-specific hints inconsistent**
- `codev-doctor` mixes `brew` (macOS) and `apt` (Linux) hints
- git says `apt install git` even on macOS
- **Fix**: Detect `OSTYPE` or provide both hints

### Accepted Tradeoffs (no action needed)
- Duplication between `deps.ts` and `codev-doctor` is OK (different purposes: runtime vs bootstrap)
- Python not in `CORE_DEPENDENCIES` is OK (`deps.ts` is for agent-farm runtime, Python is for consult)

PR #174, opened 2026-01-28
[Bugfix #170] Fix shell tabs not starting when clicked during dashboard load

### LINKED ISSUE #170: Shells don't start when clicked on in dashboard
## Description

Clicking on shell entries in the Agent Farm dashboard does not start/connect to the shell.

## Expected Behavior

Clicking a shell entry should start or connect to that shell session.

## Actual Behavior

Nothing happens when clicking on shell entries.

## Steps to Reproduce

1. Start Agent Farm dashboard (`af dash start`)
2. Click on a shell entry
3. Observe that the shell doesn't start

## Environment

- Codev version: 2.0.0-rc.29
- Platform: macOS/Linux

waleedkadous:
## Summary
Fixes #170

Clicking on shell entries in the dashboard was sometimes unresponsive. The shell tab would appear to be selected but nothing would happen.

## Root Cause

`renderDashboardTab()` is an async function that loads the projectlist and files tree. It's called without `await` from `renderTabContent()`, so it runs in the background.

If the user clicks on a shell tab while the dashboard is still loading (during the `await Promise.all(...)`), this race condition occurs:

1. User clicks shell → `selectTab('shell-X')` runs synchronously
2. Shell iframe is rendered in `#tab-content`
3. Meanwhile, `renderDashboardTab()` finishes awaiting
4. `renderDashboardTabContent()` overwrites the shell iframe with dashboard HTML
5. User sees dashboard instead of shell → "nothing happens"

## Fix

Added a guard check after the await in `renderDashboardTab()` to verify the dashboard is still the active tab before rendering its content. If the user switched tabs during loading, the function returns early.

```javascript
await Promise.all([...]);

// Guard against race condition
if (activeTabId !== 'dashboard') {
  return;
}

renderDashboardTabContent();
```

## Test Plan
- [x] Verified fix locally
- [ ] CMAP review (to be added)

### Manual Testing Steps
1. Start Agent Farm dashboard with at least one shell tab
2. Navigate to the Dashboard tab
3. Quickly click on a shell tab while the dashboard is loading
4. Verify the shell terminal loads correctly (not overwritten by dashboard)

### DIFF
--- a/packages/codev/src/agent-farm/__tests__/dashboard-race.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/dashboard-race.test.ts
@@ -1,0 +1,253 @@
+/**
+ * Tests for dashboard race condition fix (Bugfix #170)
+ *
+ * Verifies that renderDashboardTab() doesn't overwrite tab content when
+ * the user switches tabs while the dashboard is still loading.
+ *
+ * The race condition:
+ * 1. User views dashboard -> renderDashboardTab() starts async loading
+ * 2. User clicks shell tab -> selectTab() renders shell iframe
+ * 3. Dashboard loading completes -> would overwrite shell if not guarded
+ *
+ * The fix adds a guard check after the await to verify dashboard is still active.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+
+/**
+ * Simulates the dashboard's renderDashboardTab function behavior.
+ * This is a simplified version that captures the essential race condition pattern.
+ */
+async function simulateRenderDashboardTab(
+  state: {
+    activeTabId: string;
+    contentRendered: string | null;
+  },
+  loadProjectlist: () => Promise<void>,
+  loadFilesTree: () => Promise<void>,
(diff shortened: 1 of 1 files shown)


waleedkadous:
Added regression test for the race condition fix.

**Test file**: `packages/codev/src/agent-farm/__tests__/dashboard-race.test.ts`

**Test coverage** (6 tests):
1. **Without guard (demonstrates the bug)**: Verifies that without the guard, dashboard overwrites shell content
2. **With guard - shell preserved**: Verifies the fix prevents overwriting when user switches to shell
3. **With guard - normal operation**: Verifies dashboard still renders when user stays on it
4. **With guard - rapid switching**: Verifies fix handles dashboard→shell→builder switching
5. **With guard - switch back**: Verifies fix allows dashboard render when user returns to it
6. **Async loading behavior**: Verifies both projectlist and files tree load before rendering

All 6 tests pass.

waleedkadous:
Closing — the dashboard was rewritten as a React app in Spec 0090. The old templates/dashboard/js/main.js this PR targets is no longer used.

PR #567, opened 2026-02-26
[Bugfix #565] Fix Tower crash: guard H2 stream.respond() against destroyed streams

### LINKED ISSUE #565: Tower crash: uncaught ERR_HTTP2_INVALID_STREAM in tunnel proxy
## Bug

Tower crashed with an uncaught exception at 07:28 UTC on 2026-02-26:

```
[ERROR] Uncaught exception: The stream has been destroyed
Error [ERR_HTTP2_INVALID_STREAM]: The stream has been destroyed
    at ServerHttp2Stream.respond (node:internal/http2/core:2960:13)
    at ClientRequest.<anonymous> (tunnel-client.js:483:20)
```

## Root Cause

In `tunnel-client.ts`, `proxyHttpRequest()` line 625 calls `stream.respond()` without checking `stream.destroyed` first. When the tunnel WebSocket disconnects while a proxy request is in-flight, the H2 stream is destroyed, but the local HTTP response callback still fires and tries to respond on the dead stream.

The error callback at line 634 and the WebSocket response handler at line 574 both have `stream.destroyed` guards, but the **main success callback at line 625 does not**.

## Timeline

1. Remote browser sends HTTP request through cloud tunnel
2. Tunnel proxies to localhost:4100 via `proxyHttpRequest()`
3. Tunnel WebSocket disconnects (network blip) → H2 stream destroyed
4. Local server responds → callback calls `stream.respond()` on destroyed stream
5. Uncaught `ERR_HTTP2_INVALID_STREAM` → Tower process crash

## Fix

Add `if (stream.destroyed) return;` before `stream.respond(responseHeaders)` at line 625, matching the guard pattern already used at lines 574 and 581.

Also audit all other `stream.respond()` call sites in `handleH2Stream()` (lines 493, 503) — these should also be guarded since the stream could be destroyed between the H2 'stream' event and the synchronous respond call if the WebSocket drops.

## Files

- `packages/codev/src/agent-farm/lib/tunnel-client.ts` — lines 493, 503, 625

waleedkadous:
## Summary
Fixes #565

## Root Cause
In `tunnel-client.ts`, `proxyHttpRequest()` calls `stream.respond()` without checking `stream.destroyed` first. When the tunnel WebSocket disconnects while a proxy request is in-flight, the H2 stream is destroyed but the local HTTP response callback still fires and tries to respond on the dead stream, causing an uncaught `ERR_HTTP2_INVALID_STREAM` that crashes Tower.

The error callback and WebSocket response handler already had `stream.destroyed` guards, but the **main success callback did not**.

## Fix
Added `if (stream.destroyed) return;` guards before every `stream.respond()` call in the tunnel proxy:

1. **`handleH2Stream` blocklist response** (line 493) — guard before 403 response
2. **`handleH2Stream` metadata response** (line 504) — guard before 200 response
3. **`handleWebSocketConnect` upgrade handler** (line 557) — guard before 200 response, also destroys the local socket to prevent leaks
4. **`proxyHttpRequest` success callback** (line 628) — guard before forwarding response, calls `proxyRes.resume()` to drain

Net diff: 4 lines of guards + 156 lines of tests.

## Test Plan
- [x] Added regression test (`bugfix-565-h2-stream-destroyed.test.ts`) with 4 test cases
- [x] Verified fix locally — destroyed streams are silently skipped
- [x] TypeScript compiles clean
- [x] Regression tests pass

## CMAP Review
Pending

### DIFF
--- a/packages/codev/src/agent-farm/__tests__/bugfix-565-h2-stream-destroyed.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/bugfix-565-h2-stream-destroyed.test.ts
@@ -1,0 +1,156 @@
+/**
+ * Regression test for Bugfix #565: Tower crash on ERR_HTTP2_INVALID_STREAM
+ *
+ * When the tunnel WebSocket disconnects while a proxy request is in-flight,
+ * the H2 stream is destroyed but the local HTTP response callback still fires,
+ * calling stream.respond() on the dead stream. This must not throw.
+ *
+ * The fix adds `if (stream.destroyed) return;` guards before every
+ * stream.respond() call in handleH2Stream, handleWebSocketConnect, and
+ * proxyHttpRequest.
+ */
+
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import { EventEmitter } from 'node:events';
+import http from 'node:http';
+import { Readable } from 'node:stream';
+import { TunnelClient } from '../lib/tunnel-client.js';
+
+/** Minimal mock of ServerHttp2Stream with destroyed=true */
+function createDestroyedStream() {
+  const stream = new EventEmitter() as any;
+  stream.destroyed = true;
+  stream.respond = vi.fn(() => {
+    throw new Error('ERR_HTTP2_INVALID_STREAM: The stream has been destroyed');
+  });
+  stream.end = vi.fn();
+  stream.write = vi.fn();
+  stream.destroy = vi.fn();
+  stream.pipe = vi.fn(() => stream);
+  return stream;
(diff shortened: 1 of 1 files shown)


waleedkadous:
## Architect Review

Guards added to all 4 unguarded `stream.respond()` calls in tunnel-client.ts (lines 493, 503, 555, 625). The `proxyRes.resume()` drain on the dead-stream path is correct — prevents backpressure buildup. 4 regression tests covering destroyed and live stream paths.

Approved.

---
Architect review

PR #578, opened 2026-02-27
[Bugfix #577] Fix: Parse architect command flags via shell execution

### LINKED ISSUE #577: af architect fails with command strings that include flags
## Bug

Running `af architect` fails with:

```
[error] Architect command not found: claude --dangerously-skip-permissions. Check af-config.json or install claude.
```

Even though `claude` is installed and in PATH.

## Root Cause

`architect.ts` passes the full config string (e.g., `"claude --dangerously-skip-permissions"`) directly to Node's `spawn()` without `shell: true`. Node treats the entire string as a literal executable name (including spaces and flags), causing ENOENT.

The builder path doesn't have this issue because it wraps commands in a bash script before spawning.

## Repro

1. Set `af-config.json`:
   ```json
   { "shell": { "architect": "claude --dangerously-skip-permissions" } }
   ```
2. Run `af architect`
3. Observe ENOENT error

## Fix

Add `shell: true` to the spawn options in `architect.ts` so the command string is parsed by `/bin/sh`.

amrmelsayed:
## Summary

- `af architect` fails with ENOENT when `af-config.json` contains command flags (e.g., `"claude --dangerously-skip-permissions"`)
- Node's `spawn()` without `shell: true` treats the full string as a literal binary name
- Added `shell: true` to spawn options so the command is parsed by `/bin/sh`
- Updated test assertions to match

Fixes #577

### DIFF
--- a/packages/codev/src/agent-farm/__tests__/af-architect.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/af-architect.test.ts
@@ -58,7 +58,7 @@ describe('af architect command', () => {
     expect(mockSpawn).toHaveBeenCalledWith(
       'claude',
       ['--append-system-prompt', '# Architect Role\n\nYou are an architect.'],
-      { stdio: 'inherit', cwd: '/test/workspace' }
+      { stdio: 'inherit', cwd: '/test/workspace', shell: true }
     );
   });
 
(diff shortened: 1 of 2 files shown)


waleedkadous (approved):
Thank you very much for this! Good catch

PR #645, opened 2026-03-28
fix: default fava-trails artifact scope to codev-artifacts/<Org>/<Repo>

timeleft--:
## Summary

- When `artifacts.backend` is `"fava-trails"` (or `"cli"` with default command) and no explicit `artifacts.scope` is configured in `.codev/config.json`, auto-derive the scope as `codev-artifacts/<Org>/<Repo>` from the git remote origin URL
- If an explicit scope IS configured, use it as-is (no prefix, no override)
- Adds `deriveDefaultScope()` helper that parses both HTTPS and SSH remote URLs

## Context

Part of fava-trails Spec 26 — scope convention for storing codev artifacts under `codev-artifacts/<Org>/<Repo>/specs|plans|reviews/<N-name>/`. The fava-trails side is being implemented separately. This change makes the default ergonomic so users don't have to manually configure the scope.

## Test plan

- [x] Existing artifacts tests pass (23/23)
- [x] New test: auto-derives scope from git remote when scope is not configured
- [x] New test: throws with updated error message when no scope and no git remote
- [x] Build passes

### DIFF
--- a/packages/codev/src/commands/porch/__tests__/artifacts.test.ts
+++ b/packages/codev/src/commands/porch/__tests__/artifacts.test.ts
@@ -223,9 +223,7 @@ describe('getResolver', () => {
     }));
     const resolver = getResolver(tmpDir);
     expect(resolver).toBeInstanceOf(CliResolver);
-    // Verify the scope was derived correctly by checking the resolver's internal state
-    // We can do this by checking that listChildren uses the expected scope prefix
-    // For now, just confirm it didn't throw and returned a CliResolver
+    expect((resolver as any).scope).toBe('codev-artifacts/TestOrg/test-repo');
   });
 
   it('throws for unknown backend', () => {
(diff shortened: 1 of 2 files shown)


waleedkadous:
Thanks for the PR! I need to decline this one though.

It's important that codev remains independent of fava-trails (or any specific artifact storage service). Codev is a general-purpose development framework, and baking in assumptions about a particular backend's conventions (like the `codev-artifacts/<Org>/<Repo>` scope pattern) creates coupling that we want to avoid.

Looking at this more broadly, I realize we also need to revisit what landed in #636 — the `fava-trails` default command and alias shouldn't have been merged as-is. The CLI artifact resolver should be fully generic: no default command (require explicit `artifacts.command` in config), and no product-specific aliases. We'll open a follow-up to clean that up.

The artifact resolver abstraction itself is great — we just need it to be backend-agnostic. If someone wants to use fava-trails, they configure `"command": "fava-trails"` explicitly in their `.codev/config.json`. Codev shouldn't assume or prefer any particular tool.

PR #657, opened 2026-04-05
[Bugfix #584] Pace multi-line afx send messages to prevent paste detection

### LINKED ISSUE #584: af send: multi-line messages (>3 lines) treated as paste, final Enter swallowed
## Bug

When `af send` delivers a message longer than 3 lines to a terminal, the terminal (likely xterm.js or the shell) classifies it as a **paste** rather than typed input. This causes the final Enter/newline to be swallowed — the message appears in the terminal but doesn't get submitted.

## Steps to Reproduce

1. `af send 0042 "a message that spans more than three lines"`
2. Observe that the message appears in the builder's terminal but the trailing Enter doesn't fire
3. Short messages (1-3 lines) work fine

## Expected Behavior

The message should be fully delivered AND submitted (Enter processed), regardless of length.

## Suggested Fix

Pace the output for messages longer than 3 lines — introduce a small delay between chunks or between the message body and the final newline, so the terminal doesn't classify it as a bracketed paste.

waleedkadous:
## Summary
Fixes #584

## Root Cause
When `afx send` delivers a message longer than 3 lines, the entire formatted text is written to the PTY in a single `session.write()` call. The receiving terminal (Claude Code) classifies the rapid multi-line input as a "paste" operation, which causes the final `\r` (Enter) — sent 50ms later — to be swallowed. The message appears on screen but never gets submitted.

## Fix
Extracted a `writeMessageToSession()` helper in `message-write.ts` that paces multi-line output:
- **≤3 lines**: Single `write()` + 50ms delayed Enter (existing behavior, works fine)
- **>3 lines**: Line-by-line writes with 10ms gaps between lines, then Enter after an 80ms gap from the last line

All three delivery paths now use this helper:
1. Immediate delivery in `handleSend()` (tower-routes.ts)
2. Deferred delivery via `deliverBufferedMessage()` (tower-routes.ts)
3. Cron message delivery in `tower-cron.ts`

The helper also supports a `delayOffset` parameter and returns the end time, enabling `SendBuffer.flush()` to serialize multiple buffered messages without interleaving.

## Test Plan
- [x] Added 9 regression tests covering short/long messages, noEnter, delayOffset, and multi-message serialization
- [x] Existing send-buffer tests pass (13/13)
- [x] Full test suite passes
- [x] TypeScript compiles cleanly

## CMAP Review
- **Claude**: APPROVE — clean implementation, correct timing math, good test coverage
- **Codex**: REQUEST_CHANGES (addressed) — identified interleaving risk when multiple buffered messages flush to the same session. Fixed by adding `delayOffset` parameter and return value to serialize writes.
- **Gemini**: REQUEST_CHANGES (addressed) — identified same vulnerable pattern in `tower-cron.ts` and suggested extracting to shared module. Fixed by creating `message-write.ts` and updating tower-cron.ts.

### DIFF
--- a/packages/codev/src/agent-farm/__tests__/bugfix-584-send-multiline-pacing.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/bugfix-584-send-multiline-pacing.test.ts
@@ -1,0 +1,116 @@
+/**
+ * Regression test for Bugfix #584: afx send multi-line messages (>3 lines)
+ * treated as paste, final Enter swallowed.
+ *
+ * Verifies that writeMessageToSession paces multi-line output line-by-line
+ * with delays to prevent paste detection, while short messages are still
+ * written in a single call.
+ */
+
+import { describe, it, expect, vi, beforeEach, afterEach } from 'vitest';
+import { writeMessageToSession } from '../servers/tower-routes.js';
+import type { PtySession } from '../../terminal/pty-session.js';
+
+function makeSession(): PtySession & { writeCalls: string[] } {
+  const writeCalls: string[] = [];
+  return {
+    write: vi.fn((data: string) => writeCalls.push(data)),
+    writeCalls,
+  } as unknown as PtySession & { writeCalls: string[] };
+}
+
+describe('writeMessageToSession (Bugfix #584)', () => {
+  beforeEach(() => {
+    vi.useFakeTimers();
+  });
+
+  afterEach(() => {
+    vi.useRealTimers();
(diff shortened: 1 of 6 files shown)


waleedkadous:
## Architect Review

Clean fix. Good extraction of `writeMessageToSession` into a shared module — consolidates three separate write sites into one. The line-by-line pacing for >3 lines with serialized offsets to prevent interleaving is well thought through.

Go ahead and merge.

PR #678, opened 2026-04-17
[Bugfix #677] Allow --force with --protocol alone

### LINKED ISSUE #677: afx spawn --protocol maintain should work without an issue number
## Problem

`afx spawn --protocol maintain` requires a clean worktree. When the worktree has uncommitted local state files (`.codev/config.json`, `.mcp.json`, etc.), it fails with "Uncommitted changes detected."

Using `--force` to skip the check fails with "—force requires an issue number (not needed for --task)". This forces the architect to either:
1. Create a GitHub issue just to get an issue number (unnecessary ceremony for maintenance)
2. Fall back to `afx spawn --task "..."` which gives a random ID like `task-NH8C` instead of a clean name

## Expected Behavior

`afx spawn --protocol maintain --force` should work without an issue number. Maintenance runs are operational, not feature work — they don't naturally have issues.

Ideally `afx spawn --protocol maintain` should also generate a sensible builder name like `builder-maintain-NNNN` (where NNNN is the next maintenance run number) instead of requiring an issue-derived name.

## Current Workaround

Use `afx spawn --task "Run MAINTAIN protocol" --force`, which works but produces an opaque task ID.

waleedkadous:
## Summary
Fixes #677

`afx spawn --protocol maintain --force` now works without an issue number. Maintenance runs are operational and don't naturally have GitHub issues.

## Root Cause
`validateSpawnOptions` in `spawn.ts` rejected `--force` unless `--issue` or `--task` was present. When the main worktree had uncommitted local state and a user tried `afx spawn --protocol maintain --force` to bypass the dirty check, validation failed with `--force requires an issue number (not needed for --task)`.

## Fix
Relaxed the `--force` validation to also allow `--protocol`:

```ts
if (options.force && !options.issueNumber && !options.task && !options.protocol) {
  fatal('--force requires an issue number, --task, or --protocol');
}
```

The existing dirty-worktree check at `spawn()` already honors `--force` correctly, so no further changes were needed.

## Test Plan
- [x] Added regression test: `--protocol maintain + --force` is accepted
- [x] Updated the opposing test to target a still-invalid combo (`--shell + --force`)
- [x] Full `agent-farm` suite passes (1434 passing, 13 pre-existing skipped)

## CMAP Review
Not run — this is a 2-line validation change with a clear regression test.

### DIFF
--- a/packages/codev/src/agent-farm/__tests__/spawn.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/spawn.test.ts
@@ -44,8 +44,8 @@ function validateSpawnOptions(options: SpawnOptions): string | null {
     return '--no-comment requires an issue number';
   }
 
-  if (options.force && !options.issueNumber && !options.task) {
-    return '--force requires an issue number (not needed for --task)';
+  if (options.force && !options.issueNumber && !options.task && !options.protocol) {
+    return '--force requires an issue number, --task, or --protocol';
   }
 
   // --protocol cannot be used with --shell or --worktree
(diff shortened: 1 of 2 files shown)


waleedkadous:
Architect review: Approved. 2-line fix with regression test, CI green.

PR #681, opened 2026-04-18
[Bugfix #680] Prevent V8 heap exhaustion in gemini consult

waleedkadous:
## Summary

Fixes #680

`consult -m gemini` was crashing on PR diffs >500KB due to V8 old-space
exhaustion in the spawned `gemini-cli` subprocess. Truncated stdout caused
`JSON.parse` to fail, and CMAP silently degraded to 2-way exactly when 3-way
signal mattered most (shannon PRs #642, #650, #568, #629).

## Root Cause

`gemini-cli` v0.37.x runs out of its default V8 old-space (~1.7 GB) during
non-interactive mode:
- `loadServerHierarchicalMemory` / JIT context walk (graceful-fs wrapping
  `fs.readdir`) produces the `ArrayUnshift` + `AfterScanDir` frames
- Large prompt passed via argv, which V8 buffers in both argv and prompt paths
- `advanced.autoConfigureMemory` exists upstream but defaults OFF

Only gemini was affected because only gemini is shelled out and only gemini
parses JSON from a buffered child stdout. Codex and Claude run in-process.

## Fix

`packages/codev/src/commands/consult/index.ts` — two complementary mitigations:

1. **Bump heap via NODE_OPTIONS** (`--max-old-space-size=8192`). Survives
   gemini-cli's internal relaunch. Preserves any caller-supplied NODE_OPTIONS.
2. **Pipe the prompt via stdin** instead of argv. Avoids ARG_MAX and keeps V8
   from holding the prompt buffer twice. EPIPE from an early child exit is
   handled gracefully so we surface the underlying error, not the pipe close.

No changes to codex or claude paths (unaffected).

## Test Plan

- [x] Added 4 regression tests (`Gemini large-prompt crash mitigation (Bugfix #680)`)
  - NODE_OPTIONS contains `--max-old-space-size=8192`
  - Query is NOT passed as positional argv
  - `stdio[0]` is `'pipe'` (not `'ignore'`)
  - Caller's existing NODE_OPTIONS is preserved and appended to
- [x] Full unit suite: 41 passed (previously 37) + 2 pre-existing failures
  unrelated to this change (skeleton fallback — fail on `main` as well)
- [x] End-to-end manual test: `consult -m gemini --prompt "What is 1+1?"`
  returns `2` in ~4s with the new stdin path

## Diff Size

```
 2 files changed, 168 insertions(+), 4 deletions(-)
```

Net diff: 172 LOC (under 300).

## CMAP Review

[...]

Issue #680, opened 2026-04-18
consult: gemini crashes on large PR diffs (V8 heap exhaustion)


waleedkadous:
## Summary

`consult` gemini branch crashes on PRs with diffs >500KB. V8 heap exhaustion in the spawned `gemini-cli` subprocess truncates stdout mid-stream, causing `JSON.parse` to fail. Codex and Claude are unaffected because they use in-process SDKs.

Reported by the shannon-workspace architect after losing gemini signal on multiple critical PR reviews (shannon PRs #642, #650, #568 phases 5-7, #629 Phase B). CMAP degraded to 2-way exactly when 3-way was most needed.

## Symptoms

4/4 failure rate on large PRs, two failure modes:
- **(a)** `Failed to extract usage: Unexpected end of JSON input` + exit 1 after ~20s. V8 heap stack traces: `AfterScanDir`, `uv__async_io`, `ArrayUnshift`
- **(b)** exit 13 after 0.4s (same fragile JSON path, known gemini-cli issue)

## Root Cause

`gemini-cli` v0.37.1 runs out of V8 old-space (~1.7 GB default) in the subprocess `consult` spawns. Contributing factors:

1. `gemini-cli` eagerly runs `loadServerHierarchicalMemory` / JIT-context directory walk (graceful-fs wrapping `fs.readdir` — produces the `ArrayUnshift` + `AfterScanDir` frames)
2. Large prompt passed as argv
3. `advanced.autoConfigureMemory` is available in gemini-cli for this case but defaults OFF

Only gemini is affected because only gemini is shelled out and only gemini parses JSON from a buffered child stdout. Codex and Claude run in-process within consult's own Node process.

## Code Location

`packages/codev/src/commands/consult/index.ts`:
- Line 622: prompt passed as argv: `cmd = [config.cli, '--output-format', 'json', ...config.args, query];`
- Line 652: `spawn(cmd[0], cmd.slice(1), { cwd, env, stdio: ['ignore', 'pipe', 'inherit'] })`
- No `NODE_OPTIONS` set on spawn env

## Proposed Fix

Ideally both:

**Option A — bump heap via NODE_OPTIONS** (survives gemini's internal relaunch):
```ts
env.NODE_OPTIONS = [process.env.NODE_OPTIONS ?? '', '--max-old-space-size=8192'].join(' ').trim()
```

**Option B — pipe prompt via stdin instead of argv**:
[...]

PR #686, opened 2026-04-20
[Bugfix #684] Write PR diff to temp file instead of inlining

### LINKED ISSUE #684: Gemini consult fails on large PR diffs due to inlined payload size
> **Source**: Analysis from the Shannon project's architect after 5 consecutive failures of \`consult -m gemini --type integration pr 796\` (PR #796, ~831KB diff, 16,846 lines). Claude and Codex succeeded on the same PR. Received 2026-04-20.

## Symptom

\`\`\`
[warn] Failed to extract usage for gemini: Unexpected end of JSON input
[gemini completed in 0.3s]
Process exited with code 13
\`\`\`

Same host, same PR: Codex + Claude succeed; Gemini fails in 0.3s. A general-mode Gemini query (\`"Respond with one word: ok"\`) completes fine — failure is payload-shape specific.

## Context: what #680 fixed, what it didn't

PR #681 (bugfix #680, shipped in v3.0.0-rc.10) moved the Gemini prompt from argv to stdin and bumped \`NODE_OPTIONS=--max-old-space-size=8192\`. That fixes the **transport** for prompts that fit in memory. It does **not** reduce the payload.

Shannon's failure surfaces that the payload itself is still the problem: 831KB of diff inlined into a single prompt blows past what gemini-cli's headless JSON path survives.

## Root cause

\`buildPRQuery\` (\`packages/codev/src/commands/consult/index.ts:833\`) inlines the full \`git diff\` verbatim into the user prompt:

\`\`\`
## PR Diff
\`\`\`diff
\${diff}
\`\`\`
\`\`\`

\`buildImplQuery\` (same file, line 926) already uses the **correct pattern**: diff-stat + file list + "read files from disk for surrounding context." No inline diff. This is what PR review should mirror.

Only Gemini fails because only Gemini:
- Runs as a CLI subprocess (Claude/Codex run in-process via SDKs)
- Uses non-streaming \`--output-format json\` (buffered)
- Feeds raw stdout to \`JSON.parse\` in \`extractReviewText\` / \`extractGeminiUsage\` (\`usage-extractor.js:42, 108\`)

When gemini-cli exits non-zero or partial, \`JSON.parse('')\` throws \`Unexpected end of JSON input\` and we lose all diagnostic info.

## Proposed fix (scope: tight)

**Make \`buildPRQuery\` mirror \`buildImplQuery\`'s filesystem pattern.** Replace the inline diff with:
- Diff-stat + changed-file list
- Write the full diff to a temp file (e.g., \`/tmp/codev-pr-<N>-<ts>.diff\`)
- Instruct the model to read that file from disk for full context

Applies to all three models uniformly — not a gemini-specific patch, a structural improvement.

## Out of scope (file separate issues if we hit them)

- Switch Gemini to \`--output-format stream-json\` with incremental parsing
- Harden \`extractGeminiUsage\` to log first/last 200 bytes on parse failure
- Explicit size-limit error instead of silent crash
- Fallback to per-phase / per-file review for oversized PRs

## Call-site references

- \`buildPRQuery\` — \`packages/codev/src/commands/consult/index.ts:833\`
- \`buildImplQuery\` (the pattern to mirror) — \`packages/codev/src/commands/consult/index.ts:926\`
- Gemini CLI spawn — same file, ~line 484 onwards
- Usage extractor — \`packages/codev/src/commands/consult/usage-extractor.ts\`

## Acceptance

- \`consult -m gemini --type integration pr <N>\` succeeds on the PR #796 class of inputs (>500KB diff)
- \`buildPRQuery\` no longer inlines the raw diff; model is pointed at a temp file path
- \`consult -m codex\` and \`consult -m claude\` for PR review continue to work (they benefit from the payload reduction equally)

waleedkadous:
Fixes #684

## Root Cause

`buildPRQuery` (`packages/codev/src/commands/consult/index.ts:833`) inlined the full `git diff` verbatim into the prompt. For large PRs (~800KB+, e.g. PR #796 with 831KB / 16,846 lines), this payload:

- Blew past gemini-cli v0.37.x's `--output-format json` buffered path → `Unexpected end of JSON input` in 0.3s → exit 13
- Worked for Claude/Codex because they run in-process via SDKs (no CLI buffering), but still bloated the prompt unnecessarily

PR #681 (bugfix #680) fixed the *transport* (stdin + `NODE_OPTIONS=--max-old-space-size=8192`). This fix attacks the *payload* itself.

## Fix

`buildPRQuery` now mirrors `buildImplQuery`'s filesystem pattern:

- Writes the full diff to a private `mkdtempSync` dir with mode `0600` and `flag: 'wx'` (no symlink follow, no clobber)
- Prompt includes: file list, diff size (bytes / lines), and the temp-file path
- Instructs the model to `Read` the file from disk for exact changes
- Applies uniformly to Gemini, Codex, and Claude — all three benefit from the smaller prompt

Split out a pure `composePRQueryText()` helper from the I/O wrapper so the prompt-shape can be tested without mocking the forge layer.

## Test Plan

- [x] Added regression test (`packages/codev/src/commands/consult/__tests__/pr-query.test.ts`) that asserts:
  - The prompt references the diff file path
  - No `` ```diff `` fence appears (the pre-fix pattern)
  - Prompt stays small (< 10KB) regardless of simulated diff size
  - Temp dir + file follow the secure convention (owner-only mode, EEXIST on collision)
- [x] All 74 existing consult tests still pass
- [x] `porch check` (build + tests) green

## CMAP Review

3-way PR review completed:

| Model  | Verdict          | Confidence |
|--------|------------------|------------|
| Gemini | APPROVE          | HIGH       |
| Codex  | REQUEST_CHANGES  | HIGH       |
| Claude | APPROVE          | HIGH       |

**Codex's REQUEST_CHANGES**: flagged the temp-file was written to `os.tmpdir()` with default perms (world-readable risk + symlink race). **Addressed** in commit `e58b5f19`:
- `fs.mkdtempSync` for a private owner-only dir (no predictable filename to pre-plant)
- `writeFileSync(..., { mode: 0o600, flag: 'wx' })` (refuses to follow symlinks or overwrite)
- New test asserts the security contract end-to-end

## Out of Scope (per issue)

- Switch Gemini to `--output-format stream-json` with incremental parsing
- Harden `extractGeminiUsage` to log first/last 200 bytes on parse failure
- Explicit size-limit error instead of silent crash
- Fallback to per-phase / per-file review for oversized PRs

These remain as follow-up improvements.

### DIFF
--- a/packages/codev/src/commands/consult/__tests__/pr-query.test.ts
+++ b/packages/codev/src/commands/consult/__tests__/pr-query.test.ts
@@ -1,0 +1,91 @@
+/**
+ * Regression tests for PR query construction (#684).
+ *
+ * Covers the fix for "Gemini consult fails on large PR diffs due to inlined
+ * payload size": buildPRQuery must write the diff to a temp file and
+ * reference the path, NOT inline the raw diff in the prompt.
+ */
+
+import { describe, it, expect } from 'vitest';
+import * as fs from 'node:fs';
+import { tmpdir } from 'node:os';
+import * as path from 'node:path';
+
+import { _composePRQueryText as composePRQueryText } from '../index.js';
+
+describe('composePRQueryText (#684)', () => {
+  const baseParams = {
+    prId: '796',
+    info: '{"title":"Some PR","state":"OPEN","additions":16846,"deletions":0}',
+    changedFiles: ['src/foo.ts', 'tests/foo.test.ts'],
+    comments: '(No comments)',
+    diffPath: '/tmp/codev-pr-796-1234567890.diff',
+    diffBytes: 851_000,
+    diffLines: 16846,
+  };
+
+  it('references the diff file path', () => {
+    const q = composePRQueryText(baseParams);
+    expect(q).toContain(baseParams.diffPath);
+  });
+
+  it('does NOT inline the raw diff content', () => {
+    // A recognisable chunk of real diff syntax we passed (or would pass) should
+    // not appear in the prompt — the model is pointed at the file instead.
+    const fakeDiff = 'diff --git a/src/foo.ts b/src/foo.ts\n+++ b/src/foo.ts\n+SENTINEL_DIFF_BODY';
+    const q = composePRQueryText({ ...baseParams, info: `${baseParams.info}\n${fakeDiff}` });
+    // The sentinel only appears because we stuffed it into `info` above —
+    // verify the prompt has no diff-fenced block with raw diff syntax.
+    // Specifically: no "```diff" fence (that was the pre-fix pattern).
+    expect(q).not.toContain('```diff');
+  });
+
+  it('lists changed files', () => {
+    const q = composePRQueryText(baseParams);
+    for (const f of baseParams.changedFiles) {
+      expect(q).toContain(`- ${f}`);
+    }
+    expect(q).toContain('## Changed Files (2)');
+  });
+
+  it('reports diff size in the prompt', () => {
+    const q = composePRQueryText(baseParams);
+    expect(q).toContain(`${baseParams.diffBytes} bytes`);
+    expect(q).toContain(`${baseParams.diffLines} lines`);
+  });
+
+  it('keeps the prompt small regardless of diff size', () => {
+    // Simulated 800KB diff — the prompt must remain small (< ~10KB).
+    // This is the core regression guard: previously, a large diff blew the
+    // prompt past what gemini-cli's JSON path could survive.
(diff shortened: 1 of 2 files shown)


waleedkadous:
Excellent fix — thank you for the quick turnaround. The split into a pure `composePRQueryText` helper is a nice touch; it made the tests much cleaner. The security hardening Codex asked for (mkdtempSync + 0o600 + wx) is exactly right, and the <10KB prompt-size assertion is a great regression guard.

Approved. Please merge.

PR #715, opened 2026-04-30
[Bugfix #710] Skip E2E tests that depend on removed porch run orchestrator

### LINKED ISSUE #710: E2E: 12 porch single-phase / happy-path failures (singlePhaseResult is null)
## Summary

12 of 160 vitest E2E tests fail in `packages/codev/src/commands/porch/__tests__/e2e/`, all in two files. All failures share the same shape: the test harness expects a structured `__PORCH_RESULT__` payload from porch's stdout, but receives `null`.

Discovered while running `pnpm --filter @cluesmith/codev test:e2e` during the v3.0.1 release on 2026-04-30. Not a release blocker (143/160 = 89% pass rate, all failures in one functional area), but worth fixing.

## Failing tests

**`src/commands/porch/__tests__/e2e/scenarios/happy-path.test.ts`**
- `Happy Path > continues to plan phase after spec approval` — *expected null to be 'plan-approval'*
- `Full Lifecycle > completes full SPIR protocol with auto-approve` — *expected false to be true*

**`src/commands/porch/__tests__/e2e/scenarios/single-phase.test.ts`**
- `Single Phase Mode > runs specify phase and returns structured result` — *expected null not to be null*
- `Single Phase Mode > can resume with another --single-phase after gate approval` — *expected null not to be null*
- `Single Phase Result Format > includes artifact path in result for specify phase` — *expected null not to be null*
- (plus 7 additional cases in these files with the same shape)

**Total: 12 failed | 143 passed | 5 skipped (160).** Test files: 6 failed | 11 passed (17).

## Hypotheses

The `__PORCH_RESULT__` JSON sentinel in porch's stdout isn't reaching the test's parser. Likely either:

1. **Test harness regression** — the parser for `__PORCH_RESULT__` (in the test fixtures) broke after some output buffering or line-splitting change.
2. **Porch output regression** — porch isn't emitting the result sentinel anymore (e.g., a recent change to the result-output codepath).
3. **Build artifact staleness** — `pnpm test:e2e` runs `pnpm build && vitest run`. If the build is incremental and a stale dist file masks the latest source, tests would run against old porch behavior. Worth ruling out by `pnpm clean && pnpm test:e2e`.

## Repro

```bash
pnpm --filter @cluesmith/codev test:e2e
```

The failures are deterministic across runs (not flakes).

## Impact

- Unit tests (2,564 of them) all pass — only the E2E single-phase scenarios are affected
- The `--single-phase` mode is for programmatic/planner-style use, not the human-driven SPIR workflow
- Released 3.0.1 ships with this issue; not surfaced in the published release notes

## Suggested approach

Bugfix protocol. First step is figuring out whether porch is the regressed party or the test harness is. A 5-minute investigation will probably reveal which side broke.

waleedkadous:
Fixes #710

## Root Cause

The four affected E2E scenario files (`single-phase`, `happy-path`, `feedback-loop`, `benchmark`) all spawn `porch run` or `porch run --single-phase` via the helpers in `runner.ts`:

- `runPorchUntilGate`, `runPorchWithAutoApprove` → `porch run`
- `runPorchSinglePhase` → `porch run --single-phase`
- `runPorchOnce` → `porch run --once`

`porch run` was removed in spec 0095 (commit `ed2012ae`, "[Spec 0095][Phase: phase_2] refactor: Remove porch orchestrator"). The CLI now responds with:

```
Error: 'porch run' has been removed. Use 'porch next <id>' instead.
```

…and exits with code 1 before writing anything to stdout. The runner's gate / `__PORCH_RESULT__` regexes therefore never match, `singlePhaseResult` stays `null`, and `hitGate` stays `null` — which is exactly the failure shape described in the issue.

## Fix

The orchestration loop these tests describe was deliberately moved out of porch — the parent agent now executes tasks emitted by `porch next`, so an "orchestrator-style" E2E test no longer applies. Reviving the coverage requires a planner-aware harness and belongs to its own spec.

This PR:

- Marks the four affected scenario describes as `describe.skip` with a clear annotation pointing to issue #710 and spec 0095.
- Adds `__tests__/run-removed.test.ts`, a regression test that pins the migration message so anyone (including the previously-broken runner) who still invokes `porch run` keeps getting a clear pointer to `porch next`. The test invokes `cli()` directly (mocking `console.error` + `process.exit`) so it does not depend on `dist/` — same pattern as `bugfix-711-pending-and-version.test.ts`.

## Test Plan

- [x] All 277 porch unit tests pass (`pnpm test src/commands/porch/__tests__`)
- [x] New regression test passes locally (and in CI without a prior build, after the cli()-rewrite fix)
- [x] E2E scenario tests now report as skipped (`vitest run --config vitest.e2e.config.ts` on each affected file shows 0 failed, all skipped)
- [x] Net diff: 110 LOC (well under 300)

## CMAP Review

Three-way review run with `--protocol bugfix --type pr`:

| Reviewer | Verdict | Confidence | Notes |
|----------|---------|------------|-------|
| Gemini   | APPROVE | HIGH | "Correctly skips obsolete E2E tests relying on the deleted 'porch run' orchestrator and adds a robust regression test." |
| Codex    | REQUEST_CHANGES | HIGH | Flagged stale in-progress `status.yaml` and missing CMAP section in PR body. Both addressed. |
| Claude   | COMMENT | HIGH | Flagged the regression test's `dist/` dependency via `spawnSync`. Addressed by switching to direct `cli()` invocation. |

All blocking issues addressed:
- `status.yaml` advanced from `investigate` → `fix` via `porch done` with `--pr 715 --branch builder/bugfix-710` (recorded in pr_history)
- Regression test rewritten to call `cli()` directly so it works without `dist/`
- This PR body updated with CMAP results

### DIFF
--- a/codev/projects/bugfix-710-e2e-12-porch-single-phase-happ/status.yaml
+++ b/codev/projects/bugfix-710-e2e-12-porch-single-phase-happ/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-710
 title: e2e-12-porch-single-phase-happ
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,7 +9,7 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-04-30T19:33:52.093Z'
-updated_at: '2026-04-30T23:05:42.271Z'
+updated_at: '2026-04-30T23:05:47.776Z'
 pr_history:
   - phase: investigate
     pr_number: 715
--- a/codev/projects/bugfix-710-e2e-12-porch-single-phase-happ/status.yaml
+++ b/codev/projects/bugfix-710-e2e-12-porch-single-phase-happ/status.yaml
@@ -9,4 +9,9 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-04-30T19:33:52.093Z'
-updated_at: '2026-04-30T19:33:52.094Z'
+updated_at: '2026-04-30T23:05:42.271Z'
+pr_history:
+  - phase: investigate
+    pr_number: 715
+    branch: builder/bugfix-710
+    created_at: '2026-04-30T23:05:42.270Z'
--- a/codev/projects/bugfix-710-e2e-12-porch-single-phase-happ/status.yaml
+++ b/codev/projects/bugfix-710-e2e-12-porch-single-phase-happ/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-710
 title: e2e-12-porch-single-phase-happ
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,7 +9,7 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-04-30T19:33:52.093Z'
-updated_at: '2026-04-30T23:05:47.776Z'
+updated_at: '2026-04-30T23:08:27.967Z'
 pr_history:
   - phase: investigate
     pr_number: 715
--- a/codev/projects/bugfix-710-e2e-12-porch-single-phase-happ/status.yaml
+++ b/codev/projects/bugfix-710-e2e-12-porch-single-phase-happ/status.yaml
@@ -1,0 +1,12 @@
+id: bugfix-710
+title: e2e-12-porch-single-phase-happ
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates: {}
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-04-30T19:33:52.093Z'
+updated_at: '2026-04-30T19:33:52.094Z'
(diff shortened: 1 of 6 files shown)


waleedkadous:
## Architect review — CI failure needs fixing before merge

CI Unit Tests job fails on `run-removed.test.ts`:

```
Error [ERR_MODULE_NOT_FOUND]: Cannot find module '.../dist/commands/porch/index.js'
imported from .../bin/porch.js
    at finalizeResolution (node:internal/modules/esm/resolve:283:11)
```

The regression test spawns `node bin/porch.js run some-id` as a subprocess. `bin/porch.js` imports from `dist/`, but the unit test job doesn't build first (only `test:e2e` does `pnpm build && vitest run`). Locally it passes because dist is already built — CI starts clean.

**Fix**: rewrite the test to call `cli()` directly instead of via subprocess. The bugfix-711 PR (#714) does this for `porch --version`:

```ts
import { cli } from '../index.js';

it('exits 1 with a migration message pointing to porch next', async () => {
  const stderr: string[] = [];
  const spy = vi.spyOn(console, 'error').mockImplementation((...a) => {
    stderr.push(a.map(String).join(' '));
  });
  // Note: cli() throws or process.exits — wrap accordingly
  await expect(cli(['run', 'some-id'])).rejects.toThrow(); // or check process.exit mock
  expect(stderr.join('\n')).toContain("'porch run' has been removed");
  expect(stderr.join('\n')).toContain('porch next');
  spy.mockRestore();
});
```

(Adjust to the actual exit mechanism — looking at line 929 of `index.ts`, it logs to `console.error` and presumably `process.exit(1)`s. You may need to mock `process.exit` too.)

The rest of the PR (4 `describe.skip`s with clear comments + the migration message regression test) is correctly scoped. Please fix the test approach and CI should go green.

— Architect

PR #720, opened 2026-05-02
[Bugfix #717] Fall back to Tower terminal_sessions when local state misses a builder

### LINKED ISSUE #717: afx attach can't find builders registered only in Tower's terminal registry
## Summary

`afx attach` fails to find a builder that `afx status` shows as active. The builder is visible in the Tower web UI and responds to `afx send`, but `afx attach -p <id>` says "Builder not found."

## Root Cause

`afx status` (when Tower is running) queries the Tower API, which reads from `global.db terminal_sessions`. But `afx attach` only queries the **local** `.agent-farm/state.db builders` table via `getBuilder()` / `getBuilders()` in `state.ts`.

These are two separate databases with no synchronization. If the local `builders` row was never written (e.g., spawned by a different session, race condition during spawn, or state.db was cleared), the builder exists in Tower but is invisible to `attach`.

## Reproduction

1. Have a Tower running with an active builder (visible in `afx status` and Tower web UI)
2. Verify `.agent-farm/state.db` builders table is empty: `sqlite3 .agent-farm/state.db "SELECT * FROM builders;"`
3. Run `afx attach -p <builder-id>` → "Builder not found"
4. Run `afx send <builder-id> "hello"` → works (routes through Tower API)

## Expected Behavior

`afx attach` should fall back to querying Tower's terminal registry when the local builders table has no match, similar to how `afx status` already does.

## Relevant Code

- `commands/attach.ts:41-63` — `findBuilderById()` only queries local state
- `commands/status.ts:26-62` — queries Tower API when Tower is running
- `servers/tower-terminals.ts:819-828` — Tower tracks builders by `role_id`
- `state.ts:75-111` — `upsertBuilder()` writes only to local state.db

## Workaround

Use `afx send <builder-id> "<message>"` (routes through Tower) or interact via the Tower web UI at `http://localhost:4100`.

waleedkadous:
## Summary
Fixes #717

## Root Cause
Two related bugs combined to make `afx attach` unreliable:

1. **Lookup miss**: `findBuilderById` / `findBuilderByIssue` only consulted the local `state.db builders` table. Tower's terminal registry (`global.db terminal_sessions`) was never consulted, so a builder visible to `afx status` and `afx send` could still error as "Builder not found" from `afx attach -p <id>` whenever the local row was missing — e.g. spawned in a different session, lost during a `state.db` reset, or never written due to a spawn-time race.

2. **Socket attribution** *(uncovered by CMAP review)*: `findShellperSocket` queried `terminal_sessions WHERE workspace_path = ?` using `normalizeWorkspacePath(builder.worktree)` — but Tower stores `workspace_path = config.workspaceRoot` at spawn time. The SQLite lookup never matched a builder, so socket discovery always fell through to a directory scan that returns the *first* shellper socket — i.e. potentially the wrong builder when multiple are running. The first commit's Tower fallback inherited this bug.

## Fix
**Commit 1**: Tower-registry fallback in `findBuilderById` and `findBuilderByIssue`. When local lookup misses, query `terminal_sessions WHERE workspace_path = ? AND type = 'builder'` and reconstruct a minimal `Builder` from `role_id` (id), `cwd` (worktree), and `label` (name). For `--issue`, parse the `bugfix-<n>` suffix from `role_id`.

**Commit 2**: Switch `findShellperSocket`'s SQLite query to use `getConfig().workspaceRoot` instead of `builder.worktree`, so socket discovery correctly identifies the right shellper for the right builder.

Local-state lookup is unchanged when it succeeds; existing happy-path behavior is preserved.

## Test Plan
- [x] 28 attach tests pass (23 pre-existing + 5 new)
- [x] New regression tests:
  - `--issue` falls back to Tower when local state has no match
  - `--project` exact match falls back to Tower
  - `--project` substring match falls back to Tower
  - `--project` errors when neither registry knows the builder
  - End-to-end terminal-mode attach for a Tower-only builder verifies SQLite is queried with workspace ROOT (not worktree) and finds the correct socket
- [x] Pre-existing test that asserted the broken `findShellperSocket` arg signature was corrected
- [x] TypeScript check passes
- [x] Net diff: 226 LOC across both commits (under 300 LOC bugfix threshold)

## CMAP Review

| Model  | Verdict          | Notes |
|--------|------------------|-------|
| Gemini | REQUEST_CHANGES → addressed | Flagged the `findShellperSocket` workspace-root mismatch as critical. Fixed in commit 2. |
| Codex  | REQUEST_CHANGES → addressed | Same finding as Gemini, plus noted that the first commit's tests only exercised `--browser` mode. Both fixed in commit 2 (terminal-mode end-to-end test added). |
| Claude | APPROVE          | Verified the workspace_path semantics for the new fallback are correct. Noted the `findShellperSocket` issue as pre-existing/unrelated; commit 2 chooses to fix it anyway since the fallback inherits the same trap. |

### DIFF
--- a/codev/projects/bugfix-717-afx-attach-can-t-find-builders/status.yaml
+++ b/codev/projects/bugfix-717-afx-attach-can-t-find-builders/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-717
 title: afx-attach-can-t-find-builders
 protocol: bugfix
-phase: pr
+phase: verified
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-02T01:51:59.248Z'
-updated_at: '2026-05-02T02:10:38.520Z'
+updated_at: '2026-05-02T02:10:47.664Z'
--- a/codev/projects/bugfix-717-afx-attach-can-t-find-builders/status.yaml
+++ b/codev/projects/bugfix-717-afx-attach-can-t-find-builders/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-717
 title: afx-attach-can-t-find-builders
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-02T01:51:59.248Z'
-updated_at: '2026-05-02T02:09:12.105Z'
+updated_at: '2026-05-02T02:10:38.520Z'
--- a/codev/projects/bugfix-717-afx-attach-can-t-find-builders/status.yaml
+++ b/codev/projects/bugfix-717-afx-attach-can-t-find-builders/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-717
 title: afx-attach-can-t-find-builders
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-02T01:51:59.248Z'
-updated_at: '2026-05-02T01:51:59.249Z'
+updated_at: '2026-05-02T02:09:12.105Z'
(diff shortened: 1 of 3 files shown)


waleedkadous:
## Architect Review

Excellent work — thank you. The fix is clean, and the CMAP review process did exactly what it's supposed to: Gemini and Codex caught the related `findShellperSocket` workspace_path bug that the original ticket never named, and you addressed both REQUEST_CHANGES with a focused commit 2 plus a proper end-to-end regression test rather than rebutting them.

The Tower-row → Builder reconstruction is sensible (placeholder fields only matter in `displayBuilderList`, which the fallback path doesn't reach), and the corrected workspace_path query closes a real multi-builder mis-attribution risk.

All 6 CI checks green. Approved — please merge.

waleedkadous:
@timeleft-- this is now merged to `main`. When you get a chance, could you pull latest and confirm `afx attach` works for you in the scenario you originally hit? Thanks again for the well-researched bug report — the diagnostic detail in the issue made this a really clean fix.

PR #743, opened 2026-05-14
Fix #742: rewrite BUGFIX consult templates to drop SPIR criteria

waleedkadous:
## Summary

The BUGFIX protocol's `pr-review.md` and `impl-review.md` consult templates were byte-identical to the SPIR versions, causing reviewer models (Codex in particular) to return `REQUEST_CHANGES` on legitimate BUGFIX PRs for criteria that don't apply to the BUGFIX protocol.

Fixes #742

## Root Cause

Reported by Shannon architect after seeing it 4× the same evening (PRs #1357, #1358, #1359, #1370). Confirmed via:

```bash
$ diff codev/protocols/bugfix/consult-types/pr-review.md \
       codev/protocols/spir/consult-types/pr-review.md
$  # (empty — files were byte-identical)
```

The same was true for `impl-review.md`, both in `codev/` (self-hosted) and `codev-skeleton/` (shipped template). Reviewer models followed the SPIR-flavored instructions correctly — they asked about the SPIR three-document trinity (`codev/specs/`, `codev/plans/`, `codev/reviews/`), demanded `[Spec NNNN][Phase]` commit format, flagged `status.yaml.build_complete: false` as a builder-fixable issue, and applied multi-phase scoping that doesn't exist in BUGFIX. None of this is valid for a BUGFIX PR.

## Fix

Rewrote four template files to reflect what BUGFIX actually produces:

- `codev/protocols/bugfix/consult-types/pr-review.md`
- `codev/protocols/bugfix/consult-types/impl-review.md`
- `codev-skeleton/protocols/bugfix/consult-types/pr-review.md`
- `codev-skeleton/protocols/bugfix/consult-types/impl-review.md`

Each template now:
- Identifies itself as a BUGFIX template
- Lists the protocol's actual artifacts: GitHub Issue + code fix + regression test + PR body
- Documents the BUGFIX commit format (`Fix #N:` / `[Bugfix #N]`)
- Includes an explicit "Out of Scope" section that forbids citing the SPIR-only artifacts (specs/plans/reviews trinity, `[Spec NNNN][Phase]` format, `status.yaml.build_complete`, multi-phase scoping) as REQUEST_CHANGES reasons
- Keeps the same `APPROVE | REQUEST_CHANGES | COMMENT` verdict format so the consult-output pipeline doesn't need changes

`codev-skeleton/` copies match `codev/` copies byte-for-byte, so projects that adopt the template via `codev init` / `codev adopt` get the same fix.

## Test Plan

[...]

PR #748, opened 2026-05-14
[Bugfix #745] Ignore untracked files in afx spawn dirty-worktree check

### LINKED ISSUE #745: Bug: afx spawn dirty-worktree check is too strict on untracked files; forces --force on every spawn
## Summary

`afx spawn`'s pre-spawn worktree-cleanliness check fails on **untracked files** at the repo root, even though those files don't move into the builder's worktree. Chronic untracked artifacts (local-install symlinks, build outputs, Claude state files) force every spawn to use `--force`.

## Reporter

Shannon architect, 2026-05-14 ~20:35 PDT via team channel. Independently confirmed locally in this repo 2026-05-14 ~20:35 PDT (had to `--force` issue #742's bugfix builder spawn).

## Repro

```
$ git status
On branch main
Your branch is up to date with 'origin/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.claude/scheduled_tasks.lock
	bin/
	packages/codev/dashboard/

$ afx spawn 742 --protocol bugfix
[error] Uncommitted changes detected in main worktree.
  Builders branch from HEAD, so uncommitted files (specs, plans,
  codev updates) will NOT be visible to the builder.
  Please commit or stash your changes first, then retry.
  Use --force to skip this check.
```

These untracked files are NOT specs/plans/codev updates — they're local artifacts:
- `.claude/scheduled_tasks.lock` — Claude Code internal state
- `bin/` — symlinks created by `pnpm local-install`
- `packages/codev/dashboard/` — local build output

The check exists to prevent silently dropping work the architect just wrote (specs, plans, codev/ changes). Untracked files at the root are almost never that.

## Fix shape (two options)

**Option A (preferred): only fail on modified-tracked files, not untracked.** Untracked files at HEAD time are not "work the builder needs to see" — they're local artifacts that didn't exist on `main`. Loosen the check to `git diff --quiet HEAD` (tracked modifications only), ignoring untracked.

**Option B: `.builderignore`.** Honor a `.builderignore` file (or reuse `.gitignore`-style globs in `.codev/config.json`) listing paths to ignore in the cleanliness check.

Option A is simpler and covers the actual concern (don't lose unstaged work). Option B is more flexible but adds a new config surface.

## Counter-argument to consider

The current strict behavior caught a real bug class once: an architect wrote a spec, didn't commit, spawned a builder, and the builder couldn't see the spec. Loosening to ignore untracked files keeps that protection (a `git add`-but-not-committed spec IS a modified-tracked file via the index).

Actually — `git status` shows untracked as "Untracked files" and staged-but-uncommitted as "Changes to be committed." Once `git add codev/specs/foo.md` runs, foo.md becomes tracked (in the index). So Option A correctly catches the spec case while ignoring `bin/` symlinks.

## Files to fix

The dirty-worktree check is in `packages/codev/src/agent-farm/commands/spawn.ts` (or similar). Need to grep for the "Uncommitted changes detected" string.

## Suggested protocol

BUGFIX — single small behavioral change in one file.

waleedkadous:
## Summary

Fixes #745

The `afx spawn` pre-spawn cleanliness check used `git status --porcelain`, which flags **untracked files** (status `??`). In repos with chronic untracked artifacts (`bin/` from `pnpm local-install`, `.claude/scheduled_tasks.lock`, `packages/codev/dashboard/` build output), this forced `afx spawn --force` on every spawn — defeating the purpose of the safety check.

## Root Cause

`git status --porcelain` returns one line per untracked file:

```
?? .claude/scheduled_tasks.lock
?? bin/
?? packages/codev/dashboard/
```

The check then aborts as if real work were uncommitted, even though those files don't move into the builder's worktree and weren't authored as part of any spec/plan.

## Fix

Switch to `git status --porcelain --untracked-files=no`. This:

- **Ignores untracked files** (status `??`) — local artifacts that didn't come from `main`
- **Still catches modifications to tracked files** (` M`, `M `)
- **Still catches the spec case via the index** — `git add codev/specs/foo.md` puts the file in the index where it shows as `A ` (tracked addition), not `??` (untracked)

The check is also extracted into `utils/git.ts#hasUncommittedTrackedChanges` so it can be unit-tested without spawn.ts's transitive imports.

### Documented tradeoff

If the architect creates a new file but **never** runs `git add` (totally untracked, never staged), the check no longer fires. The issue reporter explicitly accepted this:

> Once `git add codev/specs/foo.md` runs, foo.md becomes tracked (in the index). So Option A correctly catches the spec case while ignoring `bin/` symlinks.

The protected workflow is `git add` then spawn; an entirely-unstaged file is treated as draft work the architect is still authoring. The prior behavior produced chronic false positives that forced `--force` on every spawn, which defeated the check entirely.

A regression test pins this behavior in place, and the docstring on `hasUncommittedTrackedChanges` makes the tradeoff visible at the call site.

## Test Plan

- [x] Added regression tests (`spawn-dirty-check.test.ts`, 8 cases):
  - clean worktree → false
  - **untracked-only → false** (the bug)
  - tracked + modified → true
  - tracked + staged → true
  - **previously-untracked + staged (spec case) → true**
  - mixed untracked + tracked-modified → true
  - **brand-new untracked file → false (documented tradeoff)**
  - non-git directory → false (fail-open)
- [x] All 222 spawn-related tests pass (215 pre-existing + 7 new)
- [x] `porch check` passes (build 5.4s, tests 20.2s)
- [x] Net diff: 155 LOC (well under 300 LOC threshold)

## CMAP Review

| Model | Verdict | Notes |
|---|---|---|
| Claude | APPROVE | "Clean, well-tested fix that correctly relaxes the over-strict dirty-worktree check while preserving the spec-case safety invariant." |
| Codex | REQUEST_CHANGES | Flagged the unstaged-new-file regression. Addressed by documenting the tradeoff in the docstring + adding a regression test that pins it. The issue reporter explicitly chose this tradeoff. |
| Gemini | — | Could not complete review due to a workspace-allowlist environment issue (`/var/folders/.../codev-pr-*.diff` outside Gemini's allowed paths). Unrelated to this PR; matches an existing CMAP-tool limitation. |

### DIFF
--- a/codev/projects/bugfix-745-bug-afx-spawn-dirty-worktree-c/status.yaml
+++ b/codev/projects/bugfix-745-bug-afx-spawn-dirty-worktree-c/status.yaml
@@ -1,0 +1,12 @@
+id: bugfix-745
+title: bug-afx-spawn-dirty-worktree-c
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates: {}
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-05-14T20:59:04.416Z'
+updated_at: '2026-05-14T20:59:04.417Z'
--- a/codev/projects/bugfix-745-bug-afx-spawn-dirty-worktree-c/status.yaml
+++ b/codev/projects/bugfix-745-bug-afx-spawn-dirty-worktree-c/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-745
 title: bug-afx-spawn-dirty-worktree-c
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-14T20:59:04.416Z'
-updated_at: '2026-05-14T20:59:04.417Z'
+updated_at: '2026-05-14T21:04:09.773Z'
--- a/codev/projects/bugfix-745-bug-afx-spawn-dirty-worktree-c/status.yaml
+++ b/codev/projects/bugfix-745-bug-afx-spawn-dirty-worktree-c/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-745
 title: bug-afx-spawn-dirty-worktree-c
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-14T20:59:04.416Z'
-updated_at: '2026-05-14T21:04:09.773Z'
+updated_at: '2026-05-14T21:05:56.856Z'
--- a/codev/projects/bugfix-745-bug-afx-spawn-dirty-worktree-c/status.yaml
+++ b/codev/projects/bugfix-745-bug-afx-spawn-dirty-worktree-c/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-745
 title: bug-afx-spawn-dirty-worktree-c
 protocol: bugfix
-phase: pr
+phase: verified
 plan_phases: []
 current_plan_phase: null
 gates: {}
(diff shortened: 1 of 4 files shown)


waleedkadous:
Architect approval — clean fix and a sound rebuttal of Codex's objection. The unstaged-new-file tradeoff was explicitly anticipated in the issue body; documenting it in the docstring + pinning it with a regression test is exactly the right shape. Approved — please wait for CI green, then merge with `gh pr merge 748 --merge --admin` (same self-review workaround as #743).

PR #775, opened 2026-05-19
[Bugfix #774] Open workspace state.db in detectCurrentBuilderId

### LINKED ISSUE #774: Multi-architect routing: builder→architect messages misrouted to 'main' when worktree has its own state.db
## Summary

Builder→architect messages are silently misrouted to `main` (instead of the spawning sibling architect) when run inside a builder worktree that has its own `.agent-farm/state.db`. The affinity primitive shipped in v3.0.5 is bypassed end-to-end for sibling-architect workspaces — every sibling architect (ob-refine in the reporting case) gets cut out of its own builders' messages.

## Repro

In a workspace with two architects registered:

- `main` (terminal `fe6b1cd8...`)
- `ob-refine` (terminal `48ad2d87...`)

`ob-refine` spawned `builder-bugfix-1599`. The builder's `.agent-farm/state.db` row correctly records:

```
builder-bugfix-1599 | implementing | spawned_by_architect=ob-refine | terminal=513db860...
```

The builder then ran `afx send architect "..."` from inside `.builders/bugfix-1599/`. Tower log:

```
[2026-05-19T22:01:13.318Z] [INFO] Message sent: bugfix-1599 → architect (terminal fe6b1cd8...)
```

Terminal `fe6b1cd8...` is **main**, not ob-refine. Wrong terminal.

## Root cause

The sender logged is `bugfix-1599`, **not** the canonical `builder-bugfix-1599`. That mismatch breaks the affinity lookup downstream.

`detectCurrentBuilderId()` (`packages/codev/src/agent-farm/commands/send.ts:48`) does:

```ts
const worktreeDirName = match[1];                 // 'bugfix-1599'
const state = loadState();                        // <-- singleton DB
const builder = state.builders.find(b => /* worktree path match */);
return builder ? builder.id : worktreeDirName;    // fallback when miss
```

`loadState()` uses the singleton `getDb()`, which resolves `stateDir` via `findWorkspaceRoot()`. When CWD is `.builders/<id>/`, `findWorkspaceRoot()` returns the **worktree path**, not the workspace path, because the worktree is itself a full git checkout with its own `codev/` (config.ts:79). So `getDb()` opens:

```
/Users/.../shannon/.builders/bugfix-1599/.agent-farm/state.db
```

This file exists (created on first DB open) and is **empty**:

```
$ sqlite3 .../bugfix-1599/.agent-farm/state.db "SELECT COUNT(*) FROM builders;"
0
```

So the `state.builders.find(...)` returns `undefined`, and `detectCurrentBuilderId` falls back to the worktree directory name `bugfix-1599`. Tower receives `from='bugfix-1599'`.

In `tower-messages.ts:318`, `lookupBuilderSpawningArchitect('bugfix-1599', '/Users/.../shannon')` returns `undefined` because no row exists with `id='bugfix-1599'` — the row is keyed `builder-bugfix-1599`. With `spawningArchitect === undefined`, the routing skips the builder-context branch (line 320-342) and falls through to the **non-builder-sender** branch at line 344-347:

```ts
const terminalId =
  entry.architects.get(DEFAULT_ARCHITECT_NAME) ?? entry.architects.values().next().value!;
return { terminalId, workspacePath, agent: 'architect' };
```

→ Routes to `main`. The spawning architect is silently bypassed.

## Why this matters

Affinity routing was the headline feature of v3.0.5/v3.0.6 multi-architect. This bug means **every sibling-architect workspace** loses messages from its builders today. The porch wake-up path (`gate-approved`, `pr-ready` notifications) and the builder's own architect-bound updates all hit the wrong terminal. The data is correct in state.db, the dashboard is correct, the in-memory architect registry is correct — only the sender-ID resolution at send time is wrong.

## Fix sketch

`detectCurrentBuilderId` should open the **workspace's** state.db directly, not the singleton (which resolves to the worktree's). The correct pattern already exists in `state.ts:380` (`lookupBuilderSpawningArchitect`):

```ts
const dbPath = path.join(workspacePath, '.agent-farm', 'state.db');
if (!existsSync(dbPath)) return undefined;
const wsDb = new Database(dbPath, { readonly: true });
try {
  // ... lookup ...
} finally {
  wsDb.close();
}
```

Apply the same pattern in `detectCurrentBuilderId`: from CWD `.builders/<id>/`, the workspace root is two levels up; open *that* state.db (readonly), find the builder by worktree-path match, return its canonical `id`. The fallback to `worktreeDirName` should remain as a safety net but should never be hit in practice.

Add a regression test: spawn-fixture, builder in worktree with empty worktree-level state.db, assert `detectCurrentBuilderId()` returns the canonical `builder-<X>` ID by reading the workspace's state.db.

## Verification

Manual repro on shannon@2026-05-19:
- `ob-refine` spawned `bugfix-1599`
- builder's `afx send architect` → routed to `main` terminal `fe6b1cd8...`
- Expected: routed to `ob-refine` terminal `48ad2d87...`

## Severity

High — this silently breaks the multi-architect feature for every adopter (Shannon in production). User-visible symptom is "my sibling architect never hears from its builders." No data loss, but the feature is non-functional.

## Affected versions

v3.0.5 onward (anything with sibling-architect support).

waleedkadous:
## Summary
Fixes #774

## Root Cause
`detectCurrentBuilderId()` (in `packages/codev/src/agent-farm/commands/send.ts`) used the singleton `getDb()` via `loadState()`. From a CWD inside `.builders/<id>/`, `findWorkspaceRoot()` resolves to the **worktree** (worktrees are full git checkouts with their own `codev/`), so `getDb()` opens the worktree-local `.agent-farm/state.db`.

That DB is empty (it gets created on first open with the schema, but no builder rows). `state.builders.find(...)` returns `undefined`, and the function falls back to the worktree directory name (e.g. `bugfix-1599`) — not the canonical `builder-bugfix-1599`.

Downstream, in `tower-messages.ts:318`, `lookupBuilderSpawningArchitect('bugfix-1599', workspacePath)` returns `undefined` because the row is keyed `builder-bugfix-1599`. The router skips the builder-context branch and routes to `'main'` instead of the sibling architect (`ob-refine`) that spawned the builder. Sibling-architect affinity routing was effectively non-functional.

## Fix
`detectCurrentBuilderId()` now opens the **workspace's** `state.db` directly (readonly), mirroring the per-workspace-handle pattern already used in `lookupBuilderSpawningArchitect` (`state.ts:380`):

- Parse the workspace path out of the `.builders/<id>/` CWD pattern (it's always two levels up).
- Open `<workspace>/.agent-farm/state.db` readonly via a fresh `better-sqlite3` handle.
- Match by canonical worktree path first, then a tail-segment match as a legacy-row safety net.
- Fall back to the worktree directory name only when the workspace DB is missing or has no matching row.

The worktree-local `state.db` is no longer touched by the send path. The `loadState()` import is retained — it's still used by the broadcast (`--all`) helper, which legitimately reads from the singleton because it's enumerating builders the singleton workspace knows about.

## Test Plan
- [x] Reproduced the bug locally: from `.builders/bugfix-774/`, `detectCurrentBuilderId()` returned `bugfix-774` and created an empty worktree-level state.db.
- [x] Verified after the fix: returns `builder-bugfix-774` and does **not** create a worktree-level state.db.
- [x] Added 5 regression tests in `bugfix-774-detect-builder-id.test.ts` covering:
  - canonical resolution when CWD is the worktree (no worktree-local DB)
  - canonical resolution when the worktree has its **own empty** state.db (the original bug)
  - fallback to worktree dir name when workspace DB is missing
  - fallback when no row matches
  - returns `null` when CWD is not a builder worktree
- [x] All 1775 agent-farm unit tests pass (`vitest run src/agent-farm`).

## CMAP Review
Skipped per BUGFIX scope — fix is < 200 LOC, narrowly scoped to a single function, with high-coverage regression tests. Happy to run if the architect would like.

## Scope
- 1 source change (commands/send.ts): replace singleton lookup with workspace-direct readonly DB read.
- 1 new test file (5 cases).
- Net diff: ~135 LOC (well under 300).

### DIFF
--- a/codev/projects/bugfix-774-multi-architect-routing-builde/status.yaml
+++ b/codev/projects/bugfix-774-multi-architect-routing-builde/status.yaml
@@ -1,0 +1,12 @@
+id: bugfix-774
+title: multi-architect-routing-builde
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates: {}
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-05-19T22:15:47.639Z'
+updated_at: '2026-05-19T22:15:47.639Z'
--- a/packages/codev/src/agent-farm/__tests__/bugfix-774-detect-builder-id.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/bugfix-774-detect-builder-id.test.ts
@@ -1,0 +1,100 @@
+/**
+ * Regression test for Issue #774: Builder→architect messages misrouted when
+ * worktree has its own state.db.
+ *
+ * The bug: `detectCurrentBuilderId()` used `loadState()` (singleton getDb()),
+ * which resolves to the worktree's own .agent-farm/state.db when CWD is
+ * inside `.builders/<id>/`. The worktree DB is empty, so the lookup falls
+ * back to the worktree directory name (e.g. `bugfix-774`) instead of the
+ * canonical builder ID (`builder-bugfix-774`). That breaks affinity routing
+ * downstream — the sibling architect that spawned the builder is bypassed
+ * and the message lands on 'main'.
+ *
+ * Fix: open the workspace's state.db directly (not the singleton).
+ */
+
+import { describe, it, expect, beforeEach, afterEach } from 'vitest';
+import { mkdirSync, mkdtempSync, rmSync } from 'node:fs';
+import { tmpdir } from 'node:os';
+import { join } from 'node:path';
+import Database from 'better-sqlite3';
+
+import { detectCurrentBuilderId } from '../commands/send.js';
+import { LOCAL_SCHEMA } from '../db/schema.js';
+
+function writeBuilderRow(dbPath: string, id: string, worktree: string): void {
+  const db = new Database(dbPath);
+  db.exec(LOCAL_SCHEMA);
+  db.prepare(
+    `INSERT INTO builders (id, name, worktree, branch, type, status, spawned_by_architect)
+     VALUES (?, ?, ?, ?, 'bugfix', 'implementing', 'ob-refine')`,
+  ).run(id, id, worktree, `builder/${id}`);
+  db.close();
+}
+
+describe('detectCurrentBuilderId — issue #774', () => {
+  let tmpRoot: string;
+  let workspacePath: string;
+  let worktreePath: string;
+  const origCwd = process.cwd();
+
+  beforeEach(() => {
+    tmpRoot = mkdtempSync(join(tmpdir(), 'bugfix-774-'));
+    workspacePath = join(tmpRoot, 'workspace');
+    worktreePath = join(workspacePath, '.builders', 'bugfix-1599');
+
+    mkdirSync(join(workspacePath, '.agent-farm'), { recursive: true });
+    mkdirSync(worktreePath, { recursive: true });
+
+    // Populate the WORKSPACE state.db with the canonical builder row.
+    writeBuilderRow(
+      join(workspacePath, '.agent-farm', 'state.db'),
+      'builder-bugfix-1599',
+      worktreePath,
+    );
+  });
+
+  afterEach(() => {
+    process.chdir(origCwd);
+    rmSync(tmpRoot, { recursive: true, force: true });
+  });
+
+  it('returns canonical ID when CWD is the worktree and worktree has no state.db', () => {
+    process.chdir(worktreePath);
+    expect(detectCurrentBuilderId()).toBe('builder-bugfix-1599');
+  });
+
+  it('returns canonical ID even when worktree has its own EMPTY state.db (the original bug)', () => {
+    // Simulate the v3.0.5 bug: the worktree opened its own state.db (created
+    // by a stray getDb() call) which has zero builder rows.
+    mkdirSync(join(worktreePath, '.agent-farm'), { recursive: true });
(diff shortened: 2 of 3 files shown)


waleedkadous:
Reviewed and verified end-to-end against shannon's real state.db. Walked through the fix from CWD `/Users/.../shannon/.builders/bugfix-1599`:

- Regex extracts workspace path + worktree dir
- Opens workspace `state.db` readonly
- Canonical-worktree match → returns `builder-bugfix-1599`

That's the right answer. Downstream `lookupBuilderSpawningArchitect` then resolves `ob-refine` and routes to its terminal. Bug fully addressed.

Particularly liked:
- `readonly: true` on the new handle — sidesteps the root cause (worktree state.db getting bootstrapped by stray `getDb()` calls from inside the worktree).
- Two-stage match (canonical path → tail fallback) handles both the common case and any legacy rows from a different machine.
- Mirrored the existing per-workspace-handle pattern from `state.ts:380` rather than inventing a new one.

The 5 new tests cover the bug scenario (worktree has its own empty state.db) cleanly.

One thought for a future PR (NOT this one): worktree-local empty `state.db` files will keep getting created by other code paths that call `getDb()` from inside a worktree. Doesn't break routing anymore, but a guard at `getDb()` itself ("refuse to bootstrap state.db inside `.builders/*/`") would be a nice belt-and-braces. Out of scope here.

Pending architect approval to merge.

Issue #832, opened 2026-05-23
Multi-architect conversation resume: disambiguate via per-architect session UUID


amrmelsayed:
## Context

Follow-up to #830 (architect session revival). #829 shipped builder conversation resume via on-disk jsonl discovery. That same mechanism extends cleanly to the **main** architect (committed on the #829 branch) because main is the only architect using Tower's automatic launchInstance path. It does **not** extend to named sibling architects added via `afx workspace add-architect` (Spec 755), because they share `cwd = workspacePath` with each other and with main — and the jsonl-discovery heuristic (newest `*.jsonl` by mtime) cannot tell which jsonl belongs to which named architect.

This issue scopes the multi-architect disambiguation work that #830 left open.

## Problem

A workspace with main + named siblings (e.g., `architect-2`, `reviewer-bob`) has multiple Tower-managed Claude PTYs running in the same cwd. Each writes its session jsonl to `~/.claude/projects/<encoded-workspacePath>/`. After a reboot:

- The reboot kills every architect's shellper.
- On `afx workspace start`, Tower's `launchInstance` re-spawns main. Spec 786 Phase 3 then iterates `state.db.architect` to re-add each persisted sibling via `addArchitect()`.
- With #830's main-only resume, main correctly picks the right jsonl. But sibling revivals would either skip the resume path entirely (current behavior in the #829 branch's `addArchitect` — deliberate) or, if naively enabled, attach to whichever architect's jsonl was most recent — possibly the wrong one.

So named siblings come back as fresh sessions, losing all conversation context. Operators using multi-architect workspaces lose more on reboot than single-architect users do.

### Why the named-architect case is sharper than main's

For specialised siblings (`reviewer`, `demos`, `casa`, etc.), there is no native per-architect role-doc loading in Codev — every architect boots with the same shared `codev/roles/architect.md`, and specialisation comes from the **first user message after `afx workspace add-architect`** (the brief that sets identity, scope, operating mode, output header conventions, etc.). That brief lives in the conversation, not on disk.

When a sibling loses its conversation, it loses the brief. It doesn't come back as "reviewer without recent context" — it comes back as a generic architect that doesn't know its lane. Main has the same role-doc preloaded plus an incremental conversation, so a main-only context loss is annoying; a sibling context loss erases the specialisation itself. The user has to re-send the brief by hand to restore identity. This is the failure mode worth fixing — not just "siblings lose context" but "siblings lose their job description."

### A second, silent path: shellper auto-restart

[...]

Issue #830, opened 2026-05-23
Workspace recover: revive Tower-managed architects after machine reboot


amrmelsayed:
## Context

Follow-up to #829, which built `afx workspace recover` for builders. The recovery flow (process revival + conversation-context restoration via `claude --resume <uuid>`) is now in place but only covers builders. Tower-managed architects die in the same machine-reboot scenario and have no equivalent recovery path.

## Problem

Two architect spawn surfaces exist today:

- **`afx architect`** — runs Claude directly in the user's current terminal. No Tower dependency. If the machine reboots, the user just opens a new terminal and runs `afx architect` again. **Not in scope** for recovery.
- **`afx workspace add-architect`** (Spec 755) and the **main architect** spawned implicitly by `afx workspace start` — Tower-managed PTYs, persisted in `terminal_sessions`, killed by reboot. **In scope.**

After a reboot of a workspace with several Tower-managed architects (main + named siblings per Spec 755), the user has to manually re-spawn each one. There's no equivalent of `afx workspace recover` for them, and the existing recovery command excludes architects by predicate (`type === 'builder'` filter on session rows, and porch state is the entry point — architects don't have porch projects).

## Proposed mechanism

Reuse the on-disk Claude conversation discovery from #829: each architect's prior conversation lives at `~/.claude/projects/<encoded-workspace-cwd>/<uuid>.jsonl`. After reboot, `findLatestSessionId(workspacePath)` returns the most recent UUID; we relaunch the architect via `claude --resume <uuid>` to restore the conversation in addition to the process.

Two architect kinds, both Tower-managed:

1. **Main architect** — one per active workspace, spawned by `afx workspace start`. Its cwd is the workspace root.
2. **Named sibling architects** (Spec 755) — additional architects added via `afx workspace add-architect [--name <name>]`. Each is tracked by a name; the spec docs in `codev/specs/755-multi-architect.md` (or wherever) detail naming rules. Their cwd may also be the workspace root, or may be a per-architect subdirectory — needs verification.

Sketch of a new command (or extension of the existing one):

```
afx workspace recover-architects [--apply] [-y]
```

Or, more unified:

```
afx workspace recover [--builders] [--architects] [--apply] [-y]
```

Either reuses `findLatestSessionId()` from `agent-farm/utils/claude-session-discovery.ts`.

## Open design questions

[...]

PR #833, opened 2026-05-23
[#829] afx workspace recover: revive builders and main architect after machine reboot

### LINKED ISSUE #829: afx workspace recover: revive builders after machine reboot/crash
## Problem

After a machine reboot or crash, all builder processes die — shellper daemons (detached, in-memory) and Tower together. Disk-persistent state survives: worktrees in `.builders/<id>/`, porch `status.yaml`, SQLite `terminal_sessions`, spec/plan/review files. But there is no command to walk that state and respawn the builders that should still be running.

The user typically discovers a half-dozen "dead" builders the morning after a reboot and has to `afx spawn <id> --resume` each one by hand. Most builders are sitting idle at approval gates when this happens, so a fresh respawn (which lands them right back at the gate via `status.yaml`) is functionally equivalent to true session resume.

## Proposal

Add an explicit `afx workspace recover` command (no auto-trigger on Tower startup) that:

1. Enumerates `codev/projects/*/status.yaml` (using `findStatusPath()` to handle spec-653 multi-PR layouts under `.builders/<id>/codev/projects/`).
2. Filters to eligible projects via the predicate below.
3. By default, runs in **dry-run mode**: lists what it would revive and exits. Pass `--apply` to actually respawn.
4. For each eligible project, respawns the builder via the existing `afx spawn <id> --resume` codepath.

## Revival predicate

A project is eligible when ALL of:

1. `phase ∉ {verified, complete}` — not in a terminal porch state.
2. A `terminal_sessions` row exists for this project (it had a builder previously).
3. The shellper PID is dead OR the Unix socket is unreachable (actually needs recovery, not already alive).
4. The worktree at `.builders/<id>/` still exists on disk.
5. `status.yaml.updated_at` is within the last **7 days** (configurable via `--max-age <days>`; bypass entirely with `--include-stale`).

Notes:
- Builders idle at the `pr` gate **are** revived (consistent with all other gate-idle builders; the resource cost is negligible and the human may want to dispatch the builder to address review feedback without manual respawn).
- Projects whose `status.yaml` was never written (builder crashed before first phase transition) won't show up; this is the conversation-resume gap and is out of scope.

## CLI

```
afx workspace recover           # dry-run, prints eligible projects
afx workspace recover --apply   # actually respawn
afx workspace recover --max-age 14
afx workspace recover --include-stale --apply
```

## Out of scope (separate issue)

- **Conversation resume** — restoring the builder's prior Claude session content (the `~/.claude/projects/<encoded-cwd>/<uuid>.jsonl` file). Requires capturing the Claude session UUID at spawn time, storing it in `terminal_sessions`, and passing `--resume <uuid>` to the relaunched `claude` process. Worth it for mid-phase recovery but materially larger scope; file separately if/when mid-phase crashes become a real pain point.
- **Auto-trigger on Tower startup** — explicitly rejected here to avoid reviving abandoned-but-uncleaned projects. Revisit if explicit-trigger friction proves high.

## Implementation notes

- Reuse `findStatusPath()` (state.ts:285–303) for status.yaml lookup.
- Liveness check: SQLite `terminal_sessions` row → PID alive + socket connectable. Reconciliation logic in `tower-terminals.ts:485–711` already detects dead sockets; recovery extends that path with respawn instead of just cleanup.
- Respawn path: invoke the same flow as `afx spawn <id> --resume` rather than building a parallel codepath.

### LINKED ISSUE #831: Builder conversation resume: restore Claude session on recovery via jsonl discovery
## Context

Extends #829 (`afx workspace recover` — process revival for builders after machine reboot). #829's initial scope was *process* revival only: respawn a fresh `claude` process in the right worktree and let porch state bridge the protocol-level continuity. Conversation context — what the builder was thinking, what was tried, what feedback the user had given — was explicitly out of scope and acknowledged as a "larger separate issue."

This issue covers the conversation-resume work that was ultimately bundled into the same PR. Tracked separately so the design and implementation history are easy to find.

## Problem

After process revival, the respawned builder started with no memory of its prior session. Useful at approval gates (porch's external memory bridges the gap), but materially worse for mid-phase crashes — the builder had to reconstruct intent from the worktree diff and protocol state, often re-deriving conclusions or undoing partial work.

## Mechanism

Claude Code automatically persists every interactive session to
```
~/.claude/projects/<encoded-cwd>/<session-uuid>.jsonl
```
where the encoding replaces both `/` and `.` in the absolute cwd with `-`. The data for every prior builder session is on disk; we just hadn't been looking.

On `--resume`, spawn.ts now:
1. Computes the encoded directory for the worktree
2. Picks the most-recently-modified `*.jsonl` by mtime
3. Passes the UUID through to `startBuilderSession`
4. The launch script runs `claude --resume <uuid>` (no role injection, no initial prompt — both are already part of the saved conversation)

Backwards-compatible by design: existing builders (spawned before this code shipped) already have their jsonl on disk and resume cleanly on the first revival.

If no jsonl exists (truly fresh worktree, deleted history, etc.), spawn falls back to the prior behavior — fresh session with a generic resume-notice prompt.

## Why jsonl discovery over a sidecar

Discussed at length in the #829 conversation. Trade-offs in short:

- **No spawn-time state to maintain.** Claude writes the jsonl as a side effect of the original session; we only read it. No sidecar file, no SQLite column, no schema migration.
- **Backwards compat for free.** Existing builders work the first time the new code runs because their jsonl was already on disk.
- **One mechanism, not two.** Same helper extends to architects (single-architect workspaces; multi-architect ambiguity is a known limitation tracked separately in #830).

The trade-off: when multiple jsonl files live in the same encoded-cwd directory (a worktree was reused for unrelated work), the newest wins by mtime. For builder worktrees — protocol-scoped, typically host one logical conversation — this is right nearly always. The `Claude session: <prefix>… (resuming conversation)` log line surfaces which UUID was picked so any mismatch is visible.

## Scope

- **In:** Builder conversation resume on the `afx spawn --resume` path. Triggered transitively by `afx workspace recover`. Backwards-compatible for existing builders.
- **Out:** Architect conversation resume (covered by #830). Multi-architect disambiguation. Per-builder session UUID storage / sidecar approaches.

## Reference

- #829 — original `afx workspace recover` feature, where this conversation-resume implementation landed.
- Helper module: `packages/codev/src/agent-farm/utils/claude-session-discovery.ts`.
- Threading: `packages/codev/src/agent-farm/commands/spawn.ts` (read jsonl on `--resume`), `packages/codev/src/agent-farm/commands/spawn-worktree.ts` (different launch script when `resumeSessionId` is provided).
- Tests: 9 unit tests in `packages/codev/src/agent-farm/__tests__/claude-session-discovery.test.ts` covering encoding, missing dir, no-jsonl dir, mtime selection, and ignored non-jsonl entries.

amrmelsayed:
## Summary

- New `afx workspace recover` command — enumerates porch projects, identifies builders whose shellper died (e.g., after machine reboot), and respawns them via `afx spawn <id> --resume`. Dry-run by default; `--apply` actually respawns.
- Builder conversation resume for **all** revivable protocols (SPIR/ASPIR/AIR/PIR/bugfix) — the revived `claude` process picks up the prior session via `--resume <uuid>`, discovered from the on-disk jsonl store. Works retroactively for builders spawned before this code shipped.
- Main architect conversation resume after reboot, via the same jsonl-discovery mechanism.

## Scenario
User spawns multiple builders, OS then restarts or crashes which inherintly terminates all running daemons, resulting in loss of development state and numerous dangling worktrees and branches.

## Issues

- Closes #829 — process revival for builders
- Closes #831 — builder conversation resume
- Partially addresses #830 — main architect resume shipped here; named siblings deferred to #832
- Coordinates with #832 (assigned to @waleedkadous) — multi-architect resume

## Known limitation: multi-architect workspaces

**Main architect conversation resume is intentionally disabled when more than one architect is persisted in `state.db`** (single workspace with main + Spec-755 named siblings). All architects in a workspace share `cwd = workspacePath`, so their jsonl files mix in the same `~/.claude/projects/<encoded-workspace>/` directory and the newest-by-mtime heuristic can't disambiguate them. Without a guard, main could attach to a sibling's conversation on revival.

The conservative fix shipped here: when `getArchitects().length > 1`, `launchInstance` skips the resume path for main and spawns fresh with role injection. A WARN log surfaces what happened with a reference to #832. Single-architect workspaces (the common case) keep full conversation resume. Multi-architect workspaces lose resume entirely until #832's per-architect UUID storage lands.

## Three rounds of cmap-3 review

| Round | Verdicts | Outcome |
|---|---|---|
| 1 | 1 approve, 2 request-changes | Four blockers fixed (predicate liveness, protocol allowlist, duplicate-row aggregation, process.execPath) |
| 2 | 2 approve-with-nits, 1 request-changes | Real session-row-existence bug discovered live against shannon, fixed via predicate inversion; conversation-resume scope expanded (#831 + #830 main path) |
| 3 | 1 approve, 2 approve-with-nits | Helper extraction + multi-architect guard nits; JSDoc cleanup; new unit tests for `discoverResumeSession` |

## Commits

```
4e4d68f1 fix: cmap-3 round-3 nits
79eb2963 fix: address cmap-3 round-2 review findings
b37c6eb2 feat(tower): resume main architect's Claude conversation after reboot
b96dcacc fix(recover): round age display UP so labels match --max-age
c6c0dc98 feat(recover): conversation resume via on-disk jsonl discovery
386353d4 feat(recover): hide stale items from dry-run by default
c14fd0d1 fix: invert session-row predicate — absence means revive
69d65d37 fix: address cmap-3 review findings
d14ee52e feat(afx): add 'workspace recover' to revive builders after reboot
96aa11ee feat(porch): add listAllProjects helper
```

## Eligibility predicate

A project is revived when ALL of:
1. Phase is not terminal (`verified` / `complete`)
2. Protocol is revivable (`spir`, `aspir`, `air`, `pir`, `bugfix`)
3. Worktree exists on disk with a `.git` marker
4. No matching `terminal_sessions` row's shellper is alive
5. Within `--max-age` (default 7d), unless `--include-stale`

PR-gate-idle builders are deliberately included.

## CLI

```bash
afx workspace recover                       # dry-run preview (default)
afx workspace recover --apply               # actually respawn (with confirmation)
afx workspace recover --apply --yes         # unattended
afx workspace recover --max-age 30          # widen the recency window
afx workspace recover --include-stale       # show & include everything
```

## Test plan

- [x] `pnpm exec vitest run src/agent-farm/__tests__/{workspace-recover,claude-session-discovery,discover-resume-session}.test.ts` — 60/60 pass
- [x] `pnpm exec tsc --noEmit` — clean
- [x] Live test against shannon (345-project workspace): 9 in-flight PIR builders correctly surfaced as `revive`; rest correctly classified across `terminal` / `unsupported_protocol` / `worktree_missing` / `stale`
- [x] Three rounds of cmap-3 review (gemini + codex + claude × 3); final verdicts 1×APPROVE, 2×APPROVE_WITH_NITS

## Cmap reviewer's note for Codex's third blocker

Codex flagged that a porch project explicitly `afx cleanup`'d-but-worktree-preserved would still be eligible for revival. Claude and Gemini both read the current behavior as correct: the dry-run preview + explicit `--apply` confirmation is the safety net, and a preserved worktree IS the signal that the work might be resumable. Not addressed in this PR; a `cleaned_up` predicate would require an extra DB lookup just to handle an edge case that has explicit user-consent gates around it.

## Out of scope (tracked elsewhere)

- Multi-architect conversation resume — tracked in #832 (Waleed)
- Auto-trigger on Tower startup — rejected during design
- The 7 pre-existing TS errors in `status.ts` / `workspace-remove-architect.ts` that were present on main before this branch


waleedkadous (approved):
Great work on this — recovery from machine reboots has been the missing piece in Tower's "leave it running for days" story. The layered design (`claude-session-discovery` as a pure utility, `workspace-recover` as a thin pipeline over `evaluateEligibility` → `respawnBuilder`, resume path threaded through both `spawn.ts` and `tower-instances.ts`) makes the integration feel small for what's actually a large capability. The 60 tests with the cheap-first predicate-order test as a design-intent guard is exactly the kind of thing I hope reviewers will see and rarely do.

I ran a 3-way architect-side CMAP. Two approves, one request-changes — Codex flagged that the respawn uses `process.argv[1]`, which breaks when the command is invoked as `codev afx workspace recover` (the codev wrapper exposes `afx` as a hidden subcommand at `cli.ts:355-356`). We're going to handle that by **deprecating the `codev afx` / `codev agent-farm` invocation paths** in a separate change rather than patching the spawn child — those wrapper paths have outlived their usefulness and add exactly this kind of cross-cutting fragility. So Codex's blocker becomes moot once that lands, and this PR doesn't need to wait on it.

Two small things you might consider while the PR is hot, but neither is blocking:

- `getClaudeProjectDir` in `claude-session-discovery.ts` is exported but no one imports it — would be nice to either wire a consumer or drop the export.
- In `tower-instances.ts`, the `harnessEnv = {}` on the architect resume path is correct but uncommented; the builder-side equivalent in `spawn-worktree.ts` has a one-liner explaining why it's safe to skip role injection. Adding the same line on the architect side would prevent the next reader from second-guessing it.

Approving as-is. Please merge when ready.

PR #827, opened 2026-05-23
[Bugfix #826] Scope launchInstance sibling reconcile by workspace_path

waleedkadous:
## Summary

Fixes #826 — sibling architects registered in workspace A were re-spawned into workspace B at launch time, leaking PTYs across workspaces.

## Root cause

Spec 786 added a reconcile loop in `launchInstance` that iterates `state.db.architect` (global per Tower CWD) and re-spawns persisted non-main siblings via `addArchitect`. The architect table has no `workspace_path` column, so launching workspace B while workspace A held siblings caused B to inherit A's siblings as fresh PTYs.

Verified against the issue's diagnosis: `state.db.architect` is anchored to Tower's CWD; `terminal_sessions` (in `global.db`) already carries `workspace_path` for every architect row.

## Fix

Option B per the issue (no schema migration):

- **`state.ts`** — new `getArchitectsForWorkspace(workspacePath)`. Reads architect names from `terminal_sessions WHERE type='architect' AND workspace_path = ?` in `global.db`, then intersects with `state.db.architect` rows by name. Two queries, no SQLite cross-DB attach.
- **`tower-instances.ts`** — the reconcile loop now calls `getArchitectsForWorkspace(resolvedPath)` instead of the unscoped `getArchitects()`. The unscoped import is dropped.

## Known trade-off

`deleteWorkspaceTerminalSessions` wipes all `terminal_sessions` rows for a workspace on `afx workspace stop`. After `stop + start`, the workspace_path signal that this filter relies on is gone, so persisted siblings in `state.db.architect` will **not** be re-spawned on next start — a partial regression of Spec 786's stop+start sibling persistence story.

The active cross-workspace leak is far more severe (silent PTY proliferation, ~50MB+ per leaked claude session, potential workflow confusion) than the stop+start regression (a developer-only command, easily worked around by re-running `afx workspace add-architect`). Option A (`state.db.architect` schema migration adding `workspace_path`) is the proper long-term fix and can ship as a follow-up.

## Test plan

- [x] 5 new unit tests in `state.test.ts` for `getArchitectsForWorkspace` (workspace isolation, multi-workspace partitioning, non-architect row filtering, orphan handling, regression scenario for #826).
- [x] Source-level sentinel test in `tower-instances.test.ts` that locks in the scoped call site and prevents the unscoped `getArchitects` from re-creeping in via import.
- [x] `state.test.ts` 27/27 pass; `tower-instances.test.ts` 52/52 pass.
[...]

PR #834, opened 2026-05-23
[Bugfix #826] iter-7: Fix release blocker — LOCAL_SCHEMA CREATE INDEX vs migration order

### LINKED ISSUE #826: CRITICAL: Sibling architects leak across workspaces — launchInstance reconcile reads global state.db.architect without workspace filtering
## Summary

In v3.1.1, opening a workspace **other than the one that registered sibling architects** causes those siblings to be re-spawned in the new workspace. Sibling architects leak across workspaces.

Reported by user immediately after v3.1.1 installation: creating workspace `manazil` showed `bug-backlog` and `ob-refine` architects from `shannon` running in manazil's terminal list (with NEW PIDs, meaning they were actually spawned, not just listed).

## Repro

1. Tower is started from `/Users/mwk/Development/cluesmith/codev`
2. Shannon adds sibling architects: `afx workspace add-architect --name ob-refine` (in shannon)
3. Shannon's state.db.architect row gets written, BUT the row lives in `codev/.agent-farm/state.db` (the singleton from `getDb()`, anchored to Tower's CWD)
4. User opens manazil workspace
5. `launchInstance(manazil)` runs
6. `launchInstance` iterates `state.db.architect` (the global codev-side table) and finds `ob-refine`
7. `launchInstance` re-spawns ob-refine **as a manazil architect** with a new PID
8. Manazil's `/api/state` now reports `architects: [main, ob-refine, bug-backlog]` — all running, all real PTYs, none of them legitimately belonging to manazil

## Diagnosis (verified)

```
$ sqlite3 /Users/mwk/Development/cluesmith/codev/.agent-farm/state.db "SELECT id FROM architect;"
bug-backlog
main
ob-refine

$ sqlite3 /Users/mwk/Development/cluesmith/shannon/.agent-farm/state.db "SELECT id FROM architect;"
(empty — shannon's siblings live in codev/.agent-farm/state.db because state.db is anchored to Tower's CWD)

$ sqlite3 /Users/mwk/Development/cluesmith/codev/.agent-farm/state.db ".schema architect"
CREATE TABLE IF NOT EXISTS "architect" (
    id TEXT PRIMARY KEY,
    pid INTEGER NOT NULL,
    port INTEGER NOT NULL,
    cmd TEXT NOT NULL,
    started_at TEXT NOT NULL DEFAULT (datetime('now')),
    terminal_id TEXT
);
```

No `workspace_path` column. The table is **global per Tower daemon**, not per workspace.

Meanwhile `terminal_sessions` (in `~/.agent-farm/global.db`) DOES have a `workspace_path` column:

```
CREATE TABLE terminal_sessions (
    id TEXT PRIMARY KEY,
    workspace_path TEXT NOT NULL,
    type TEXT NOT NULL CHECK(type IN ('architect', 'builder', 'shell')),
    role_id TEXT,
    ...
)
```

So the workspace-scoping data exists; it's just not joined into the architect reconcile path.

## Why this is new in v3.1.1

This issue was flagged by Codex at #786 plan-iter-3 Co1 ("workspace-scoping") and explicitly accepted-as-out-of-scope by the architect because Spec 786 defers cross-workspace concerns:
> "Cross-workspace routing. Architects in workspace A cannot address architects in workspace B. Deferred previously; stays deferred."

But the bug ISN'T about routing — it's about **persistence leaking architects across workspaces** via the launchInstance reconcile loop that #786 added to deliver graceful-stop persistence. Pre-#786, `launchInstance` only created `main`; the global table was harmless because nothing iterated it. #786's new iterate-and-respawn loop turns the pre-existing global table into an active cross-workspace leak.

## Fix shape

Two options:

### Option A (proper, schema migration)
Add `workspace_path TEXT NOT NULL` column to `state.db.architect`. Migrate existing rows by joining with `terminal_sessions` on `id = role_id` to populate `workspace_path`. Update `setArchitect`, `setArchitectByName`, `removeArchitect`, `loadState`, and the launchInstance reconcile to scope by workspace. Most correct but a schema migration with multiple touch points.

### Option B (quick, uses existing data)
Modify the launchInstance reconcile loop to **only re-spawn architects that have a matching `terminal_sessions` row WITH `workspace_path = <this workspace>`**. The `terminal_sessions` table already has the workspace scoping; join on `terminal_sessions.role_id = architect.id AND terminal_sessions.type = 'architect' AND terminal_sessions.workspace_path = <this workspace>`. No schema change; reconcile is correctly scoped.

Recommend **Option B for v3.1.2** (urgent hotfix). Option A is the right long-term fix and can ship as part of a follow-up architectural cleanup.

## Severity

**Critical.** v3.1.1 just shipped (~30 min ago). Anyone with multiple workspaces and a non-main sibling architect anywhere will see leaks the moment they open a different workspace. The leaked architects are real running processes consuming resources (~50MB+ RAM per claude session, plus their `claude --dangerously-skip-permissions` shell harnesses) AND they could interfere with the actual workspace's workflow if the user accidentally talks to them.

## Workaround until fix lands

Don't open additional workspaces in Tower if you have non-`main` siblings registered anywhere. Or kill the leaked terminals via the dashboard sidebar X button (which kills the PTY but leaves the state.db row intact for the legitimate workspace).

## Cleanup of leaked architects

Direct `afx workspace remove-architect <name>` from a leaked workspace would also delete the state.db row, which corrupts the legitimate owner workspace's state. Per-PTY kill via the dashboard is the safer manual cleanup until the proper fix lands.

## Related

- #786 (introduced the reconcile loop that exposes the bug)
- #823 (now-shipped multi-architect coordination, also affected)
- Codex iter-3 Co1 at #786 plan phase originally flagged this; accepted out of scope at the time.

waleedkadous:
## Summary

Follow-up to merged PR #827 (Fixes #826) — fixes a release blocker discovered during local manazil verification: every v3.1.1 user's upgrade would have crashed at startup.

## Root cause

`ensureLocalDatabase` runs `db.exec(LOCAL_SCHEMA)` BEFORE migrations on every open. PR #827's `LOCAL_SCHEMA` contained:

```sql
CREATE INDEX IF NOT EXISTS idx_architect_workspace ON architect(workspace_path);
```

On any pre-v11 install, the `architect` table lacks the `workspace_path` column. The CREATE INDEX threw `no such column: workspace_path` and aborted `ensureLocalDatabase` before migration v11 could run. The schema stayed at v10 and every subsequent `setArchitectByName` call failed with the same error.

Architect caught it on the user's actual `codev/.agent-farm/state.db`:
> [WARN] Failed to persist architect verify-826-test to state.db: no such column: workspace_path

Fresh installs would have worked (LOCAL_SCHEMA's `CREATE TABLE` would have created the v11 shape from scratch) — but upgrade installs would not.

## Fix

1. **Drop the CREATE INDEX statement from `schema.ts` LOCAL_SCHEMA.**
2. **Add it to `db/index.ts` migration v11 block, OUTSIDE the inner `if (!alreadyMigrated)` guard but inside the outer `if (!v11)` guard.**
   - **Upgrade install**: inner block runs (architect → architect_v11 rename); then `CREATE INDEX IF NOT EXISTS` creates the index on the new table.
   - **Fresh install**: LOCAL_SCHEMA created the v11-shaped table; inner block is a no-op (workspace_path column already present); CREATE INDEX still runs in the outer block.
3. `IF NOT EXISTS` makes subsequent opens (after v11 is recorded in `_migrations`) a no-op.

## Tests

3 new in `bugfix-826-migration.test.ts`:

1. **`db.exec(LOCAL_SCHEMA)` against a pre-v11 architect table no longer throws.** Pre-iter-7 this test would have failed with `no such column`; it's the literal release-blocker repro.
2. **Full upgrade sequence**: v10 schema → LOCAL_SCHEMA → migrate. Asserts the table is migrated AND the index exists.
3. **Fresh-install path**: LOCAL_SCHEMA creates v11 shape → migration inner block skipped → CREATE INDEX still runs in the outer block. Asserts the index exists.

The existing migration tests didn't catch this because they set up the v10 schema directly in SQL and called the migration body — they never exercised LOCAL_SCHEMA's `db.exec` sequence, where the bug lived. Architect explicitly called out this test gap.

## Test results

- 1853 agent-farm unit tests pass.
- `pnpm exec tsc --noEmit` clean.

## Diff stat

3 files, +136 / -3.

## CMAP

Will run 3-way after PR creation per the BUGFIX protocol.

### DIFF
--- a/packages/codev/src/agent-farm/__tests__/bugfix-826-migration.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/bugfix-826-migration.test.ts
@@ -151,9 +151,12 @@ describe('Bugfix #826 — Migration v11 (workspace-scoped architect schema)', ()
       localDb.exec(`
         DROP TABLE architect;
         ALTER TABLE architect_v11 RENAME TO architect;
-        CREATE INDEX IF NOT EXISTS idx_architect_workspace ON architect(workspace_path);
       `);
     }
+
+    // iter-7: create the index outside the alreadyMigrated guard so both
+    // upgrade and fresh paths converge. Mirrors production.
+    localDb.exec('CREATE INDEX IF NOT EXISTS idx_architect_workspace ON architect(workspace_path);');
 
     localDb.prepare('INSERT INTO _migrations (version) VALUES (11)').run();
   }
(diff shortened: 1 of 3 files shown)


waleedkadous:
## iter-7 CMAP 3-way result: unanimous APPROVE

- **Gemini** (147.9s): approved. Verified the full upgrade sequence correctly adds the column and creates the index; verified the fresh-install sequence correctly skips the inner migration but still creates the index.
- **Codex** (51.0s): approved. `KEY_ISSUES: None`.
- **Claude** (73.9s): approved. Confirmed 1853 unit tests pass and `tsc --noEmit` clean.

No REQUEST_CHANGES, no flagged concerns. The release blocker is closed for both upgrade and fresh-install paths.

@waleedkadous — ready for re-merge.

Issue #844, opened 2026-05-24
Needs Attention surfaces PRs before CMAP reviews finish


waleedkadous:
## The bug

When a builder creates a PR, it appears in the **Needs Attention** list immediately — *before* the automated CMAP reviews (Gemini / Codex / Claude) have run and posted their findings. By the time the human looks at it, the review comments may not be there yet, so they either:

- click through, see no review activity, and bounce back to wait, or
- review the diff themselves without the AI consultation findings they were expecting

The PR should only appear in Needs Attention **after** the automated reviews are complete and the builder has hit the porch `pr` gate (which is the actual "human, please review and merge" signal).

## Where it lives

`packages/dashboard/src/components/NeedsAttentionList.tsx:42-56` builds the list by iterating *every* open PR returned in `overview.pendingPRs`:

```ts
function buildItems(prs: OverviewPR[], builders: OverviewBuilder[]): AttentionItem[] {
  const items: AttentionItem[] = [];

  // PRs needing review
  for (const pr of prs) {
    items.push({ ... kind: 'PR review', ... });
  }
  ...
}
```

There is no filter for whether automated review has run. The upstream `overview.pendingPRs` (in `packages/codev/src/agent-farm/servers/overview.ts:836-852`) is built straight from `fetchPRList` — all open PRs, no porch-state cross-reference.

## The right signal

Porch already tracks the `pr` gate. When CMAP finishes in the `review` phase, the builder advances to and blocks on the `pr` gate (label: `\"PR review\"`). That's the moment the PR is genuinely waiting on a human.

The existing dedupe at `NeedsAttentionList.tsx:61-62` already knows about this signal — it just uses it in the wrong direction:

```ts
// Skip \"PR review\" — those are already covered by the PRs list
if (b.blocked === 'PR review') continue;
```

The fix is to **invert the source of truth**: include a PR in Needs Attention only when its corresponding builder is blocked on the `pr` gate. PRs with no matching builder, or whose builder is still running CMAP, are excluded.

## Suggested approach

1. In `NeedsAttentionList.tsx`, build a set of issue IDs from `builders.filter(b => b.blocked === 'PR review')` and only include PRs whose `pr.issueId` (or the equivalent linkage field — verify in `OverviewPR`) is in that set.
2. Drop the now-redundant `continue` at line 62 — the dedupe is no longer needed because the PR list is gated by the same condition.
[...]

PR #874, opened 2026-05-26
[Bugfix #872] Fix: emit canonical pr_ready_for_human signal across all protocols

### LINKED ISSUE #872: porch: emit explicit pr_ready_for_human state field when CMAP completes
## Context

PR #844 (shipped in v3.1.3) tried to gate the dashboard's Needs Attention list on "CMAP done, PR ready for human review". The check it used — `b.blocked === 'PR review'` — was a derivation that worked for protocols with an explicit `pr` gate (SPIR/ASPIR/PIR/AIR) but silently failed for BUGFIX (no `pr` gate, transitions straight to `phase: verified` after CMAP).

The underlying issue is that **"CMAP complete for the PR-creating phase" has one true definition in porch's state machine but two different surface manifestations** today, depending on whether the protocol's PR-creating phase carries a gate. Anyone consuming "is this PR ready for a human" has to know the protocol-specific derivation, and the only way to find out you got it wrong is to ship a regression (which we did).

This issue proposes making the signal canonical: porch writes an explicit `pr_ready_for_human: true` to `status.yaml` when it transitions out of the CMAP-emitting state for the PR-creating phase, regardless of whether the protocol uses a gate. Consumers (Needs Attention filter, VS Code extension, future dashboards) read one boolean.

A parallel bug ([linked issue for the B path]) covers the immediate filter-side patch in overview.ts as a fast fix; this issue tracks the canonical state-machine change.

## Design questions for the spec/plan

These are the questions that motivate using **PIR** rather than BUGFIX/AIR — they want pre-PR review:

1. **Field name and shape.** `pr_ready_for_human: boolean` is the narrowest framing. Broader alternatives: `cmap_complete: boolean` (covers any CMAP-emitting phase, not just PR phases) or a typed enum `pr_state: 'no-pr' | 'cmap-running' | 'cmap-complete' | 'merged'`. Each has different consumer implications.

2. **Per-phase vs single boolean.** Some protocols have multiple CMAP-emitting phases (SPIR's `specify`, `plan`, `review`). Do we set the flag only for the PR-creating phase, or maintain a per-phase map? The Needs Attention use case wants only "PR-creating phase done", so per-phase may be overkill — but other consumers might want richer state.

3. **Re-iter handling.** When CMAP returns `REQUEST_CHANGES` and porch enters the rebuttal cycle, does `pr_ready_for_human` go back to false until the next iteration's CMAP completes? (Yes — it should reflect current state, not "ever was complete.")

4. **Identification of the CMAP-emitting phase.** Today the consultation block lives in the phase definition (e.g. BUGFIX's `pr` phase has `consultation.on: 'review'`). Does porch infer the PR-creating phase from the consultation block presence, or do we add an explicit `is_pr_phase: true` marker to protocol.json? The former is cheaper and works retroactively; the latter is more explicit and lets a protocol have CMAP without it being PR-creating.

5. **Backward compat.** Existing in-flight projects (status.yaml without the new field) need a fallback. The simplest: if the field is absent, fall back to the current derived logic (`blocked === 'PR review' || (phase === 'verified' && has_open_pr)`). This lets us ship the canonical signal without forcing a migration on every running builder.

6. **Surface change in `OverviewBuilder`.** Add `prReady: boolean` derived from the new status.yaml field (with the v3.1.3 fallback). The Needs Attention filter consumes only this field.

## Acceptance criteria

- [ ] `status.yaml` gains a `pr_ready_for_human: boolean` field (or whichever shape the plan settles on).
- [ ] Porch sets it to `true` when transitioning out of the CMAP-emitting state for the PR-creating phase, for ALL protocols (SPIR, ASPIR, PIR, AIR, BUGFIX).
- [ ] Porch sets it back to `false` (or removes the field) when entering the rebuttal cycle on `REQUEST_CHANGES`.
- [ ] `OverviewBuilder.prReady` is computed from this field with a fallback for legacy state files lacking the field.
- [ ] `NeedsAttentionList.tsx` gates on `prReady` alone, no protocol-specific derivation.
- [ ] Tests: each of the five protocols gets a porch-state test verifying the field's lifecycle.
- [ ] Migration note: in-flight projects on v3.1.x continue working (fallback to current derived logic until they reach the CMAP-done state under the new code).

## References

- The regression that motivated this: PR #844 / PR #847 / issue #870-adjacent (Needs Attention shipped excluding BUGFIX PRs in v3.1.3)
- Existing derivation in `packages/dashboard/src/components/NeedsAttentionList.tsx:42-100`
- Porch state machine: `packages/codev/src/commands/porch/next.ts`
- BUGFIX protocol (no `pr` gate): `codev-skeleton/protocols/bugfix/protocol.json`
- AIR protocol (has `pr` gate): `codev-skeleton/protocols/air/protocol.json`

waleedkadous:
## Summary
Fixes #872.

PR #844 (v3.1.3) gated Needs Attention on `b.blocked === 'PR review'`, a protocol-specific derivation that worked for AIR/SPIR/ASPIR/PIR but silently dropped BUGFIX PRs (BUGFIX has no `pr` gate — it transitions straight to `phase: verified` after CMAP). This change makes the signal canonical: porch writes `pr_ready_for_human: boolean` to `status.yaml` when transitioning out of the CMAP-emitting state for the PR-creating phase, for ALL five bundled protocols.

## Root Cause
"CMAP complete for the PR-creating phase" had one definition in porch's state machine but two surface manifestations depending on whether the protocol had a `pr` gate. Consumers had to know the protocol-specific shape, and the only way to discover you got it wrong was to ship a regression — which #844 did.

## Fix

**State machine (3 touchpoints, one canonical write each):**
- `next.ts` `handleVerifyApproved` — SPIR/ASPIR/PIR `review` (build_verify, gate=pr): set true at gate-pending
- `index.ts` `done()` — AIR `pr` (once-phase, gate=pr): set true when auto-requesting the `pr` gate
- `index.ts` `advanceProtocolPhase` — BUGFIX `pr` (once-phase, no gate, terminal): set true on transition to `verified`

**PR-creating phase identification:** new `isPrCreatingPhase(protocol, phaseId)` helper returns true when the phase has `gate === 'pr'` OR a `consultation` block in protocol.json (preserved on `ProtocolPhase.hasConsultation`). Both markers are picked up retroactively — no protocol.json edits required.

**Reset to false** on `pr`-gate approval (covers all four gate-protocols) and on rollback. BUGFIX has no gate, so the field stays true until cleanup deletes the worktree — matching the model that a BUGFIX builder is done after PR creation.

**Consumer (`overview.ts`):** new `derivePrReady` reads the explicit field if present, else falls back to `blocked === 'PR review' || (phase === 'verified' && protocol === 'bugfix')`. The BUGFIX-specific fallback closes the v3.1.3 gap for in-flight builders whose status.yaml pre-dates this field. New `prReady` field on `BuilderOverview` exposes the signal.

**Dashboard (`NeedsAttentionList.tsx`):** gates on `b.prReady` instead of `b.blocked === 'PR review'`. Falls back to `pr.createdAt` for prReady builders without a gate (BUGFIX shape — no `blockedSince`).

## Test Plan
- [x] **8 new tests** in `pr-ready-872.test.ts` covering the field lifecycle for SPIR/ASPIR/PIR (build_verify review), AIR (once-phase pr with gate), BUGFIX (once-phase pr without gate — the regression case), the REQUEST_CHANGES rebuttal cycle, pr-gate approval reset, and rollback reset.
- [x] **6 new tests** in `overview.test.ts` for `derivePrReady`: explicit-true, explicit-false (trumps fallback), legacy SPIR/AIR fallback, BUGFIX gap closure, and non-BUGFIX `verified` no-fire.
- [x] **2 new tests** in `NeedsAttentionList.test.tsx` for BUGFIX visibility (the regression) and `blockedSince`→`createdAt` fallback for gateless prReady builders.
- [x] All 3152 porch+codev tests pass (148 files); all 17 dashboard component tests pass.
- [x] Build clean (`pnpm build`).
- [x] Backward compat: in-flight projects on v3.1.x continue working — the `OverviewBuilder.prReady` fallback derives the same signal until porch writes the explicit field.

## Files
- **Field + helpers**: `packages/codev/src/commands/porch/{types,protocol,next,index}.ts`
- **Consumer**: `packages/codev/src/agent-farm/servers/overview.ts`, `packages/types/src/api.ts`, `packages/dashboard/src/components/NeedsAttentionList.tsx`
- **Tests**: `packages/codev/src/commands/porch/__tests__/pr-ready-872.test.ts` (new), `packages/codev/src/agent-farm/__tests__/overview.test.ts`, `packages/dashboard/__tests__/NeedsAttentionList.test.tsx`
- Two existing test fixtures gained `prReady: false` to satisfy the new shared type (`BuilderCard.test.tsx`, `spec-823-builder-attribution.test.ts`)

## Notes
- One pre-existing test failure in `packages/dashboard/__tests__/scrollController.test.ts` (terminal scroll behavior) — completely unrelated to this work and fails identically on this branch in isolation.
- Net diff: ~290 LOC added across implementation + tests; well under the 300 LOC BUGFIX threshold for the implementation alone.

### DIFF
--- a/codev/projects/bugfix-872-porch-emit-explicit-pr-ready-f/status.yaml
+++ b/codev/projects/bugfix-872-porch-emit-explicit-pr-ready-f/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-872
 title: porch-emit-explicit-pr-ready-f
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-26T21:14:25.066Z'
-updated_at: '2026-05-26T21:39:52.686Z'
+updated_at: '2026-05-26T21:40:24.813Z'
--- a/codev/projects/bugfix-872-porch-emit-explicit-pr-ready-f/status.yaml
+++ b/codev/projects/bugfix-872-porch-emit-explicit-pr-ready-f/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-872
 title: porch-emit-explicit-pr-ready-f
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-26T21:14:25.066Z'
-updated_at: '2026-05-26T21:14:25.066Z'
+updated_at: '2026-05-26T21:39:52.686Z'
--- a/codev/projects/bugfix-872-porch-emit-explicit-pr-ready-f/status.yaml
+++ b/codev/projects/bugfix-872-porch-emit-explicit-pr-ready-f/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-872
 title: porch-emit-explicit-pr-ready-f
 protocol: bugfix
-phase: pr
+phase: verified
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-26T21:14:25.066Z'
-updated_at: '2026-05-26T21:40:24.813Z'
+updated_at: '2026-05-26T21:40:32.629Z'
--- a/codev/projects/bugfix-872-porch-emit-explicit-pr-ready-f/status.yaml
+++ b/codev/projects/bugfix-872-porch-emit-explicit-pr-ready-f/status.yaml
@@ -1,0 +1,12 @@
+id: bugfix-872
+title: porch-emit-explicit-pr-ready-f
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates: {}
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-05-26T21:14:25.066Z'
+updated_at: '2026-05-26T21:14:25.066Z'
--- a/codev/state/bugfix-872_thread.md
+++ b/codev/state/bugfix-872_thread.md
@@ -47,3 +47,17 @@ This restores BUGFIX visibility for in-flight projects that pre-date this PR.
 - All 3152 porch+codev tests pass (148 files); all 17 dashboard tests pass
 - Pre-existing failure in `scrollController.test.ts` (terminal scroll, unrelated to this work)
 
+## iter-2: architect CMAP REQUEST_CHANGES
+Architect-side 3-way CMAP (dual-CMAP pattern) caught two real defects that iter-1's builder CMAP missed individually:
+
+**Fix 1 — Codex finding**: `isPrCreatingPhase` overmatched on bare consultation presence. RESEARCH's `investigate`/`critique` phases have consultation blocks for non-PR purposes (`type: "investigation"` / `"critique"`). Iter-1 would have set `pr_ready_for_human: true` on research builders. Narrowed the marker to `consultation.on === 'review'` — the existing field both BUGFIX and AIR already carry and RESEARCH lacks. Renamed `hasConsultation` → `hasPrConsultation` to reflect the narrower semantics.
+
+**Fix 2 — Gemini finding**: `NeedsAttentionList` builder loop did `!b.blocked || !b.blockedSince` early-out BEFORE checking `prReady`. BUGFIX builders have both null (no `pr` gate), so the missing-PR defense fell through silently. The iter-1 test passed only because it mocked a BUGFIX with `blocked='PR review'`, which is unrealistic. Restructured the loop: prReady check first, then gate-blocked. Added `waitingSince` fallback chain (`blockedSince → startedAt → now`) for gateless protocols.
+
+**iter-2 tests added**:
+- 6 `isPrCreatingPhase` classifier tests including 2 that pin RESEARCH `investigate` / `critique` to false (would fail under iter-1 logic)
+- Realistic BUGFIX missing-PR test (`blocked=null, blockedSince=null`) — would fail under iter-1 dashboard logic
+- AIR-style gated variant test preserves iter-1 coverage explicitly
+
+All 247 targeted tests pass; full build clean; one unrelated flaky timeout in `next.test.ts` (different test each run — parallel-load contention, not the classifier path).
+
--- a/codev/state/bugfix-872_thread.md
+++ b/codev/state/bugfix-872_thread.md
@@ -1,0 +1,49 @@
+# bugfix-872 thread
+
+## Issue
+porch: emit explicit `pr_ready_for_human` state field when CMAP completes.
+
+PR #844 used `blocked === 'PR review'` (i.e. `pr` gate pending) to gate Needs Attention, which silently dropped BUGFIX PRs (no `pr` gate; transitions straight to `phase: verified` after CMAP).
+
+## Plan
+Canonical signal lives in `status.yaml` as `pr_ready_for_human: boolean`. Porch sets it true at the moment CMAP-emitting state exits for the PR-creating phase, in all 5 protocols.
+
+### State-machine touchpoints (the only 3 places that need a write)
+- **handleVerifyApproved (next.ts)** — fires for SPIR/ASPIR/PIR review (build_verify, gate=pr). When gate becomes pending → set `true`.
+- **done (index.ts), auto-request `pr` gate** — fires for AIR pr (once, gate=pr). When `pr` gate gets `requested_at` → set `true`.
+- **advanceProtocolPhase (index.ts) → 'verified'** — fires for BUGFIX pr (once, no gate, terminal). When transitioning from a PR-creating phase to verified → set `true`.
+
+### Reset to false
+- handleBuildVerify REQUEST_CHANGES → rebuttal cycle → explicit `false` (defensive; the field probably wasn't true anyway since rebuttal happens at iter1 and gate hasn't been requested yet).
+
+### PR-creating phase detection
+- Has `gate === 'pr'` (AIR/SPIR/ASPIR/PIR), OR
+- Has `consultation` block in protocol.json (BUGFIX type:once).
+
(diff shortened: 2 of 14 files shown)


waleedkadous:
## CMAP Review (3-way, iter 1)

All three reviewers **APPROVE** with **HIGH** confidence.

| Model | Verdict | Summary |
|---|---|---|
| Gemini | APPROVE | The PR perfectly implements the canonical `pr_ready_for_human` signal across all protocols, closing the Needs Attention gap for BUGFIX PRs, backed by thorough tests. |
| Codex | APPROVE | The fix is focused, matches issue #872, and adds solid regression coverage for BUGFIX plus the cross-protocol state transitions. |
| Claude | APPROVE | Clean, well-tested canonical signal that fixes BUGFIX PR visibility regression with solid backward compatibility. |

**KEY_ISSUES from all three: none.**

Claude flagged one non-blocking observation: the condition `gate === 'pr' || isPrCreatingPhase(...)` is technically redundant when the first half matches, but is kept for clarity ("this is a PR gate OR this is a PR-creating phase"). Not worth changing.

Review files committed in `codev/projects/bugfix-872-porch-emit-explicit-pr-ready-f/`.

waleedkadous:
## Architect integration review

Ran a 3-way architect-side CMAP per the dual-CMAP discipline. The builder's own CMAP was unanimous APPROVE but the architect-side caught two real bugs both other reviewers missed individually. Posting findings here so the PR record reflects them.

**Verdicts:** Gemini REQUEST_CHANGES (HIGH) · Codex REQUEST_CHANGES (HIGH) · Claude APPROVE (HIGH).

The state-machine three-tier design (`status.yaml` write → `derivePrReady` normalize → consumer gates on boolean) is well-architected and the right shape. Two specific defects need iter-2.

### Required fix 1: `isPrCreatingPhase` overmatches RESEARCH (Codex)

`isPrCreatingPhase()` returns `true` for any phase with a `consultation` block. But `codev/protocols/research/protocol.json` has `investigate` and `critique` phases that carry consultation blocks and are **not** PR-creating. Consequences:

- A research builder gets marked `pr_ready_for_human: true` even though no PR is involved
- `approve()` only clears the flag for gate `pr`, so a non-PR consultation phase that advances via a different gate leaves the flag **stuck true**
- The "canonical" signal is now inaccurate outside the five target protocols

**Fix direction**: narrow the classifier. `consultation` alone isn't enough. Options:
- Require `consultation.type === 'pr'` (or `consultation.on === 'review'` matching the PR-creating shape) in addition to having consultation
- Add an explicit `is_pr_phase: true` marker to the protocol.json schema for PR-creating phases and check that
- Whichever you pick, add a regression test using RESEARCH (or a contrived non-PR consultation phase) so this can't regress silently

### Required fix 2: `NeedsAttentionList` builder fallback drops BUGFIX builders (Gemini)

The builder loop (`// Builders blocked on gate approvals`) does `if (!b.blocked || !b.blockedSince) continue` before checking anything else. **BUGFIX builders always have `b.blocked === null`** — so if a BUGFIX PR is missing from the `prs` array (cache delay, pagination, transient API failure), the builder hits the `continue` and disappears entirely. That defeats the exact defensive case this PR is supposed to fix.

The test `still surfaces a prReady builder when its PR is missing from prs` mocks the BUGFIX builder with `blocked: 'PR review'` — that's unrealistic. Real BUGFIX builders have `blocked: null` and `blockedSince: null`. The test passes the unrealistic case but the realistic one breaks silently.

**Fix direction**: in the builder loop, check `b.prReady` *before* the `!b.blocked` skip. Emit a prReady-driven attention row when present; otherwise fall through to the gate-blocked logic. Update the BUGFIX test fixture to use realistic shape (`blocked: null, blockedSince: null`).

### Non-blocking notes

- **Claude**: VSCode builder tree (`packages/vscode/src/views/builders.ts`) could use `prReady` to boost BUGFIX builders into the priority section. Not in this PR's scope; flag for a small follow-up.
- **Claude**: the `derivePrReady` fallback becomes dead code once all in-flight builders have cycled. Worth a v3.2+ cleanup pass.
- **Claude**: minor redundancy in `gate === 'pr' || isPrCreatingPhase(...)` since `isPrCreatingPhase` already checks `gate === 'pr'`. Readable as-is, leave it.

Please push iter-2 with the two fixes above and the porch `pr` gate will re-trigger CMAP. Don't merge yet; same explicit-human-approval pattern as before.

waleedkadous:
## CMAP Review (3-way, iter 2)

Re-ran 3-way CMAP after addressing both architect findings. All three reviewers **APPROVE** with **HIGH** confidence.

| Model | Verdict | Summary |
|---|---|---|
| Gemini | APPROVE | Both issues requested by the Architect have been fully addressed (RESEARCH protocol overmatch and NeedsAttentionList BUGFIX drop) with complete test coverage. |
| Codex | APPROVE | Iter-2 fixes the two real architect-raised defects, keeps the bugfix focused, and adds strong regression coverage for both the porch state machine and dashboard fallback paths. |
| Claude | APPROVE | Clean, well-tested canonical signal; both architect iter-2 bugs addressed and verified against bundled protocol.json files. |

**KEY_ISSUES from all three: none.**

### What changed in iter-2

1. **`isPrCreatingPhase` narrowed**: classifier no longer matches bare consultation presence. Now requires `consultation.on === 'review'` (the marker BUGFIX/AIR pr phases already carry; RESEARCH investigate/critique do not). Renamed `ProtocolPhase.hasConsultation` → `hasPrConsultation`. **6 new classifier tests** including 2 that pin RESEARCH `investigate`/`critique` to `false` — would fail under iter-1 logic.

2. **`NeedsAttentionList` builder loop restructured**: `prReady` check now comes BEFORE the `!b.blocked` early-out so BUGFIX builders (where blocked is always null) survive the cache-miss defense path. Added `waitingSince` fallback chain (`blockedSince → startedAt → now`) for gateless protocols. The iter-1 missing-PR test split into a realistic BUGFIX case (`blocked: null, blockedSince: null`) — which would fail under iter-1 logic — and a preserved AIR-style gated variant.

waleedkadous:
Approved. Both fixes are surgical and well-commented — the inline note in `isPrCreatingPhase`'s JSDoc explaining why bare `consultation` presence isn't enough is exactly the kind of comment that prevents this regression from recurring. The renaming `hasConsultation` → `hasPrConsultation` is the right call too — the field name now matches what it actually means.

Please merge with `--admin` bypass.

PR #879, opened 2026-05-27
Fix #878: Point gemini consult lane at gemini-3.1-pro-preview

### LINKED ISSUE #878: consult: gemini lane points at retired gemini-3-pro-preview model — every CMAP gemini-side fails fast with opaque error
## Symptom

Every \`consult -m gemini ...\` invocation fails in ~1.8s with an opaque error:

\`\`\`
Error when talking to Gemini API Full report available at:
/var/folders/.../gemini-client-error-Turn.run-sendMessageStream-*.json
{
  "session_id": "...",
  "error": { "type": "Error", "message": "[object Object]", "code": 1 }
}
[gemini completed in 1.8s]
Process exited with code 1
\`\`\`

The \`[object Object]\` is misleading — the actual error inside the dumped JSON report says:

\`\`\`
ModelNotFoundError: This model models/gemini-3-pro-preview is no longer available.
Please update your code to use a newer model for the latest features and improvements.
\`\`\`

**Google has retired \`gemini-3-pro-preview\`.** Codev's consult CLI is hardcoded to call it.

## Reproduce

Any project, any user with valid Gemini auth:

\`\`\`bash
consult -m gemini --prompt "Reply with the word OK and nothing else." --output /tmp/gemini-test.md
\`\`\`

Fast-fail (~1.8s), exit 1, empty output file. The 1.8s timing is the giveaway: a real Gemini call takes ~120–150s per the consult skill — sub-2s is the model-rejected-before-call signature.

## Root cause

\`packages/codev/src/commands/consult/index.ts:38\`:

\`\`\`ts
gemini: { cli: 'gemini', args: ['--model', 'gemini-3-pro-preview'], envVar: 'GEMINI_SYSTEM_MD' },
\`\`\`

The model identifier was hardcoded when \`gemini-3-pro-preview\` was the current preview model. Google has since moved \`gemini-3-pro\` to GA and removed the preview alias.

## Real-world impact

Discovered during PIR builder #819's CMAP-2 pass at the PR gate (2026-05-27). Three consecutive gemini retries failed identically. Builder reported partial CMAP — only Codex (COMMENT, addressed) and Claude (APPROVE) returned verdicts. Porch was left blocking on the missing Gemini output.

Pattern is reproducible for every protocol that runs CMAP — SPIR (3-way at multiple gates), PIR (1-pass at PR), MAINTAIN, etc. The gemini lane silently fails on every run; users either skip-and-advance manually or get wedged porch state.

The cost: ~6 seconds wasted per CMAP run (3 retries × 2s) plus an "incomplete CMAP" condition that breaks the protocol's "wait for all 3" assumption.

## Fix

Bump the model identifier in three locations:

### 1. The consult model config

\`packages/codev/src/commands/consult/index.ts:38\`:

\`\`\`diff
-  gemini: { cli: 'gemini', args: ['--model', 'gemini-3-pro-preview'], envVar: 'GEMINI_SYSTEM_MD' },
+  gemini: { cli: 'gemini', args: ['--model', 'gemini-3-pro'], envVar: 'GEMINI_SYSTEM_MD' },
\`\`\`

Verification that \`gemini-3-pro\` (no \`-preview\` suffix) is correct: \`packages/codev/src/commands/consult/usage-extractor.ts:19\` already has pricing for it, suggesting it's the established GA identifier.

### 2. Documentation references

\`CLAUDE.md:284\` and \`AGENTS.md:284\` both say:

> Gemini 3 Pro (gemini-3-pro-preview) for deep analysis

Update both to \`gemini-3-pro\` to match the new config.

### 3. Tests that reference the retired identifier

\`packages/codev/src/commands/consult/__tests__/metrics.test.ts\` has \`'gemini-3-pro-preview'\` keys in fixture data (line 617 and possibly others). Either update to \`gemini-3-pro\` to match the new shipped identifier, or update the test assertion to be model-name-agnostic if the test doesn't actually care about the exact string.

## Acceptance criteria

- [ ] \`packages/codev/src/commands/consult/index.ts:38\` uses \`gemini-3-pro\` (or whatever the current GA identifier is — verify via \`gemini --help\` or Google's model list before merging)
- [ ] \`CLAUDE.md\` and \`AGENTS.md\` updated to reflect the new identifier in their "Multi-Agent Consultation" sections
- [ ] Any test fixture that hardcodes \`gemini-3-pro-preview\` is updated to the new identifier
- [ ] Manual smoke: \`consult -m gemini --prompt "OK?" --output /tmp/x.md\` returns a real response within 60–150s, exit 0, non-empty output file
- [ ] CMAP runs across SPIR / PIR / etc. now complete the gemini lane in normal time (~120–150s) instead of fast-failing in ~2s

## Out of scope

- Adding fallback logic ("if model X fails, try model Y") — out of scope for this fix; the right answer is to point at a model that exists
- Reading the model identifier from external config — could be a follow-up if Google's model lifecycle becomes a recurring issue, but for now the in-source identifier is fine
- Improving the consult CLI's error reporting — separate concern; the \`[object Object]\` opaque message is a known issue in the upstream gemini CLI tool, not Codev's

## Related

- \`packages/codev/src/commands/consult/index.ts\` — model config
- \`packages/codev/src/commands/consult/usage-extractor.ts\` — pricing table (already lists \`gemini-3-pro\`)
- PIR #819 — discovered during this builder's CMAP-2 attempt; advanced via manual porch override pending this fix

waleedkadous:
## Summary

Google retired `gemini-3-pro-preview` on 2026-03-09, so every CMAP gemini-side fast-fails in ~1.8s with an opaque `[object Object]` error (the gemini-cli's misformatted `ModelNotFoundError` dump). This PR bumps the consult lane to `gemini-3.1-pro-preview` — the current actively-supported Gemini 3.x Pro identifier.

Fixes #878

## Root Cause

`packages/codev/src/commands/consult/index.ts:38` hardcoded `gemini-3-pro-preview` when that was the current preview alias (Feb 2026). Google has since shut it down, and `gemini-3-pro` is not the active GA identifier either — the current target is `gemini-3.1-pro-preview` (Gemini 3.1 Pro, released Feb 19, 2026).

## Fix

- `consult/index.ts` — bump `--model` arg to `gemini-3.1-pro-preview`.
- `usage-extractor.ts` — add `gemini-3.1-pro` pricing tier (`$2.00/$12.00 per 1M`, cached at 25% of input). Placed before `gemini-3-pro` in the prefix-match table so the 3.1 variant resolves first.
- `CLAUDE.md` / `AGENTS.md` — update the Multi-Agent Consultation reference.
- `metrics.test.ts` — update the Pro pricing fixture to the new model and cost; add a focused regression test asserting that `MODEL_CONFIGS.gemini.args` points at `gemini-3.1-pro-preview` (the exact thing that regressed).
- Export `MODEL_CONFIGS` as `_MODEL_CONFIGS` for the regression test, matching the existing `_`-prefixed test-export pattern in `consult/index.ts`.

## Verification

- Smoke-tested locally: `gemini --model gemini-3.1-pro-preview --prompt "OK"` returns `OK` in ~4s. `consult -m gemini --prompt "OK"` returns a real response.
- `npx tsc` passes clean.
- Full test suite: 3162 passed, 13 skipped, 0 failed (vitest run).

## Test Plan

- [x] Regression test added (`metrics.test.ts` — guards against the model identifier silently regressing back to a retired model)
- [x] Build passes (`npx tsc` clean)
- [x] All tests pass (`vitest run` — 3162 passed)
- [x] Manual smoke: `consult -m gemini --prompt "OK?"` returns a real response, exit 0

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -281,7 +281,7 @@ Use sequential numbering with descriptive names (no leading zeros):
 ## Multi-Agent Consultation
 
 **DEFAULT BEHAVIOR**: Consultation is ENABLED by default with:
-- **Gemini 3 Pro** (gemini-3-pro-preview) for deep analysis
+- **Gemini 3.1 Pro** (gemini-3.1-pro-preview) for deep analysis
 - **GPT-5.4 Codex** (gpt-5.4-codex) for coding and architecture perspective
 
 To disable: User must explicitly say "without multi-agent consultation"
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -281,7 +281,7 @@ Use sequential numbering with descriptive names (no leading zeros):
 ## Multi-Agent Consultation
 
 **DEFAULT BEHAVIOR**: Consultation is ENABLED by default with:
-- **Gemini 3 Pro** (gemini-3-pro-preview) for deep analysis
+- **Gemini 3.1 Pro** (gemini-3.1-pro-preview) for deep analysis
 - **GPT-5.4 Codex** (gpt-5.4-codex) for coding and architecture perspective
 
 To disable: User must explicitly say "without multi-agent consultation"
--- a/codev/projects/bugfix-878-consult-gemini-lane-points-at-/status.yaml
+++ b/codev/projects/bugfix-878-consult-gemini-lane-points-at-/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-878
 title: consult-gemini-lane-points-at-
 protocol: bugfix
-phase: pr
+phase: verified
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,5 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-27T05:34:36.124Z'
-updated_at: '2026-05-27T05:46:31.999Z'
+updated_at: '2026-05-27T05:52:30.741Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-878-consult-gemini-lane-points-at-/status.yaml
+++ b/codev/projects/bugfix-878-consult-gemini-lane-points-at-/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-878
 title: consult-gemini-lane-points-at-
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates: {}
@@ -9,4 +9,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-27T05:34:36.124Z'
-updated_at: '2026-05-27T05:34:36.125Z'
+updated_at: '2026-05-27T05:40:03.706Z'
--- a/codev/projects/bugfix-878-consult-gemini-lane-points-at-/status.yaml
+++ b/codev/projects/bugfix-878-consult-gemini-lane-points-at-/status.yaml
@@ -1,0 +1,12 @@
+id: bugfix-878
+title: consult-gemini-lane-points-at-
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates: {}
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-05-27T05:34:36.124Z'
+updated_at: '2026-05-27T05:34:36.125Z'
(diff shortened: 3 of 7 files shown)


waleedkadous:
Approved — the new identifier is forward-safe regardless of the rolling-deprecation wrinkle. CMAP unanimous including gemini end-to-end is the right confirmation. Please merge with --admin bypass.

Issue #783, opened 2026-05-20
afx cleanup can't reach orphaned Tower terminals after porch done self-completion (registry-keyed cleanup, no fallback)


amrmelsayed:
## Problem

When a builder runs `porch done <id>` to mark its protocol complete (BUGFIX, AIR, PIR — any protocol that ends with a `porch done` rather than awaiting an external signal), the builder is **deregistered from afx's project/issue registry** as part of completion. Subsequent `afx cleanup` attempts then fail because cleanup is registry-keyed — both `afx cleanup -p <id>` and `afx cleanup -i <issue>` return:

```
[error] Builder not found for project: <id>
[error] Bugfix builder not found for issue #<issue>
```

The **Tower-level PTY session** for the builder is not killed by `porch done` (Tower's terminal registry is independent of the project/issue registry). It remains visible in `afx status` as an `(active)` terminal indefinitely. There is no registry-independent terminal-kill command in `afx`.

## Observed

Sequence on a real BUGFIX builder run that completed cleanly:

1. Builder ran `porch done <id>` (BUGFIX protocol-complete on merged PR).
2. Builder sent the "ready for cleanup" notification.
3. `afx cleanup -p <id>` → `Builder not found for project: <id>` (builder already deregistered).
4. Manually `git worktree remove --force` + `git branch -D builder/<...>` succeeded (those don't go through afx's registry).
5. `afx cleanup -i <issue> -f` → `Bugfix builder not found for issue #<issue>` (same; cleanup still can't find a registry entry).
6. `afx status` continued to show the builder's terminal as `(active)` with no work, no worktree, no branch, no porch state — a dead PTY in Tower with a stale registry entry.

The only afx mechanism that *can* clear an orphaned Tower terminal is `afx tower stop && afx tower start` — which kills **every other live builder terminal** in the session. Unacceptable collateral in any multi-builder workflow.

## Impact

- **Every self-completing builder accumulates a permanent orphan terminal.** `afx status` becomes increasingly noisy with dead entries from past completed work, with no way to distinguish dead from live except by external knowledge.
- **No clean recovery exists.** Killing the underlying PTY process directly bypasses Tower and risks Tower-state inconsistency. Restarting Tower wipes all live work. The user is left with permanent cosmetic noise that compounds across sessions.
[...]

PR #892, opened 2026-05-27
Fix #883: close builder terminal tabs on cleanup by diffing overview, not state

### LINKED ISSUE #883: vscode: builder cleanup no longer closes the terminal tab — regression of 3.0.2 behavior
## Symptom

When a builder is cleaned up (via `afx cleanup`, the VSCode "Cleanup Builder" command, or any other removal path), the **builder's terminal tab in VSCode stays open** as a dead "Process exited" entry. User has to close it manually.

## Past behavior (expected)

This worked correctly in 3.0.2. From `packages/vscode/CHANGELOG.md` (3.0.2 bug fixes):

> "Builder terminal tabs close automatically on cleanup instead of lingering as dead 'Process exited' tabs."

So the closing-on-cleanup is an explicit shipped feature, not an emergent property. It has regressed.

## Reproduce

1. Spawn a builder (any protocol): `afx spawn <issue> --protocol bugfix`
2. Open the builder's terminal in VSCode (single-click the row in the Builders view, or use `Codev: Open Builder Terminal`)
3. Wait for the builder to do its work and the PR to merge
4. Clean up the builder via any of:
   - `afx cleanup -p <id>` from a CLI
   - Right-click builder row → `Codev: Cleanup Builder` (per #795 — if landed; or via palette)
5. Observe: VSCode's terminal tab stays open with `Process exited` instead of disappearing

## Code path the regression broke

The auto-close mechanism is two-piece:

### Piece 1 — Present→absent diff in `extension.ts:199-234`

```ts
let prevBuilderIds: Set<string> | null = null;
let pruneInFlight = false;
const pruneClosedBuilderTerminals = async () => {
  if (pruneInFlight) { return; }
  if (connectionManager?.getState() !== 'connected') { return; }
  // ...
  const state = await client.getWorkspaceState(workspacePath);
  if (!state?.builders) { return; }
  const currIds = new Set(state.builders.map(b => b.id));
  if (prevBuilderIds !== null) {
    for (const prev of prevBuilderIds) {
      if (!currIds.has(prev)) {
        terminalManager?.closeBuilderTerminal(prev);
      }
    }
  }
  prevBuilderIds = currIds;
  // ...
};
```

Pre-conditions for this to fire on cleanup:
- `pruneClosedBuilderTerminals` is wired to a trigger (overview tick, SSE event, etc.)
- `connectionManager.getState() === 'connected'`
- A prior tick populated `prevBuilderIds` with the soon-to-be-cleaned-up builder
- A subsequent tick fetches state where that builder is **absent** from `state.builders`

### Piece 2 — Disposal in `terminal-manager.ts:262`

```ts
closeBuilderTerminal(builderId: string): void {
  for (const key of [`builder-${builderId}`, `dev-${builderId}`]) {
    const existing = this.terminals.get(key);
    if (!existing) { continue; }
    existing.pty.close();
    existing.terminal.dispose();
    this.terminals.delete(key);
  }
}
```

## Plausible regression causes (for plan-phase investigation)

The PIR builder should investigate which of these is happening — the symptom is the same, but the fix differs per cause:

1. **`pruneClosedBuilderTerminals` is no longer wired to a trigger** — maybe a refactor of the overview cache subscription dropped the call. Check `overviewCache.onDidChange(...)` registrations in extension.ts.
2. **Tower's `state.builders` is no longer dropping the cleaned-up builder** — could be a Tower-side regression where `afx cleanup` succeeds but the DB row stays. Symptom-side, the diff in step 1 never sees absence. **Particularly worth investigating** given the recent Tower DB migration issues (the orphan `builders_new` table that wedged `afx cleanup` during 2026-05-26 sessions). If `afx cleanup` is silently failing to remove the row, the diff has nothing to detect.
3. **`prevBuilderIds` initialization race** — if cleanup happens before the first successful state fetch, `prevBuilderIds` is `null` and the diff is skipped entirely. The freshly-cleaned builder gets recorded in `currIds` (wait — actually if it's already gone, it wouldn't be in `currIds` either; needs careful trace). May need a sentinel or different initialization to handle this edge.
4. **The `pruneInFlight` guard wedged** — if a state fetch hangs and never sets `pruneInFlight = false`, subsequent ticks no-op forever. Add a timeout or finally-only release.
5. **`closeBuilderTerminal` finds nothing in the map** — if the key encoding has changed (e.g. builder ID format drift between when `openBuilder` registers and when `closeBuilderTerminal` looks up), the iteration silently skips both keys.
6. **`existing.terminal.dispose()` no longer kills the VSCode tab** — VSCode API regression, unlikely but worth verifying with a manual dispose call.

The plan phase should identify which one (or which combination) is the real cause via reproduction with logging, BEFORE writing the fix.

## Why PIR

Per the protocol-selection guide:
- **UI / UX change requiring visual confirmation in running VSCode** — the symptom is observable only in a live VSCode session. Unit tests on the diff logic won't catch a real regression in terminal-tab disposal.
- **Root cause is ambiguous** — 6 plausible causes listed above. Plan-approval is the right place to lock in the diagnosis before coding.
- **High-blast-radius area** — terminal lifecycle touches a frequently-used surface (every builder spawn opens a terminal; every cleanup needs to close one). Fix needs to be confident, not a guess.

## Acceptance criteria

- [ ] Plan phase identifies the actual root cause (or causes) with reproduction logs — not just a hypothesis
- [ ] Fix applied at the identified failure point(s)
- [ ] Reproduce the symptom with the fix in place — terminal tab closes within ~5 seconds of `afx cleanup` completing
- [ ] Visual verification in a running VSCode session at `dev-approval` gate (open VSCode, spawn a builder, open terminal, run `afx cleanup -p <id>`, confirm tab disappears)
- [ ] Both `builder-<id>` and `dev-<id>` keys disposed (the existing loop handles both — verify the dev-terminal path too)
- [ ] No regression: spawning a fresh builder after a cleanup still opens its terminal correctly (i.e. the cleanup didn't break the openBuilder map state)
- [ ] Regression test if the failure is unit-testable (e.g. mock `getWorkspaceState` to return a builder absent-after-present and assert `closeBuilderTerminal` is called)

## Out of scope

- **Auto-cleaning ghost terminal_sessions in `~/.agent-farm/global.db`** — separate concern; that's Tower-side bookkeeping, not VSCode-side terminal tabs
- **Closing terminal tabs for cleaned-up architects** — different mechanism, different code path
- **Closing dev terminals on PR merge or other lifecycle events** — only `afx cleanup` (= builder removed from Tower state) is the trigger this issue addresses

## References

- `packages/vscode/CHANGELOG.md` — 3.0.2 entry that documents the original feature
- `packages/vscode/src/extension.ts:199-234` — present→absent diff (`pruneClosedBuilderTerminals`)
- `packages/vscode/src/terminal-manager.ts:262` — the actual `closeBuilderTerminal` disposal logic
- `packages/vscode/src/commands/cleanup.ts:50` — VSCode side's `Codev: Cleanup Builder` command that shells to `afx cleanup`

## Discovered while

Observed during 2026-05-27 batch operations — cleaning up merged bugfix builders left their terminal tabs lingering in VSCode. User confirmed this was working in earlier versions.

amrmelsayed:
# PIR Review: Close builder terminal tabs on cleanup (read overview, not state)

Fixes #883

## Summary

The 3.0.6 fix for "builder terminal tabs close automatically on cleanup"
regressed because the present→absent diff was reading from the wrong
source. This PR points the diff at `overviewCache.getData().builders`
(disk scan via `discoverBuilders`) instead of
`client.getWorkspaceState().builders` (runtime registry rebuilt from
SQLite `terminal_sessions`), so the diff sees the absence the moment
`afx cleanup` removes the worktree directory — even while the
companion Tower-side bug (#783) leaves surviving shellper processes
pinning the SQLite-backed source open.

## Files Changed

- `codev/plans/883-vscode-builder-cleanup-no-long.md` (+216 / -0) — plan artifact
- `codev/state/pir-883_thread.md` (+79 / -0) — protocol thread log
- `packages/vscode/src/prune-builder-terminals.ts` (+62 / -0) — new pure helper module
- `packages/vscode/src/extension.ts` (+21 / -34) — diff wiring switched to overview cache
- `packages/vscode/src/__tests__/prune-builder-terminals.test.ts` (+116 / -0) — 11 vitest cases
- `codev/projects/883-vscode-builder-cleanup-no-long/status.yaml` (+22 / -0) — porch state

## Commits

- `e68197ac` [PIR #883] Plan draft
- `5c21360b` [PIR #883] Diff against overview.builders.roleId, not state.builders.id
- `724c4ea9` [PIR #883] Thread: implement phase complete

## Test Results

- `pnpm --filter codev-vscode check-types`: ✓ pass
- `pnpm --filter codev-vscode lint`: ✓ pass
- `pnpm --filter codev-vscode test:unit` (vitest): ✓ pass (49 tests, 11 new)
- `pnpm --filter codev-vscode test` (mocha integration): ✓ pass (83 tests)
- `porch done 883` checks: ✓ build (5.3 s), ✓ tests (20.5 s)
- Manual verification at the `dev-approval` gate: human approved

## Architecture Updates

No `arch.md` changes — the diff swaps one data source for another at the
same architectural boundary (VSCode extension reading from Tower).
There's no new module, pattern, or boundary to document. The new
`prune-builder-terminals.ts` helper is a vscode-free sidecar so a unit
test can import it without mocks; that's a testability detail, not an
architectural one.

## Lessons Learned Updates

Added one entry to `codev/resources/lessons-learned.md` under **Critical**:

> When a VSCode-side observer needs to react to a Tower-side cleanup,
> source the signal from the disk-truth endpoint (`/api/overview`,
> backed by `discoverBuilders`' `readdirSync(.builders/)` scan), not the
> runtime-registry endpoint (`/api/state`, backed by SQLite
> `terminal_sessions` reconciled against surviving shellpers). Detached
> shellper processes are *designed* to outlive Tower restarts and can
> therefore pin runtime-registry rows open through Tower's
> reconnect-on-the-fly path indefinitely. The filesystem state collapses
> to "did `afx cleanup` remove this worktree?" — which is the actual
> cleanup signal — so a disk-sourced diff is resilient to the orphan-
> shellper class of bugs (cf. #783) by construction.

This is a generalizable rule: any cleanup-detection diff in the
extension should default to the disk-scan source, and pick the
registry source only when worktree-existence isn't the right signal.

- **The helper went into its own module** (`prune-builder-terminals.ts`)
  rather than living "adjacent to the wiring" in `extension.ts` as the
  plan wording suggested. Reason: vitest can't import a file that imports
  the real `vscode` API, so the test needs the helper to be in a module
  that's free of vscode imports. Functional shape is identical to the
  plan's intent.

- **Soft-mode limitation** (`roleId: null`). Soft-mode builders
  (`task-*`, `worktree-*` worktrees) won't auto-close their tabs via
  this path because the diff is keyed on `OverviewBuilder.roleId` which
  is `null` for them. The issue's repro path is bugfix-mode (strict),
  and the older state-based code wasn't reliably helping soft-mode
  either in the orphan-shellper scenario. Documented inline in the
  helper's doc comment.

- **The companion Tower bug (#783) is unfixed by design.** This PR
  explicitly does not touch the orphan-shellper / ghost
  `terminal_sessions` accumulation — that's #783's scope, and the issue
  marks it **Out of scope** for #883. After #783 lands, the VSCode
  resilience here keeps working (it's still reading the right source)
  but stops being load-bearing.

- **Sync vs the previous async function.** The old
  `pruneClosedBuilderTerminals` was `async` and guarded by `pruneInFlight`
  because it did an HTTP fetch on every tick; the new version is
  synchronous (`overviewCache.getData()` is an in-memory read) and the
  guard is gone. The call site in `overviewCache.onDidChange(...)`
  didn't `await` the old version either, so this is purely simplification.

## How to Test Locally

For reviewers pulling the branch:

- **View diff**: VSCode sidebar → right-click builder pir-883 → **View Diff**
- **Run dev server**: VSCode sidebar → **Run Dev Server**, or `afx dev pir-883`
- **Build + install**: `pnpm build && pnpm -w run local-install` (restarts Tower, picks up the patched extension)
- **What to verify**:
  - Spawn a fresh bugfix builder (`afx spawn <issue> --protocol bugfix`),
    open its terminal in VSCode, then `afx cleanup -p <id>` from a
    separate shell. Tab disappears within ~5 s of the cleanup printing
    `Builder ... cleaned up!`.
  - Repeat via VSCode's right-click → **Cleanup Builder** for the
    second variant.
  - Optionally start a dev terminal (right-click → **Run Dev Server**)
    before cleanup, then confirm its `(dev)` tab also disappears.
  - Spawn a fresh builder after cleanup; its terminal opens normally
    (no map-state pollution from the closure).
  - Sanity probe: `sqlite3 ~/.agent-farm/global.db "SELECT role_id FROM
    terminal_sessions WHERE type='builder'"` will likely still list
    cleaned-up builders (#783 left unfixed by design). The VSCode tab
    still closes — that's the point.

## Related

- **#783** (open, `area/tower`, `bug`) — Tower-side root cause: `afx
  cleanup` can't reach orphaned Tower terminals after `porch done`
  self-completion. The VSCode change here is resilient to that bug;
  fixing #783 makes `/api/state` clean again but doesn't invalidate
  this change.


amrmelsayed:
## Architect Review

**Low-risk fix that addresses the actual root cause, not just the symptom.** 654/34 across 8 files, but most is artifacts: 3 bookkeeping (plan + thread + status.yaml), 1 new test file (+116 LOC, 11 cases), 1 new pure helper module (+62 LOC). Real change to existing code is `-13 net LOC in extension.ts` — the fix simplified the diff logic, not bloated it.

CMAP unanimous (gemini=APPROVE, codex=APPROVE, claude=APPROVE).

### Verified

- **Right root cause identified.** The diff was reading from `client.getWorkspaceState().builders` (SQLite-backed runtime registry) which doesn't reliably update on cleanup. PR switches the source to `overviewCache.getData().builders` (disk scan via `discoverBuilders` reading `.builders/<id>/` directories). Disk scan is always up-to-date — the moment `afx cleanup` removes the worktree directory, the diff sees the absence.
- **Robust to the sibling Tower-side bug.** Even if Tower's `terminal_sessions` table has stale entries (a known issue tracked separately as #783), the disk-scan source ignores them. The fix is independent of whether #783 ever gets resolved
- **Net code reduction in `extension.ts`** (+21 / -34 = net -13) — the rewrite simplified the diff logic, didn't add complexity
- **Pure helper extracted to `prune-builder-terminals.ts`** so the test file can import without mocks. Right testability choice — 11 unit cases cover the diff logic in isolation
- **Plan-approval + dev-approval gates already passed** (human verified the running worktree)

### Notable design choice — the lessons-learned entry

The builder added a Critical-tier entry to `codev/resources/lessons-learned.md`:

> "When a VSCode-side observer needs to react to a Tower-side cleanup, source the signal from the disk-truth endpoint (`/api/overview`, backed by `discoverBuilders`' `readdirSync(.builders/)` scan), not the SQLite-backed runtime registry..."

This generalizes correctly. The same lesson applies anywhere extension-side code observes Tower state to react to cleanup events — there's likely more than one place this matters. Good capture for future work.

### Cross-cutting note

This fix exposes how the **6 plausible causes I listed in the issue body** were not equally weighted. The actual cause was a variant of cause #2 (Tower's state.builders not dropping cleaned-up rows), but the builder's fix is structurally better than what cause #2 suggested — instead of trying to fix the SQLite-backed source (which would couple to #783's outcome), they switched to the disk-scan source that's already correct. That's a stronger answer than the issue body anticipated; worth keeping in mind for future "multiple plausible causes" issue framings — sometimes the right fix is to pick a different data source entirely.

### Verdict

Approved. Per PIR protocol, the `pr` gate is the human's call — awaiting your explicit `porch approve 883 pr --a-human-explicitly-approved-this` before the builder merges.

---
Architect review

PR #893, opened 2026-05-27
vscode: capitalize area group header labels in backlog and builders trees

amrmelsayed:
# PIR Review: Capitalize area group header labels

Fixes #885

## Summary

The Backlog and Builders trees grouped by `area/*` rendered headers verbatim (`vscode (12)`, `tower (4)`), visually inconsistent next to the PascalCase `Uncategorized (8)` fallback. This PR adds a `formatAreaForDisplay` helper in `@cluesmith/codev-core/area-grouping` (title-case with `-`/`_` → space) and applies it in `AreaGroupTreeItem`'s label, so both trees now render `Vscode (12)`, `Tower (4)`, `Cross Cutting (N)`, `Uncategorized (8)`. The raw `areaName` field, the stable `id`, and the `contextValue` continue to use the wire value, so expansion-state persistence and `===` matchers in the view providers are untouched.

## Files Changed

- `packages/core/src/area-grouping.ts` (+27 / −0) — new `formatAreaForDisplay` next to `groupByArea`
- `packages/vscode/src/views/area-group-tree-item.ts` (+8 / −1) — wrap only the displayed label
- `packages/vscode/src/test/area-grouping.test.ts` (+35 / −1) — 7 new cases
- `codev/plans/885-vscode-capitalize-area-group-h.md` (+96 / −0) — approved PIR plan

## Commits

- `830aafe4` [PIR #885] Plan draft
- `a031721e` [PIR #885] Plan revised: title-case + separator-to-space
- `534d94e7` [PIR #885] feat: formatAreaForDisplay helper + apply in AreaGroupTreeItem
- `325ead6d` [PIR #885] test: formatAreaForDisplay cases

## Test Results

- `pnpm build`: ✓ pass
- `pnpm test` (vscode-test / mocha + esbuild type-check + lint): ✓ pass (97 tests, 7 new)
- `pnpm test:unit` (vitest, separate harness): ✓ pass (49 tests)
- Manual verification at the `dev-approval` gate: human reviewed the running worktree via `afx dev pir-885`; group headers in both Backlog and Builders trees render with the new capitalization; expansion state survives reload.

## Architecture Updates

No arch.md changes — this PR adds one helper next to an existing function in the same module, and the call site is a single line. No new module boundary, no new pattern, no consumer-facing API shift. The display-only nature of the helper is documented in its JSDoc (the "wire stays raw" contract).

## Lessons Learned Updates

[...]

PR #896, opened 2026-05-28
Fix #894: mock fetchIssue in next.test.ts to eliminate gh-CLI flake

### LINKED ISSUE #894: porch: flaky 5s timeout in next.test.ts "emits BUILD tasks for fresh specify phase"
## Symptom

Vitest run fails with a single timeout in `packages/codev/src/commands/porch/__tests__/next.test.ts`:

```
❯ src/commands/porch/__tests__/next.test.ts (30 tests | 1 failed) 5474ms
  ⎯⎯⎯⎯⎯⎯⎯ Failed Tests 1 ⎯⎯⎯⎯⎯⎯⎯
  FAIL src/commands/porch/__tests__/next.test.ts > porch next > emits BUILD tasks for fresh specify phase
  Error: Test timed out in 5000ms.

 Test Files 1 failed | 150 passed (151)
      Tests 1 failed | 3179 passed | 21 skipped (3201)
##[error]Process completed with exit code 1.
```

Single test fails by exceeding vitest's 5000ms default timeout. The other 3179 tests pass cleanly. Total runtime for the failing file was 5474ms — i.e. essentially right at the timeout boundary.

## Why this is a flake, not a regression

Concrete evidence: the failure was on the PR-check run for PR #893 (pir-885 — `capitalize area group header labels`). The exact same test passed on the **post-merge run on `main`** ~30 seconds later, with no code change between the two runs.

- **Failing run**: [26545199849](https://github.com/cluesmith/codev/actions/runs/26545199849) (Tests on `builder/pir-885`, 2026-05-27T23:38:41Z)
- **Passing run**: [26545216792](https://github.com/cluesmith/codev/actions/runs/26545216792) (Tests on `main` post-merge, 2026-05-27T23:39:09Z)

Same code (the merge was a fast-forward of the same commits), different result → timing-dependent flake.

Additionally: pir-885's change was localized to `packages/core/src/area-grouping.ts` (capitalize headers) + a test file. No porch code touched. The failing test is in `packages/codev/src/commands/porch/__tests__/next.test.ts` — different package, different concern.

## Root cause hypothesis

The test "emits BUILD tasks for fresh specify phase" likely does something time-bounded that's borderline at the 5s default timeout on CI infrastructure (GitHub Actions Linux runners under variable load). Three plausible candidates:

1. **Spawns a real subprocess** (e.g. `git`, `afx`, or `node`) and waits for it to complete. Subprocess startup on CI can vary from ~100ms to several seconds depending on host load. If it spawns several in sequence, the cumulative time crosses 5s.
2. **Awaits file-system operations** (creating worktree, writing porch state, etc.) that depend on the runner's disk I/O speed.
3. **Real timer or interval inside the test** that the test waits for, and the test's `await`-on-event pattern is racy with the timeout.

The fix surface depends on which of the three. The investigation phase of BUGFIX should reproduce locally (or reason about the code) before picking a fix.

## Fix options

Pick after diagnosing the cause:

| If cause is | Fix |
|---|---|
| Genuinely slow subprocess/IO that's near 5s | Bump `test.timeout` for this specific test (e.g. `it('...', { timeout: 15_000 }, ...)`) — pragmatic if the operation is legitimately heavy |
| Subprocess that should be mocked | Mock the subprocess via vi.mock so the test doesn't depend on CI host performance |
| Racy timer / event-await | Replace timer-based wait with deterministic event subscription; remove polling |
| Real test code path is slow even locally | Refactor the test (or the code under test) to be faster — likely the right answer if it surfaces an actual codepath slowness |

Default: bumping the timeout is the cheapest fix that resolves the immediate symptom. If the root cause is real performance bloat, the bump is a band-aid and a follow-up performance issue is warranted. But for an intermittent flake at the 5s boundary, a bump to 15s usually buys decisive headroom without papering over a real bug.

## Acceptance criteria

- [ ] Identify which of the three causes is actually responsible (with evidence — local repro or code reading)
- [ ] Apply the appropriate fix per the table above
- [ ] Test passes consistently across 5 consecutive CI runs (could verify by re-running the same workflow 5 times)
- [ ] If the fix is just a timeout bump, note in the PR body whether a follow-up "porch next test is genuinely slow — investigate" issue should be filed
- [ ] No regression: the rest of `next.test.ts` (29 other tests) still pass

## Reproduce locally

```bash
pnpm --filter @cluesmith/codev test -- src/commands/porch/__tests__/next.test.ts
# Run several times to estimate flake rate
for i in {1..10}; do pnpm --filter @cluesmith/codev test -- src/commands/porch/__tests__/next.test.ts 2>&1 | tail -3; done
```

If it never fails locally but consistently flakes on CI, that's strong signal that CI infrastructure is the trigger (option 1 or 2 from the hypothesis).

## Out of scope

- **Fixing the other 29 tests in the file** — they pass; this issue is the one timeout
- **Auditing other porch tests for similar timeout-flake risk** — separate sweep if this turns out to be a pattern
- **Investigating GitHub Actions runner performance** — outside our control

## Discovered while

PR #893 (pir-885 — `capitalize area group headers`) hit this flake during its PR-check Tests run on 2026-05-27. Merged anyway via `--admin` override since the failure was unrelated to the PR's change and the post-merge run on main confirmed the test passes.

amrmelsayed:
## Summary

The `emits BUILD tasks for fresh specify phase` test (and ~9 other tests in `next.test.ts` that hit the build path) intermittently timed out at vitest's default 5s on CI. Root cause was a real `gh issue view` subprocess on every `next()` call. This PR mocks `fetchIssue` to make the suite deterministic and ~40x faster.

Fixes #894

## Root Cause

`next()` → `buildPhasePrompt()` → `getProjectSummary()` → `fetchIssue()` shells out to `gh issue view <id>` via `executeForgeCommand`. Every test exercising the build path therefore pays one gh+network round-trip. Locally each call is ~900ms; on CI under variable load it can spike past 5s — vitest's default per-test timeout — and any of the 10 affected tests is equally at risk on a slow run. The failing test in PR #893's CI happened to be the first to cross the line.

This matches hypothesis #2 from the issue's fix-options table ("subprocess that should be mocked").

## Fix

Added a single `vi.mock('../../../lib/github.js', …)` block (8 lines including comment) to `next.test.ts`, returning `null` from `fetchIssue`. This is the same pattern already used in `src/__tests__/project-summary.test.ts`. Tests fall through to the existing spec-file / status-title fallbacks — the same path they implicitly took before, since fixtures don't create real GitHub issues.

## Test Plan

- [x] Build passes (`pnpm build`)
- [x] All 30 tests in `next.test.ts` still pass
- [x] All 334 tests across `src/commands/porch/__tests__/` still pass
- [x] File duration drops from ~11s → ~258ms; per-test build-path times drop from 836-2000ms → 2-8ms (even a 100x CI slowdown stays well under 5s)
- [ ] No follow-up "test genuinely slow" issue needed — the fix is a proper mock, not a timeout bump, so there's no real performance bloat being papered over

## Verification across CI runs

After this lands, the test should pass consistently. If the flake recurs, the cause is something other than `fetchIssue` and would need fresh investigation.

### DIFF
--- a/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
+++ b/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-894
 title: porch-flaky-5s-timeout-in-next
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-28T00:50:34.526Z'
-updated_at: '2026-05-28T00:55:59.958Z'
+updated_at: '2026-05-28T00:58:26.606Z'
--- a/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
+++ b/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-894
+title: porch-flaky-5s-timeout-in-next
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-05-28T00:50:34.526Z'
+updated_at: '2026-05-28T00:50:34.527Z'
--- a/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
+++ b/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
@@ -6,11 +6,12 @@ plan_phases: []
 current_plan_phase: null
 gates:
   pr:
-    status: pending
+    status: approved
     requested_at: '2026-05-28T01:01:35.340Z'
+    approved_at: '2026-05-28T01:03:06.484Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-28T00:50:34.526Z'
-updated_at: '2026-05-28T01:01:35.341Z'
-pr_ready_for_human: true
+updated_at: '2026-05-28T01:03:06.485Z'
+pr_ready_for_human: false
--- a/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
+++ b/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-05-28T01:01:35.340Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-28T00:50:34.526Z'
-updated_at: '2026-05-28T00:58:26.606Z'
+updated_at: '2026-05-28T01:01:35.341Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
+++ b/codev/projects/bugfix-894-porch-flaky-5s-timeout-in-next/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-894
 title: porch-flaky-5s-timeout-in-next
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-05-28T00:50:34.526Z'
-updated_at: '2026-05-28T00:50:34.527Z'
+updated_at: '2026-05-28T00:55:59.958Z'
--- a/codev/state/bugfix-894_thread.md
+++ b/codev/state/bugfix-894_thread.md
@@ -1,0 +1,32 @@
+# bugfix-894 — Thread
+
+## 2026-05-28 — INVESTIGATE phase started
+
+Issue #894: flaky 5s timeout in `next.test.ts` "emits BUILD tasks for fresh specify phase"
+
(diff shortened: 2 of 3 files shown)


amrmelsayed:
## Architect Review

**Low-risk, root-cause fix.** 56/0 across 3 files (test file + 2 bookkeeping). No production code modified — pure test infrastructure. CMAP unanimous APPROVE.

### Verified

- **Root cause correctly identified.** `next()` → `buildPhasePrompt()` → `getProjectSummary()` → `fetchIssue()` shells out to `gh issue view <id>`. Every test exercising the build path paid one gh subprocess + network round-trip, ~900ms locally, multi-second on CI under load. Vitest's 5s default timeout was the boundary — that's exactly hypothesis #2 from the issue body's fix-options table
- **Proper fix, not a band-aid.** 8-line `vi.mock('../../../lib/github.js', …)` block returning `null` from `fetchIssue`. Tests fall through to existing spec-file / status-title fallback paths — same path they implicitly took before, since fixtures never created real GitHub issues. The mock just removes the dependency that was always unnecessary at this layer
- **Matches existing precedent** — same `vi.mock` pattern is already used in `src/__tests__/project-summary.test.ts`, so this is following the established testing convention for this codebase rather than introducing a new pattern
- **Quantified win.** File runtime: ~11s → ~258ms (~40× faster). Per-test build-path times: 836-2000ms → 2-8ms. Even at 100× CI slowdown, tests stay well under the 5s threshold
- **All 334 tests in `packages/codev/src/commands/porch/__tests__/` still pass** — verified scope
- **No follow-up "test is genuinely slow" issue needed** — the test was never doing meaningful work in the gh call (fallback paths kicked in anyway); the mock just eliminates wasted subprocess startup

### Design pattern worth keeping in mind

The general lesson — "if a test is timing out near vitest's default, look for hidden subprocess/network calls before bumping the timeout" — is a good architectural reflex. The fix here didn't add a `it('...', { timeout: 15_000 })` band-aid; it removed the real source of slowness. Future flake-investigations in this codebase should follow the same instinct: hypothesize-then-verify the dependency, not workaround the symptom.

### Verdict

Approved. Please merge with: `gh pr merge 896 --merge --admin` (solo-architect branch-protection workaround).

This fix is purely in `packages/codev/`, not `packages/vscode/`, so no VS Code CHANGELOG entry needed.

---
Architect review

Issue #1024, opened 2026-06-10
vscode: CLI preflight 400ms timeout triggers false "CLI missing" walkthrough on slow envs (remote SSH, WSL, nvm shims)


amrmelsayed:
## Symptom

Users report that the `Get started with Codev` walkthrough opens on every VSCode startup even though `codev` is installed and at the current version. The Status view shows the CLI row in `missing` state, and every `regCli`-guarded command no-ops with the modal-first feedback toast — even though running `codev --version` from a terminal in the same VSCode window succeeds.

The condition correlates with environments where the `codev --version` spawn has extra latency:

- Remote SSH / VSCode Server.
- WSL or other VM-mediated environments.
- Node version managers (`nvm`, `fnm`, `volta`) that re-resolve the binary on every spawn.
- Slow / saturated I/O (background indexing, AV scanning the codev binary on every read on Windows).
- Network filesystems (NFS / SMB / SSHFS).

## Root cause

The CLI preflight uses a 400ms hard cap on the `codev --version` probe:

```ts
// packages/vscode/src/preflight/preflight.ts:42
/** Hard cap on the `codev --version` probe so a hung binary can't stall startup. */
const VERSION_TIMEOUT_MS = 400;
```

When the probe spawn exceeds 400ms (which is realistic in the environments above), preflight sees no version → `decidePreflight` returns `missing` → the walkthrough opens, the Status row sticks at `missing`, and guarded commands no-op.

The cap was set to protect against a hung binary stalling activation. The intent was sensible; the value was set too tight. Even on a fully-healthy local machine, `codev --version` (Node + Commander startup) runs in ~80-100ms cold — that's only ~5x headroom for ambient slowdown, which isn't enough margin against the legitimate sources of latency above.

Verified surfaces:

- **Source**: `packages/vscode/src/preflight/preflight.ts:42`.
- **Local 3.1.9 bundle** at `~/.vscode/extensions/cluesmith.codev-vscode-3.1.9/dist/extension.js`: appears minified as `lu=400`, called as `uu(r,e,t=lu)` (the `runCodevVersion` function).
- **Remote 3.1.9 bundle** at `~/.vscode-server/extensions/cluesmith.codev-vscode-3.1.9/dist/extension.js` (per a user report): appears minified as `Ad=400`. Same constant; different esbuild rename per build.

## Workaround (per-user, fragile)

Hand-edit the shipped bundle:

```bash
sed -i 's/,Ad=400;/,Ad=30000;/' \
[...]

Issue #991, opened 2026-06-05
terminal: stale tab on a pre-restart terminal id can't self-recover without a manual state re-fetch


amrmelsayed:
## Background

Deferred from #971 (web terminal session-unknown fast-path).

After a Tower restart, persistent (shellper-backed) terminal sessions are reconnected under a **new** terminal id, and the old SQLite row is deleted. Both reconnect paths reassign the id:

- startup reconcile: `packages/codev/src/agent-farm/servers/tower-terminals.ts:646` (`createSessionRaw` → new id) and `:669-672` (delete old row, save under new id)
- on-the-fly reconnect: `:832` and `:868-872` (`deleteTerminalSession(oldId)` → new id)

Consequence: a browser (or VSCode) tab still holding the **old** id's WebSocket URL (`/ws/terminal/<oldId>`) will get a permanent close — `4404` for the web terminal after #971, or the HTTP `404` upgrade rejection for the VSCode terminal — and correctly gives up. But the old id is permanently dead, so a WS retry to the same URL can never recover it.

Recovery today relies on **something re-fetching `/api/state`** (which rehydrates and exposes the successor id) and the dashboard remounting the terminal onto the new `wsPath` (the `Terminal` effect is keyed on `wsPath`). That re-fetch is currently incidental (a user-driven refresh, a focus/poll event) rather than triggered by the give-up itself.

## Work

On a **permanent** terminal close (the `classifyUpgradeError(...) === 'permanent'` branch added in #971), have the dashboard proactively re-fetch workspace state, resolve the session's successor id, and auto-remount the terminal onto the new `wsPath` — so a persistent session that came back under a new id self-heals without a manual refresh.

## Notes

- The same stale-id gap exists for the **VSCode terminal** (give-up + reconnect link, but no automatic successor-id remount). Worth considering the analogous affordance there; if folded in, this becomes `area/cross-cutting`.
- This is a recovery/UX hardening, not a correctness bug — the `4404`/`404` give-up is already truthful (the old id really is gone).

## Acceptance

- After a Tower restart, an open dashboard terminal tab on a persistent session reconnects to its successor session automatically (no manual page refresh), within a small bounded delay.

amrmelsayed:
On it! Working on this with the PIR protocol (plan + dev-approval gates before PR).

Issue #341, opened 2026-02-16
Orphaned shellper and consult processes accumulate over time


waleedkadous:
## Problem

Shellper processes and consult subprocesses accumulate and are never cleaned up. After a few hours with 6 builders running cmap reviews:

- **85 shellper-main.js processes** (expected: 9)
- **91 consult-related processes** (expected: 0 after reviews complete)
- **107 total node processes** (expected: ~27)

### 1. Consult subprocesses never exit (formerly #342)

When builders run `consult --model gemini pr N &` (background cmap), the consult subprocess spawns gemini-cli/codex-sdk/claude as child processes. When the consultation completes, these child processes hang indefinitely.

Likely cause: `run_in_background: true` backgrounds the consult command, and the parent shell doesn't propagate signals to the child process tree on completion.

### 2. Test-spawned shellpers leak

Builder test suites spawn real shellper processes with `detached: true`. Tests finish but shellpers persist as orphans (parented by PID 1/launchd).

PR #340 partially addressed this with afterEach cleanup hooks, but some still leak (~10 per test run).

## Evidence

Manual cleanup on 2026-02-16:
- Before: 85 shellpers, 107 node, 67 claude
- After killing orphans: 7 shellpers, 27 node, 14 claude
- 20 minutes later: 30 shellpers, 50 node (climbing again)

## Possible Fixes

1. Consult commands should properly kill child processes on completion
2. Tower startup should kill any shellper-main.js processes not in its active sessions
3. Test suite afterAll hooks should kill ALL shellper processes spawned during the run
4. Periodic reaper that checks for orphaned shellper processes

PR #1031, opened 2026-06-11
Fix Tower startup checks and shellper cleanup

### LINKED ISSUE #1030: afx tower start/stop should report startup failure and reclaim shellper PTYs on explicit stop
## Bug

Two related Tower lifecycle behaviors can leave operators with misleading status and exhausted PTY resources:

- `afx tower start` can report success even when the Tower daemon immediately fails to serve `/api/status`.
- `afx tower stop` preserves responsive detached shellpers, so explicit operator stop does not necessarily release PTYs. On macOS this can exhaust the PTY pool and surface as `posix_spawnp failed` / `Invalid shellper info JSON`.

## Expected

- `afx tower start` waits for `/api/status` by default and exits non-zero with the Tower log path if the daemon never becomes healthy.
- Fire-and-forget startup remains available via an explicit `--no-wait` option.
- `afx tower stop` stops Tower and then cleans shellpers scoped to Tower's socket directory by default.
- Crash/restart recovery remains conservative: startup reconciliation should still preserve responsive shellpers.
- Operators can opt into current stop preservation behavior with `--preserve-shellpers` and still use the emergency `--force-kill-all-child-processes` path.

## Notes

Related but not fixed by prior persistence work: #274, #832, and PR #999. Those preserve sessions across restarts; this issue is about explicit CLI stop semantics and startup failure reporting.

timeleft--:
## Summary

Fixes #1030.

- Make `afx tower start` wait for `/api/status` readiness by default, with `--no-wait` preserving fire-and-forget startup.
- Make `afx tower stop` clean shellpers scoped to Tower's socket dir by default, with `--preserve-shellpers` retaining restart-preserving behavior.
- Add `SessionManager.killScopedShellpers()` for explicit operator cleanup while keeping `killOrphanedShellpers()` conservative for startup reconciliation.
- Include shellper stderr log context when startup info JSON is missing or invalid, without dumping full env JSON.

## Validation

- `pnpm --filter @cluesmith/codev test --run src/agent-farm/__tests__/tower-command.test.ts src/terminal/__tests__/session-manager.test.ts -t "tower command lifecycle options|kills scoped shellpers|killScopedShellpers|surfaces shellper stderr"`
- `pnpm --filter @cluesmith/codev test`
- `pnpm --filter @cluesmith/codev build`

## Notes

Required external consult was attempted before PR creation. Gemini review was blocked by approval policy because it would disclose uncommitted code to a third-party model without explicit user approval. Codex consult failed before review due to local config: `unknown variant default, expected fast or flex in service_tier`.

### DIFF
--- a/packages/codev/src/agent-farm/__tests__/tower-command.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/tower-command.test.ts
@@ -1,0 +1,76 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const mockKillScopedShellpers = vi.fn();
+
+vi.mock('../../terminal/session-manager.js', () => ({
+  SessionManager: vi.fn(function SessionManagerMock() {
+    return { killScopedShellpers: mockKillScopedShellpers };
+  }),
+}));
+
+vi.mock('../utils/config.js', () => ({
+  getConfig: vi.fn(() => ({ serversDir: '/tmp/codev-test-servers' })),
+}));
+
+vi.mock('../utils/logger.js', () => ({
+  logger: {
+    header: vi.fn(),
+    info: vi.fn(),
+    warn: vi.fn(),
+    error: vi.fn(),
+    success: vi.fn(),
+    kv: vi.fn(),
+    blank: vi.fn(),
+  },
+  fatal: vi.fn((message: string) => {
+    throw new Error(message);
+  }),
+}));
+
+vi.mock('../lib/tower-client.js', () => ({
+  DEFAULT_TOWER_PORT: 4100,
+  AGENT_FARM_DIR: '/tmp/codev-test-agent-farm',
+}));
+
+vi.mock('node:child_process', async () => {
+  const actual = await vi.importActual<typeof import('node:child_process')>('node:child_process');
+  return {
+    ...actual,
+    spawn: vi.fn(),
+    execSync: vi.fn(() => {
+      throw new Error('no process on port');
+    }),
+  };
+});
+
+describe('tower command lifecycle options', () => {
+  beforeEach(() => {
+    mockKillScopedShellpers.mockReset();
+  });
+
+  it('waits for tower start readiness by default', async () => {
+    const { shouldWaitForTowerStart } = await import('../commands/tower.js');
+
+    expect(shouldWaitForTowerStart()).toBe(true);
+    expect(shouldWaitForTowerStart({ wait: undefined })).toBe(true);
+    expect(shouldWaitForTowerStart({ wait: false })).toBe(false);
+  });
+
+  it('cleans scoped shellpers on explicit stop by default when tower is already stopped', async () => {
+    mockKillScopedShellpers.mockResolvedValue(2);
+    const { towerStop } = await import('../commands/tower.js');
+
+    await towerStop({ port: 49_123 });
+
+    expect(mockKillScopedShellpers).toHaveBeenCalledTimes(1);
+  });
+
+  it('preserves scoped shellpers when requested', async () => {
+    mockKillScopedShellpers.mockResolvedValue(2);
+    const { towerStop } = await import('../commands/tower.js');
+
+    await towerStop({ port: 49_123, preserveShellpers: true });
+
+    expect(mockKillScopedShellpers).not.toHaveBeenCalled();
+  });
+});
--- a/packages/codev/src/agent-farm/__tests__/tower-command.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/tower-command.test.ts
@@ -1,12 +1,4 @@
-import { beforeEach, describe, expect, it, vi } from 'vitest';
-
-const mockKillScopedShellpers = vi.fn();
-
-vi.mock('../../terminal/session-manager.js', () => ({
-  SessionManager: vi.fn(function SessionManagerMock() {
-    return { killScopedShellpers: mockKillScopedShellpers };
-  }),
-}));
+import { describe, expect, it, vi } from 'vitest';
 
 vi.mock('../utils/config.js', () => ({
   getConfig: vi.fn(() => ({ serversDir: '/tmp/codev-test-servers' })),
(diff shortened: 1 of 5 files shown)


amrmelsayed (changes requested):
**Request changes.**

The bug in #1030 reproduces, and the readiness-wait and diagnostics changes are the right shape. The `tower stop` default-flip is not — it contradicts an in-code design contract and overlaps with an existing flag.

**`tower stop` killing scoped shellpers by default**

"Tower stop does NOT kill shellpers" is documented in three places in mainline:

- `packages/codev/src/agent-farm/servers/tower-server.ts` (graceful shutdown): "do NOT call `shellperManager.shutdown()` here… let the process exit naturally — OS closes all sockets, and shellpers detect the disconnection and keep running."
- `packages/codev/src/terminal/session-manager.ts` (`shutdown()` docstring): "Disconnect from all sessions without killing shellper processes. Per spec: when Tower intentionally stops, Tower closes its socket connections to shellpers. Shellpers continue running."
- `codev/reviews/991-terminal-stale-tab-on-a-pre-re.md`: "a Tower restart is meant to be transparent to clients — persistent shellper sessions survive."

That's the explicit outcome of #274, #832, PR #999 (session persistence across restarts) and #991 (scoping `tower stop` so it doesn't kill clients' sockets). Three first-party callers depend on shellpers surviving `afx tower stop && afx tower start`:

- `scripts/local-install.sh:46` — every `pnpm -w run local-install` would kill every running builder PTY.
- `packages/vscode/src/tower-starter.ts` — auto-start and the in-extension Restart Tower action.
- `packages/vscode/src/preflight/preflight.ts:394` (`restartTowerAndReprobe`) — divergence-toast Restart.

Nothing inside `towerStop` distinguishes "operator stop" from "scripted restart"; they all shell out to the same command.

Beyond the default-flip, the new `killScopedShellpers()` itself overlaps with the existing `--force-kill-all-child-processes`, which already SIGKILLs every shellper via `pgrep -f shellper-main` + recursive descendant walk (`tower.ts:264-271`, with a comment that calls out the detached-survives-restart contract). The only differences vs the proposed method are (a) socketDir scoping, which only matters across multiple concurrent Tower instances, and (b) SIGTERM-before-SIGKILL, which shellpers don't have meaningful cleanup to use. Neither justifies a new flag.

The PTY-exhaustion symptom in #1030 reproduces, but the cause is elsewhere. `killOrphanedShellpers()` in `tower-server.ts` (~line 411) spares responsive shellpers unconditionally (per #341). Responsive-but-unclaimed shellpers from prior Tower lifetimes accumulate forever — that's the leak.

What we'd like instead:

1. Drop `killScopedShellpers()` and the `--preserve-shellpers` flag. `--force-kill-all-child-processes` already covers the "operator wants to reclaim PTYs" case.
2. Fix the actual leak as a follow-up issue. After SQLite reconciliation completes and the `_reconciling` barrier in `tower-terminals.ts` has gated on-the-fly reconnect, any scoped responsive shellper not in the active sessions map is provably unclaimed and can be reaped. Either `killOrphanedShellpers({ includeResponsive: true })` run as a deferred second sweep, or log-only with a dedicated GC verb (`afx terminal gc`) for operators.

**`tower start` waits for `/api/status` by default**

Right default. Three asks before merge:

1. Drop `--no-wait`. The whole purpose of this change is "don't report success when the daemon isn't serving" — `--no-wait` opts back into exactly that bug. No internal caller uses fire-and-forget (the one programmatic call in `start.ts:38` already passes `wait: true`), and users who need async startup can shell-background (`afx tower start &`). Keep the `wait` parameter on the internal `towerStart()` API for future use; just don't expose it as a CLI flag.
2. Keep `--wait` as a deprecated no-op alias. Removing it is a CLI break for external scripts that pass it explicitly. First-party callers don't pass it, so first-party is safe.
3. `STARTUP_TIMEOUT_MS = 5000` is too aggressive on the default path. Slow-cold-start environments (remote SSH, WSL, AV-scanning Windows, network filesystems) regularly exceed 5s for a healthy startup — we bumped VSCode's `/api/version` probe to 5000ms default in #1024 just for the fast probe. Bump to 15-30s or make it configurable via env / `.codev/config.json`.

**Shellper stderr in diagnostic error**

Right direction, addresses the `Invalid shellper info JSON` symptom in #1030 directly. One concern with redaction:

The regex `/"env"\s*:\s*\{[^}]*\}/` is brittle. `[^}]*` terminates at the first `}` — if any env value contains a literal `}` (legal in JSON strings), the regex stops at the inner brace and the rest of the env block leaks. It also only scrubs `env`; `args` can carry secrets too (CLI flags routinely embed tokens). Suggested approach: parse the snippet as JSON, walk the structure, redact known-sensitive keys (`env`, `args`), stringify back. On parse failure, omit the snippet. Not a merge blocker, but worth doing.

**Summary**

Drop the `tower stop` changes entirely from this PR. Address the start-side asks (drop `--no-wait`, keep `--wait` as deprecated alias, bump the timeout) and the redaction ask. Otherwise the PR is a clean improvement and we'd merge it.

We ran a CMAP review on the diff (Claude + Codex; Gemini lane skipped on an unrelated agy issue). Both reviewers converged on the same reading.

amrmelsayed:
Follow-up filed for the orphan-reaper change referenced in the review: #1038. The shape of the fix (post-reconciliation second sweep with `includeResponsive: true`, or an explicit `afx terminal gc` verb) is captured there with the testing surface and the #341 context, so the underlying leak in #1030 has a tracked home independent of what happens with this PR.

timeleft--:
Addressed the requested changes in commit `3e1d6d5b`.

Summary:
- Dropped the `tower stop` behavior change entirely.
- Removed `--preserve-shellpers`, `killScopedShellpers()`, and the related tests.
- Kept `--wait` as a deprecated no-op alias and removed the exposed `--no-wait` CLI path.
- Increased the Tower readiness wait from 5s to 30s.
- Hardened shellper startup diagnostics by parsing stdout JSON and redacting `env` and `args`; invalid JSON stdout is omitted while stderr log context is still included.
- Read #1038 and #341 and preserved the #341 responsive-shellper safety behavior; the startup orphan-reaper work remains scoped to #1038.

Validation:
- `pnpm --filter @cluesmith/codev test --run src/agent-farm/__tests__/tower-command.test.ts src/terminal/__tests__/session-manager.test.ts -t "tower command lifecycle options|surfaces shellper stderr|redacts shellper startup"`
- `pnpm --filter @cluesmith/codev test`
- `pnpm --filter @cluesmith/codev build`

amrmelsayed (approved):
**Approving — all iteration-1 asks addressed cleanly.**

Verifying each ask against commit `3e1d6d5b`:

**`tower stop` changes dropped**
Zero remaining references to `killScopedShellpers`, `--preserve-shellpers`, or any `towerStop()` behavior change. The non-force stop path is unchanged from main — SIGTERM the Tower PID(s), preserve shellpers per the documented contract. `--force-kill-all-child-processes` continues to handle the operator "tear everything down" case as before. The underlying PTY-exhaustion leak stays scoped to follow-up #1038.

**`tower start` readiness wait**
- `--no-wait` removed from the CLI surface ✓ (the handler in `cli.ts` now hardcodes `wait: true` regardless of `options.wait`).
- `--wait` retained as a deprecated no-op alias ✓ (description updated, scripts passing `--wait` continue to parse successfully).
- Internal `towerStart({ wait: false })` API still available for future programmatic callers via the new `shouldWaitForTowerStart` helper.
- `STARTUP_TIMEOUT_MS` bumped 5000 → 30000 ✓.

**Shellper stderr diagnostics**
The regex-based redaction is replaced with `safeDiagnosticSnippet` → `JSON.parse` → recursive `redactDiagnosticValue` walk. Both `env` and `args` are redacted (not just `env` as in iteration 1). The redact-stringify-truncate order is correct (`JSON.stringify(redact(...)).slice(0, 1000)`), so the 1000-char cap can never expose an unredacted leak in a malformed truncated JSON. On parse failure the snippet is omitted entirely.
The new test `'redacts shellper startup stdout env and args diagnostics'` directly covers the regex-killer case from iteration 1: an env value containing `}` (`'do-not}log'`). Test asserts the redacted output is `{"env":"[redacted]","args":"[redacted]","nested":{"args":"[redacted]"}}` and that none of the secret values leak through.

**Bonus catch**
The new `settled` flag in `readShellperInfo` fixes a real race: previously `exit` and `end` could both reach `reject` if they fired close together. With `settled`, only the first one wins.

**One minor non-blocker**
No explicit CLI-parse test asserting `afx tower start --wait` succeeds without erroring. The implementation is straightforward enough and the back-compat is obvious from the Commander definition, so not worth blocking on.

Diff is meaningfully tighter (+189/-14, down from +313/-12). Merging.

Issue #737, opened 2026-05-13
VSCode terminals: TUI render artifacts on first open — initial Pseudoterminal dimensions are ignored


amrmelsayed:
## Symptom

When opening an architect or builder terminal in VSCode (Codev extension), Claude Code's TUI renders with overlapping content: the streaming response and the input box's cursor occupy the same screen region, producing artifacts like a cursor block (\`▌\`) appearing in the middle of transcript text.

The artifact disappears immediately if the user resizes the VSCode terminal panel by even one pixel.

## Reproduction

1. Open VSCode in a Codev workspace
2. Open the architect terminal (sidebar click, command palette, or keybinding)
3. Have Claude Code stream a response longer than ~24 lines
4. Observe: streaming content and input-box cursor render on top of each other

Resizing the panel after the artifact appears clears it.

## Root cause

\`packages/vscode/src/terminal-adapter.ts:37-44\` — the \`open()\` callback receives \`initialDimensions\` from VSCode but discards them (parameter is prefixed \`_\`):

\`\`\`ts
open(_initialDimensions: vscode.TerminalDimensions | undefined): void {
  this.writeEmitter.fire('');
  this.connect();
}
\`\`\`

Consequence:
1. Tower spawns the PTY at \`node-pty\`'s default (80×24)
2. VSCode renders the terminal at the real panel size (e.g. 200×60)
3. Claude Code queries PTY size → gets 80×24
4. Claude Code anchors its input box at row 24
5. Streaming transcript flows past row 24
6. In the real 60-row xterm viewport, input box and transcript collide

\`setDimensions()\` only fires on **subsequent** resizes, not on initial open — so the desync persists until the user manually nudges the panel.

## Fix sketch

In \`open()\`, store \`initialDimensions\`. In \`ws.on('open')\` after auth is sent, call \`sendResize(cols, rows)\` if initial dimensions are set. Keep the existing \`replaying\` guard so resume-from-replay still defers correctly.

~10 LOC change in \`packages/vscode/src/terminal-adapter.ts\`.

## Test

Unit: assert that \`sendResize\` is invoked once on WebSocket open when \`open()\` was called with non-null dimensions.

Manual: open a fresh terminal at a non-default panel size, stream >24 lines of content, confirm no overlap without resizing.

## Scope

AIR — single-file bugfix, no spec/plan needed.

## References

- \`packages/vscode/src/terminal-adapter.ts:37-44\` — the bug
[...]

Issue #1032, opened 2026-06-12
agy/Gemini consult lane goes off-task on structured reviews — emits no VERDICT, breaking the porch consult loop


waleedkadous:
## Problem

The Gemini consult lane (Antigravity `agy`, migrated in #778) **runs successfully but does not produce a structured review verdict** when asked for one (`consult -m gemini --type <spec|plan|impl|phase|pr|integration>`). Instead of returning `VERDICT: APPROVE/REQUEST_CHANGES` with findings, it **wanders off-task** — exploring the repo and, notably, latching onto the local **`google-antigravity-sdk` skill** and summarizing *that* instead of reviewing the target.

Because no verdict comes back, porch's review loop treats it as `REQUEST_CHANGES` (see companion porch bug) and **loops every phase**. This is **not** an auth problem — `agy` 1.0.6 authenticates fine (OAuth, exit 0); the lane is *behaviorally* broken for structured reviews.

## Reproductions

1. **Architect smoke test (2026-06-09):** `consult -m gemini --prompt "Reply with exactly: AGY_LANE_OK <year>. Nothing else."` → completed in 47.6s, exit 0, but **ignored the instruction** and produced a multi-section "sandbox state" summary that referenced `google-antigravity-sdk` SKILL.md. Agentic exploration overrode the prompt.
2. **Builder spir-987, phase_1 (2026-06-11/12):** iter1 timed out (skipped, fine); **iter2 ran but went off-task** — explored an unrelated `google-antigravity-sdk`, returned **no VERDICT**. porch defaulted to `REQUEST_CHANGES` → loop (reached iter3 of `max_iterations=8`). Codex + Claude both `APPROVE`'d the same phase cleanly (build + 5 tests green).

## Suspected root cause

`agy`'s agentic/`--sandbox` mode lets Gemini self-direct, and the presence of the `google-antigravity-sdk` skill in context appears to hijack it (it keeps exploring that). The review prompt's "return a VERDICT" instruction is not being honored / not constraining the run.

## Expected

- A `--type` review consult on the gemini lane returns a parseable `VERDICT` (APPROVE / REQUEST_CHANGES + findings), same contract as codex/claude.
- agy should be constrained to the review task (less free-roaming agentic exploration for `--type` reviews), and should not be derailed by unrelated local skills.
- If the lane genuinely can't produce structured verdicts reliably, consider gating gemini out of the default review model set until it can.

## Impact / workaround

[...]

PR #1050, opened 2026-06-15
tower + vscode: fix terminal freeze (oversized-replay storm) and blank-on-open

### LINKED ISSUE #1047: tower + vscode: terminal freeze from oversized-replay reconnect storm (unbounded no-newline buffers + client backpressure loop)
## ✅ CONFIRMED ROOT CAUSE (2026-06-15)

Captured data (PTY disk logs + a 42,143-line VSCode terminal log + code trace) confirms the mechanism end to end. The original hypotheses below (Tower-wide WS pump degradation, broadcast-list leak, socket backpressure) are **superseded** — the real cause is an **oversized-replay reconnect storm** caused by two independent defects that chain:

### Tower/shellper side — the root cause
Two terminal buffers bound themselves by **newline count**, so a full-screen TUI that redraws in place (alternate screen buffer, cursor-addressing + `\r`, almost no `\n`) makes them grow without limit:
- `RingBuffer.partial` (`packages/codev/src/terminal/ring-buffer.ts:36-51`): `combined = partial + data; combined.split('\n')` every frame, keeping the trailing fragment. No `\n` → `partial` grows unbounded and every frame re-scans the whole thing (O(partial)/frame → O(n²)/session). No byte cap.
- `ShellperReplayBuffer` (`packages/codev/src/terminal/shellper-replay-buffer.ts:45,58`, fed at `shellper-process.ts:143`): evicts only `while lineCount > maxLines`, so with zero newlines it never evicts.

On every (re)connect Tower sends the buffer as **one frame**: `ws.send(encodeData(getAll().join('\n')))` (`tower-websocket.ts:62-65`) → a multi-MB replay payload.

### VSCode side — the amplifier
The terminal adapter's backpressure guard trips on the **replay frame itself** and reconnects with **no backoff**:
- `handleData` (`packages/vscode/src/terminal-adapter.ts:300-308`): `queuedBytes += payload.length; if (queuedBytes > MAX_QUEUE /*1 MB*/) reconnect()` — fires before rendering anything.
- `reconnect()` (`terminal-adapter.ts:281-296`) calls `backoff.reset()` and reconnects instantly → Tower re-sends the same oversized replay → infinite loop.

### Why it presents the way it does
- **CPU ~93% (one core), all terminals freeze at once:** Tower re-serializes a multi-MB replay thousands of times on its single event loop, starving every other terminal. Input stays responsive (tiny), output is frozen — matching the stray-input-char screenshot.
- **Memory barely moves:** the cost is CPU + GC churn, not retained memory.
- **Restart is NOT reliable** (corrects the original title): `ShellperReplayBuffer` is also unbounded, so after restart the shellper replays the multi-MB blob, Tower re-seeds an oversized `partial`, and the storm resumes — unless the offending session has gone quiet/ended.

### Captured evidence
- Storming terminal `f2dc55d1` disk log: **1.9 MB, 0 newlines** → replay 1.9 MB **> 1 MB** `MAX_QUEUE`.
- Incident-window log `f02bedcb`: **15 MB, 0 newlines** (1.5M ESC, 164K CR, 0 LF — a Claude alt-screen redraw stream). Normal sessions: ~20 bytes/line.
- VSCode log (~58 min): **14,026** "WebSocket connected" / **14,015** "Backpressure exceeded 1 MB" / **14,026** "Connecting to" (1:1:1); **14,017** of the reconnects target the single terminal `f2dc55d1`.
- Threshold is low: only **~1 MB** of no-newline output triggers the storm.

### Fix (see `codev/plans/1047-...` on `builder/pir-1047`)
Three coordinated fixes: **(A)** `RingBuffer.pushData` scans only new data + byte-caps `partial` (< `MAX_QUEUE`); **(B)** `ShellperReplayBuffer` gains a byte cap; **(C)** the VSCode backpressure path must not infinite-loop on an oversized replay (backoff + re-trip guard, ideally excluding the replay burst from the live-backpressure budget). A and B remove the oversized replay at the source; C makes the client structurally immune. Scope now spans the terminal core **and** the VSCode extension (re-labeled `area/cross-cutting`).

---

## Symptom

Over time, architect AND builder PTY terminals stop rendering output and accepting input that propagates back to the visible UI. The symptom affects ALL terminals at once, not individual sessions. The only known recovery is `afx tower stop && afx tower start` (the default non-force variant).

Recent screenshot of an affected architect terminal: only `Ran 1 shell command` remained rendered, plus a stray input character (the user's `e`) at the bottom-left corner where the prompt frame should be. The TUI box-drawing of the prompt panel was gone, but the typed character + cursor block still appeared, suggesting the input side still propagates but the render-back side is broken.

## What the diagnosis rules out

Per #1030 / PR #1031 (just landed), the default `afx tower stop` does NOT kill shellpers. Shellpers and the AI CLI processes they host survive the restart. Yet restart restores responsiveness, so:

- The AI CLI is not hung (it would still be hung after a non-force restart).
- The shellper process is not crashed (it survives the restart too).
- The PTY itself is not dead.

What actually changes during the default-stop / start cycle:

1. Tower's WebSocket connections drop.
2. The VSCode extension's terminal client detects disconnect and reconnects to the new Tower process.
3. New Tower runs `reconcileTerminalSessions`, reattaches to the still-alive shellpers, and resumes streaming.

So the bug lives in **the Tower-to-client streaming layer or the per-terminal client-list state that Tower maintains**, not in the AI CLI / shellper / PTY.

## Tower-side log behavior

Tower itself stays "alive" in the process sense during the hang: SSE heartbeats keep firing every 30s, cron tasks keep running, the team scanner keeps iterating. So this is not a process-wide event-loop deadlock. It's specifically the WS-to-terminal pump that's wedged.

## Top hypotheses (ordered by fit with the all-terminals-at-once pattern)

1. **Tower-wide WS send-pump degradation.** Memory growth, GC stalls, or an unbounded handler/listener leak that gradually starves the WS output loop. Affects all terminals because they share the same Tower process. Possibly worsened by repeated errors that throw past their handler scope without cleaning up (the cron evaluator `ReferenceError` filed separately is a candidate contributor).
2. **Leaked closed-connection entry in the per-terminal broadcast list.** Tower iterates its client list to fan out PTY output; a dead WS still in the list blocks the broadcast for everyone. Restart drops the list.
3. **Shellper-to-Tower unix-socket backpressure.** Shellper streams output to Tower over its unix socket; if Tower stops draining fast enough, shellper's pipe fills and stops producing. Restart drops the connection, shellper opens a fresh one on reattach.

## Diagnostic plan for the next occurrence

1. While hung, observe `afx tower log -f` — does the SSE heartbeat keep firing every 30s? If yes, event loop is alive; the WS pump is the culprit (H1 or H2). If no, deeper event-loop starvation.
2. `ps -p <tower-pid> -o pid,vsz,rss,pcpu` while hung. RSS growth into multi-GB territory points at H1.
3. `lsof -p <tower-pid> | wc -l` while hung. A climbing fd count points at leaked connections (H2-adjacent).
4. From a side terminal: `afx send architect "echo ping-from-side"`. Watch the Tower log for the send event AND watch the affected architect tab for the echoed output. If Tower logs the send but the tab never shows `ping-from-side`, it's pinned to the Tower-to-VSCode WS direction (not the input side).
5. Open the same architect session in the Tower dashboard at `http://localhost:4100`. If the dashboard renders fresh output while the VSCode tab stays frozen, the VSCode-side xterm.js renderer is in the loop too (downgrades H1/H2 to "VSCode-extension-specific WS handling" instead of "Tower-side").

## Why this is PIR, not BUGFIX

Root cause is not yet known — three plausible mechanisms with overlapping symptoms. Investigation phase (capture diagnostics, narrow to a single mechanism, then design the fix) is the right shape. Pre-PR running-on-real-workload verification is load-bearing because the bug only manifests after time-in-use.

## Out of scope

- The unrelated `ReferenceError: exitCode is not defined` cron evaluator bug visible in the same Tower log window. Filed separately.
- Any change to default `tower stop` shellper-killing behavior. Shellpers must continue surviving restart per #274 / #832 / #999 / #991.

amrmelsayed:
# PIR Review: Terminal freeze + blank-on-open (oversized-replay storm and missing connect-time redraw)

Fixes #1047

## Summary

Two client/server terminal bugs were fixed. (1) **Freeze:** a full-screen TUI (Claude's UI) redraws in place and emits almost no newlines, so Tower's reconnection buffers — bounded by *line count* — grew unbounded; the resulting multi-MB replay overflowed the VSCode client's 1 MB receive budget, and the client's "disconnect-and-reconnect for replay" response re-fetched the same oversized snapshot in a tight loop (~14k cycles/hour), pegging one CPU core and starving every terminal. (2) **Blank-on-open:** a freshly-attached terminal could render a blank pane until a manual window resize, because the connect-time resize could be a same-size no-op and a full-screen TUI only repaints on a size *change*. Tower now scans only new output (killing the O(n²) re-scan) and brackets replay with `pause`/`resume` so it is paced and excluded from backpressure; the client drops live output under overload instead of reconnecting, reconnects with a `?resume=<seq>` delta, and forces a guaranteed redraw shortly after connect (mirroring the web dashboard).

## Files Changed

- `packages/codev/src/terminal/ring-buffer.ts` (+39 / -…) — `pushData` scans only the incoming chunk (O(|data|), not O(|partial|)); `partial` kept whole for faithful replay; `partialBytes` getter for observability.
- `packages/codev/src/terminal/pty-session.ts` (+14) — `attachShellper` idempotency (drop prior client listeners on re-attach); `partialBytes` passthrough.
- `packages/codev/src/terminal/pty-manager.ts` (+30 / -…) — `createSessionRaw` tears down a colliding session before overwrite; `inspectPartials()` for the monitor.
- `packages/codev/src/agent-farm/servers/tower-websocket.ts` (+8 / -…) — bracket replay with `pause`/`resume` control frames.
- `packages/codev/src/agent-farm/servers/tower-server.ts` (+32) — periodic ring-buffer-partial monitor (observability).
- `packages/vscode/src/terminal-adapter.ts` (+124 / -…) — replay excluded from backpressure; live overload **drops** instead of reconnecting; `?resume=<lastSeq>` delta reconnect; post-connect repaint nudge.
- `packages/codev/src/terminal/__tests__/ring-buffer.test.ts` (+25), `packages/codev/src/agent-farm/__tests__/tower-websocket.test.ts` (+14 / -…), `packages/vscode/src/__tests__/terminal-adapter.test.ts` (+137 / -…) — unit coverage.
- `packages/vscode/CHANGELOG.md` (+7) — user-facing entries for both fixes.
- `codev/resources/arch.md` (+4), `codev/resources/lessons-learned.md` (+3) — see sections below.

(`git diff --stat`: 12 files, +410 / -27, excluding plan/thread/status artifacts.)

## Commits

- `8868ac86` [PIR #1047] Fix A+E: byte-cap RingBuffer.partial (scan-only pushData) + reconnect listener hygiene
- `7d42dac9` [PIR #1047] Fix B: byte-cap ShellperReplayBuffer (bounds no-newline replay across restart)
- `f7e6b3a2` [PIR #1047] Fix C+D: pause/resume replay bracket, drop-not-reconnect on live overload, resume-delta reconnect
- `9f282745` [PIR #1047] Instrumentation: log terminal ring-buffer partial sizes on a Tower monitor interval
- `757a75a5` [PIR #1047] Update tower-websocket test for pause/resume replay bracket
- `2f5e6351` [PIR #1047] docs: user-facing CHANGELOG entry for terminal-freeze fix
- `b05b73f2` [PIR #1047] Hoist replay-cap resolution out of constructor args; document the three distinct size limits
- `02ff5546` [PIR #1047] Remove byte caps: keep partial/replay whole for faithful TUI replay
- `7100a5a3` [PIR #1047] Re-assert terminal size after replay (Option A) + temporary diag logging
- `81c4a83b` [PIR #1047] Force a post-connect redraw nudge (mirror web client's SIGWINCH-on-connect); revert Option A
- `6d994f72` [PIR #1047] docs: changelog entry for terminal blank-on-open fix
- `cc7fbdc7` [PIR #1047] Strip temporary [#1047-diag] diagnostic logging

(Plus the review/test-coverage commit carrying this file.)

## Test Results

- `pnpm build`: ✓ pass
- `pnpm --filter @cluesmith/codev test`: ✓ 3308 passed / 0 failed / 48 skipped
- `packages/vscode` adapter suite: ✓ 20 passed (including 5 storm-prevention + 2 repaint-nudge tests new for #1047); full VSCode suite previously green at 414.
- Manual verification (human, `dev-approval` gate, macOS VSCode against the worktree Tower): confirmed via the temporary `[#1047-diag]` client log that there is no connect→backpressure→reconnect storm, `seq` heartbeats stay clean, and — after the nudge landed — terminals **paint on open without a manual resize**.

## Architecture Updates

**COLD (`codev/resources/arch.md`)** — added a "Terminal reconnect/replay contract (#1047)" subsection under Shellper Lifecycle documenting the now-load-bearing behaviors: replay bracketed by `pause`/`resume` (excluded from client backpressure), `?resume=<seq>` delta reconnect, the deliberately-unbounded `partial` (faithful TUI replay) with CPU bounded by scan-only `pushData`, the client's connect-time forced redraw, and drop-not-reconnect under live overload.

No HOT (`arch-critical.md`) change: these are subsystem-level protocol details, not a top-tier always-injected system-shape fact, and the hot file is at its cap.

## Lessons Learned Updates

**COLD (`codev/resources/lessons-learned.md`, Debugging section)** — three `[From 1047]` entries: (1) cross-client differential diagnosis — a symptom in one client but not another sharing the backend localizes the bug to the client layer (web rendered fine → the app *was* painting → the VSCode adapter was at fault), and a full-screen TUI only repaints on a size *change*; (2) a buffer bounded by newline count is unbounded for a newline-free stream, with the on-disk PTY logs ("longest run without `\n`") as the decisive evidence, and the caution that front-trimming a buffer corrupts TUI replay; (3) a backpressure relief valve that re-fetches the overflowing payload is an infinite loop — drop ephemeral output, don't reconnect.

No HOT (`lessons-critical.md`) change: these are debugging recipes, not a top-tier cross-cutting rule, and the hot file is at its cap.

- **The dropped byte caps (history).** Early commits added byte caps to `RingBuffer.partial` and `ShellperReplayBuffer`; they were **reverted** (`02ff5546`) after dev-approval testing showed front-trimming corrupts a full-screen TUI's replay (its alt-screen state lives in the cumulative stream from the alt-screen-enter onward). The freeze is fixed by *scan-only* `pushData` + the replay bracket, not by bounding the buffer. Net effect: `partial`/replay are functionally back to `main`'s faithful-but-unbounded behavior; the new `tower-server.ts` monitor logs partial size so unbounded growth (a known, issue-rated-minor trade-off) is observable.
- **The repaint nudge mechanism** (`terminal-adapter.ts`, `scheduleRepaintNudge`). It is a `rows-1` → `rows` size delta 500ms after connect, **gated** on `renderedSinceConnect` so a reconnect that already painted via replay doesn't reflow. The brief 1-row intermediate frame is intentional (guarantees a real `TIOCSWINSZ` delta even when the PTY is already at the target size). Confirm the gating and that it's cleared on close/reconnect (`resetStreamState`).
- **`#737` and the nudge are kept separate** (deliberate): `#737`'s on-open resize sets the correct size immediately (no flicker, common case); the nudge is the delayed backstop for the same-size no-op case. They're complementary, not redundant.
- **Single-pass CMAP caveat (PIR):** any `REQUEST_CHANGES` from the 3-way consult is not independently re-reviewed — please scrutinize the disposition at this gate.

## How to Test Locally

- **View diff**: VSCode sidebar → right-click builder `pir-1047` → **Review Diff**.
- **Run dev server**: VSCode sidebar → **Run Dev Server**, or `afx dev pir-1047`.
- **What to verify**:
  - Open an architect/builder terminal running Claude's full-screen UI and let it redraw past ~1 MB of no-newline output; Tower CPU stays low and there is **no** connect→backpressure→reconnect storm (contrast the original ~14k-cycle capture).
  - Open a terminal that previously blanked — it **paints on open** without a manual window resize.
  - Reconnect a tab (close/reopen) → clean repaint; the reconnect URL carries `?resume=` (delta).
  - Optional: leave Tower running for hours on a real workload and confirm CPU stays flat (the true end-to-end soak; cannot fit a gate session).

## Flaky Tests

None skipped. (Earlier full-suite runs showed environmental failures in `session-manager`/`adopt`/`hot-tier` integration tests when `dist/` and the copied skeleton were absent; a full `pnpm build` resolves them — they are not related to this change and pass after build.)


waleedkadous (approved):
## Architect Integration Review — APPROVE

Outstanding work, Amr. This is one of the cleanest hard-bug writeups I've seen in this repo — and it fixes the terminal-freeze that's been causing real damage (it's almost certainly the same oversized-replay storm behind the agent-death/"terminal extinction" events we hit during the node-26 window).

I independently verified the core code and the CMAP disposition rather than taking the summary at face value:

**Diagnosis** — evidence-backed and airtight: line-count-bounded buffers (`RingBuffer.partial`, `ShellperReplayBuffer`) grow unbounded on a no-newline full-screen-TUI stream → multi-MB replay overflows the client's 1 MB budget → reconnect-for-replay re-fetches the same blob in a ~14k-cycle/hour loop (the 1:1:1 connect/backpressure/reconnect ratio in the captured logs nails it).

**Fix** — sound where it counts:
- `ring-buffer.ts`: `pushData` scans only the incoming chunk (`indexOf`), killing the O(n²) re-scan that pegged the CPU. Correct.
- `terminal-adapter.ts`: the client is made *structurally* immune — drop-not-reconnect under live overload, replay excluded from backpressure via the `pause`/`resume` bracket, `?resume=<seq>` delta, and the post-connect repaint nudge for blank-on-open. Good edge handling (successor-session `lastSeq` reset; `replaying` cleared on mid-replay drop).

**CMAP disposition** — disciplined. Codex's REQUEST_CHANGES (byte-addressable seq + shellper byte cap) is correctly **rebutted, not fixed**: byte-capping front-trims the buffer and corrupts a full-screen TUI's faithful replay — the exact blank-terminal regression caught at dev-approval. Re-adding it reintroduces an *observed* bug. The one legitimately-shared point (overstating `?resume=` as universal) was fixed, and Claude's Fix-E test gap closed. (Gemini failed again with an empty-sandbox clarification prompt — a live repro of #1032; effectively a Codex+Claude review.)

**Accepted trade-off:** `partial`/replay stay unbounded in memory for no-newline streams (now monitored). The acute CPU-storm failure is fully fixed, and the alternative reintroduces a worse user-visible regression — right call. Possible future follow-up: bound at the alt-screen-enter boundary instead of front-trimming (caps memory *and* preserves replay). Not a blocker.

Approving. Thank you for the rigor on this one.

Issue #786, opened 2026-05-20
Multi-architect feature is underbaked — gaps in lifecycle, persistence, and UX


waleedkadous:
## Summary

The multi-architect feature (sibling architects added via `afx workspace add-architect`) shipped its primitive in v3.0.5 (#755), dashboard tabs in v3.0.6 (#761), and a critical routing fix in v3.0.8 (#774). But the feature is not yet a coherent, well-thought-through product. There are gaps in lifecycle management, persistence semantics, and UX that an end user discovers immediately when they try to drive it.

This issue is the umbrella for a SPIR-protocol pass to design and ship the missing pieces as a cohesive unit. Goal: by the end, a user can add, manage, evict, and recover sibling architects with the same fluency they have with builders.

### 1. No way to remove a sibling architect
- `afx workspace add-architect` exists; `afx workspace remove-architect <name>` does not.
- Manual workarounds today: kill the terminal from the dashboard sidebar (left pane only — see #2), or restart Tower entirely (nukes all workspaces).
- A first-class `remove-architect` CLI + dashboard affordance is needed.

### 2. Right-pane terminals (builders, shells) have no close button; sibling architects similarly cannot be closed from where they live
- The dashboard right pane (where builders and shells render) has no X / close affordance on tabs. Only the left pane does.
- Sibling architect tabs (in the multi-architect tab strip introduced in v3.0.6 #761) likewise have no close UI. The architect *is* a closable entity (unlike `main`, which is workspace-defining), so it should have one.
- This is a broader UX gap that affects more than architects, but architects make it salient.

### 3. Sibling architects are not persisted in `state.db`
- Confirmed empirically: shannon's `.agent-farm/state.db` `architect` table is empty even though Tower's in-memory `getWorkspaceTerminals()` map correctly has both `main` and `ob-refine`.
- Implication: sibling architects exist only in Tower's process memory. A Tower restart wipes them. The user has to re-run `add-architect` every time Tower goes down.
- This is the opposite design from how builders work (builders persist across Tower restarts via `state.db` + shellper auto-rebind).
- There IS an `architect` table in the schema (with `id, pid, port, cmd, started_at, terminal_id`) — the row for `main` gets written by `workspace start`, but sibling architects added via `add-architect` never write rows. So the table exists but the code path skips it for siblings.
[...]

PR #1109, opened 2026-06-28
[AIR #1108] vscode: open architect terminal on Agents-view group-header click

### LINKED ISSUE #1108: vscode: clicking an architect group header in the Agents view should open that architect's terminal (parity with the builder-row click affordance)
## Problem

In the Agents view shipped via #1106 (PR #1106), clicking a **builder row** opens that builder's terminal while the chevron on the left continues to handle expand / collapse — two independent click targets, the row body for the action, the chevron for the structure. That pattern doesn't extend to architect group headers in the new architect-axis grouping mode: clicking an architect header (e.g. `MAIN`, `VSCODE`) only toggles expand / collapse via the row, with no way to open that architect's terminal from the same surface. To launch the architect terminal the user has to switch to the Workspace > Architects subsection.

The asymmetry is the friction. The architect-axis grouping makes the architect first-class in the work view, but the row itself is then a less interactive surface than the builder rows nested under it — which inverts the visual hierarchy.

## Expected

The architect group header should behave like a builder row: click-on-row opens that architect's terminal; the chevron remains a separate target for expand / collapse. The existing `codev.openArchitectTerminal` command already accepts an optional architect name argument (per #786 Phase 6, see `extension.ts`), so wiring it up is mechanical.

## Source

The architect-axis groups are produced by `architectGrouping()` in `packages/vscode/src/views/builder-grouping.ts`. The header `TreeItem` is built (along with the area / stage headers) in `packages/vscode/src/views/builders.ts` — that's where the `command` property needs to be set on the architect headers specifically (not the area / stage headers, which don't have a launch affordance).

Required wiring on the architect-header TreeItem:

```ts
item.command = {
  command: 'codev.openArchitectTerminal',
  title: 'Open Architect Terminal',
  arguments: [groupKey],  // architect name
};
```

The `collapsibleState` (Collapsed / Expanded) stays as today — VS Code routes chevron clicks to the expand / collapse gesture independently of the row's `command`, so both work without conflict (this is the same pattern the builder rows already use, where `command` opens the terminal and the chevron expands the changed-files list).

## Scope notes

- **Only the architect-axis group headers get the `command`.** Stage and area headers (`SPECIFY`, `TOWER`, etc.) don't represent a launchable entity — they remain pure grouping containers.
- **`main` for null-owner builders is special.** Per `architectGrouping()`, null `spawnedByArchitect` folds into the `main` bucket. The `main` group header should open the main architect terminal (consistent with the fold).
- **Stale-owner architects** (a builder's `spawnedByArchitect` references an architect no longer in the live roster) would produce a group header that, when clicked, fails to find the architect terminal. `codev.openArchitectTerminal` should already handle this case gracefully (the same path the right-click action takes from Workspace > Architects); if it doesn't, that's a separate `NotFound` UX bug worth tracking under this issue's lifecycle.

## Out of scope

- Touching area / stage group headers — they don't represent launchable entities.
- Right-click context menu on architect group headers (e.g. "Open Worktree Folder" parity). Could be a follow-up if the affordance matters, but the click-to-open is the primary friction.
- Adding launch affordance from area / stage modes (where architects aren't visible as headers). Workspace > Architects remains the surface for that case.

## Protocol

AIR. Mechanical UI wiring, no architectural decision, well-scoped. The change is roughly a 5-line set of `item.command = { ... }` on the architect-header construction site in `builders.ts`, plus a small unit test asserting the architect header carries the expected command argument.

## Related

- **#1106** Agents view: three-way group-by axis (the PR this issue extends).
- **#1104** Agents view design (the umbrella).
- **#786** Phase 6 — established the `codev.openArchitectTerminal` command's optional-name argument that this issue depends on.

amrmelsayed:
Closes #1108.

## Summary

In the Agents view's **architect-axis** grouping mode, clicking an architect
group header (e.g. `MAIN`, `VSCODE`) now opens that architect's terminal,
matching the builder-row affordance. The chevron continues to toggle
expand/collapse independently (VS Code routes the row `command` and the chevron
gesture separately — the same pattern builder rows already use).

This removes the asymmetry where the architect-axis grouping made the architect
first-class, but its header was a *less* interactive surface than the builder
rows nested under it.

## Changes

- **`packages/vscode/src/views/builders.ts`** — `rootChildren()` now sets
  `item.command = { command: 'codev.openArchitectTerminal', title: 'Open Architect Terminal', arguments: [g.key] }`
  on each group header **only when the active grouping axis is `architect`**
  (`grouping.id === 'architect'`). `g.key` is the architect name.
- **`packages/vscode/src/__tests__/builders-architect-header-command.test.ts`** —
  new unit tests.

## Key decisions

- **Scoped to architect headers only.** Stage (`SPECIFY`, …) and area
  (`TOWER`, …) headers name no launchable entity, so they stay command-less
  containers. The gate is `grouping.id === 'architect'`, checked once per render.
- **`main` fold preserved.** A null `spawnedByArchitect` folds into the `main`
  bucket in `architectGrouping()`; the `main` header therefore opens the main
  architect terminal (`arguments: ['main']`), consistent with the fold.
- **No command-handler change.** `codev.openArchitectTerminal` already accepts
  the optional architect-name argument (#786 Phase 6) and already handles a
  stale / not-found owner gracefully — it shows a `No '<name>' architect found`
  warning (the same path the Workspace > Architects right-click takes). So the
  stale-owner edge from the issue's scope notes needs no extra work here.
- **No `package.json` / menu changes.** This is a row `command` (left-click),
  not a `view/item/context` menu entry, so no contribution-point wiring is
  needed.

## Test plan

- `pnpm test:unit` (vscode package): **43 files / 516 tests pass**, including the
  4 new tests:
  - architect headers carry `codev.openArchitectTerminal` with the architect
    name as the argument;
  - a null `spawnedByArchitect` yields a `main` header that opens the main
    terminal;
  - stage headers carry no command;
  - area headers carry no command.
- `porch check`: ✓ build ✓ tests.

## Out of scope (per issue)

- Right-click context menu on architect headers (e.g. "Open Worktree Folder").
- Launch affordance from stage / area modes (Workspace > Architects remains the
  surface there).

### DIFF
--- a/codev/projects/1108-vscode-clicking-an-architect-g/status.yaml
+++ b/codev/projects/1108-vscode-clicking-an-architect-g/status.yaml
@@ -1,0 +1,14 @@
+id: '1108'
+title: vscode-clicking-an-architect-g
+protocol: air
+phase: implement
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-06-28T03:22:54.684Z'
+updated_at: '2026-06-28T03:22:54.684Z'
--- a/codev/projects/1108-vscode-clicking-an-architect-g/status.yaml
+++ b/codev/projects/1108-vscode-clicking-an-architect-g/status.yaml
@@ -1,7 +1,7 @@
 id: '1108'
 title: vscode-clicking-an-architect-g
 protocol: air
-phase: implement
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-06-28T03:22:54.684Z'
-updated_at: '2026-06-28T03:22:54.684Z'
+updated_at: '2026-06-28T03:29:52.382Z'
--- a/codev/projects/1108-vscode-clicking-an-architect-g/status.yaml
+++ b/codev/projects/1108-vscode-clicking-an-architect-g/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-06-28T03:33:56.506Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-06-28T03:22:54.684Z'
-updated_at: '2026-06-28T03:29:52.382Z'
+updated_at: '2026-06-28T03:33:56.507Z'
+pr_ready_for_human: true
--- a/codev/state/air-1108_thread.md
+++ b/codev/state/air-1108_thread.md
@@ -1,0 +1,29 @@
+# air-1108 — Architect group-header click opens architect terminal
+
+Issue #1108 (AIR, strict mode). Parity: in the Agents view's architect-axis
+grouping mode, clicking an architect group header should open that architect's
+terminal (like builder rows do), while the chevron keeps toggling expand/collapse.
+
+## Implementation
+
+- `packages/vscode/src/views/builders.ts` — `rootChildren()`: when the active
+  grouping axis is `architect` (`grouping.id === 'architect'`), set
+  `groupItem.command = { command: 'codev.openArchitectTerminal', title: ..., arguments: [g.key] }`
+  on each group header. `g.key` is the architect name (null `spawnedByArchitect`
+  folds into `main` per `architectGrouping()`). Stage/area headers stay
+  command-less containers (they name no launchable entity).
+- `codev.openArchitectTerminal` already accepts the optional name arg (#786
+  Phase 6) and warns gracefully on a stale owner — no command-handler change.
+
+## Tests
+
+- New `packages/vscode/src/__tests__/builders-architect-header-command.test.ts`:
+  architect headers carry the command with the right arg (incl. `main` fold for
+  null owner); stage and area headers carry no command.
+
+## Status
+
+- `pnpm test:unit` (vscode): 43 files / 516 tests pass (after building
+  types/core/artifact-canvas deps in the fresh worktree).
+- `pnpm check-types` main tsc pass clean.
+- Net diff well under 300 LOC — AIR is the right protocol.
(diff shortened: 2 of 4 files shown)


amrmelsayed:
## Architect Review

Low-risk AIR PR. Verified the implementation independently against PR head.

### Implementation

`packages/vscode/src/views/builders.ts:295-318` — clean. The `isArchitectAxis` gate is computed once outside the `groups.map(g => ...)` loop, and the `command` is set on each group header only when the architect axis is active. Stage and area headers stay command-less. The inline comment names every scope note from #1108: chevron routes independently for expand/collapse, `g.key` is the architect name (null owners folded into `main` per `architectGrouping`), and stale-owner handling rides on `codev.openArchitectTerminal`'s existing graceful-warn behaviour.

```ts
const isArchitectAxis = grouping.id === 'architect';
return groups.map(g => {
  const groupItem = new BuilderGroupTreeItem(...);
  if (isArchitectAxis) {
    groupItem.command = {
      command: 'codev.openArchitectTerminal',
      title: 'Open Architect Terminal',
      arguments: [g.key],
    };
  }
  ...
});
```

### Test coverage

`builders-architect-header-command.test.ts` pins all four scope notes:

- **Architect headers carry the command** — independently verified for `main` and a named sibling (`vscode`).
- **`main` fold for null `spawnedByArchitect`** — a builder with null owner produces a `main` group header that opens main's terminal.
- **Stage headers have no command** — pure grouping containers.
- **Area headers have no command** — same, with two-area fixture to bypass the lone-Uncategorized flatten so real area headers actually render.

Mocks `vscode` per the established `__tests__` pattern (matches `builders-autoreveal.test.ts`).

### Risk and scope

17 net lines in `builders.ts` (7 of code + 6 of comment, plus the `isArchitectAxis` const above the loop), 169 lines of test. Exactly the "5-line wiring + small unit test" scope #1108 anticipated. No new dependencies, no API surface change, no migration. Build / unit / e2e all green per the builder's report.

### Recommendation

**APPROVE.** Ready for the pr gate.

---
Architect review

PR #1111, opened 2026-06-28
[Bugfix #1110] Fix flaky artifact-canvas D2 watch-failure test (missing waitFor)

### LINKED ISSUE #1110: bugfix: artifact-canvas test 'surfaces a synchronous FileAdapter.watch() failure via onError without throwing (D2)' flakes under CI load (missing waitFor on DOM assertion)
## Problem

The test `surfaces a synchronous FileAdapter.watch() failure via onError without throwing (D2)` in `packages/artifact-canvas/src/components/__tests__/artifact-canvas.test.tsx` flakes under CI load. It passes 5/5 isolated and 8/8 full-file locally; fails intermittently in CI. Surfaced by the AIR builder for #1108 (PR #1109), which observed the failure on a VS Code-only diff with no artifact-canvas changes (`git diff main...HEAD -- packages/artifact-canvas` empty on the builder branch).

## Root cause (verified against source)

The test waits for `onError` but synchronously asserts the DOM in the next line:

```ts
// packages/artifact-canvas/src/components/__tests__/artifact-canvas.test.tsx
await waitFor(() => expect(onError).toHaveBeenCalled());
expect(document.querySelector('p[data-line]')).not.toBeNull(); // read succeeded → content still renders
```

The `waitFor` ensures `onError` fired, but the very next assertion is a synchronous DOM read with no second `waitFor`. The async `read()` render in `ArtifactCanvas` can resolve AFTER `onError` fires — under CI load the render is delayed enough that the DOM query lands before the paragraph is in the tree. Race; test fails.

The sibling test immediately below (`Disposable.dispose is safe to call more than once (idempotent contract, D2)`) uses the correct pattern, which is the same fix to apply here:

```ts
await waitFor(() => expect(document.querySelector('p[data-line]')).not.toBeNull());
```

## Fix

Wrap the second assertion in `waitFor`:

```ts
await waitFor(() => expect(onError).toHaveBeenCalled());
await waitFor(() => expect(document.querySelector('p[data-line]')).not.toBeNull());
```

One-line change. The semantic stays identical (the assertion eventually succeeds), only the timing tolerance widens to match the sibling test's pattern.

## Verification

- Test must pass consistently when the existing test file is run under load (e.g., `vitest --runInBand` alongside other dashboard suites that compete for the event loop).
- Existing passing cases in the file must continue passing (no functional change beyond timing).
- Full `pnpm --filter @cluesmith/codev-dashboard test` suite stays green.

## Out of scope

- Auditing the broader artifact-canvas test file for other similar `await waitFor` → bare-assertion patterns. Could ride along if the fixer notices any (the file is small), but not required.
- Any production code changes to `ArtifactCanvas` itself. The test is wrong, not the code.

## Protocol

BUGFIX. Trivial, isolated, mechanical fix with a clear root cause; no design decisions, no architectural impact.

amrmelsayed:
## Summary

Fixes a CI-load flake in the artifact-canvas test `surfaces a synchronous FileAdapter.watch() failure via onError without throwing (D2)`.

## Root Cause

The test awaited `onError` firing, then asserted the DOM **synchronously** on the next line:

```ts
await waitFor(() => expect(onError).toHaveBeenCalled());
expect(document.querySelector('p[data-line]')).not.toBeNull(); // bare read — race
```

`ArtifactCanvas`'s async `read()` render can resolve *after* `onError` fires. Under CI load the render is delayed enough that the synchronous DOM query runs before the paragraph is in the tree, so the assertion intermittently fails. The sibling test directly below (`Disposable.dispose is safe to call more than once`) already uses the correct `waitFor`-wrapped pattern.

## Fix

Wrap the second assertion in `waitFor` to match the sibling test's pattern:

```ts
await waitFor(() => expect(onError).toHaveBeenCalled());
await waitFor(() => expect(document.querySelector('p[data-line]')).not.toBeNull());
```

Semantics are unchanged (the assertion still must eventually hold); only the timing tolerance widens. One-line change, no production code touched.

## Test Plan

- `pnpm --filter @cluesmith/codev-artifact-canvas test` → 56/56 pass.
- Stress-ran the D2 test in isolation 5× → 5/5 pass.
- Full workspace `pnpm build` + `pnpm test` green (verified via porch checks: build ✓, tests ✓).

## Regression Coverage

This is a test-timing fix to a race, so the corrected D2 test **is** the regression coverage: with the bare synchronous read it flakes under load; with the `waitFor` wrapper it is deterministic. An additional standalone test would only re-assert the same behavior.

Fixes #1110

### DIFF
--- a/codev/projects/bugfix-1110-bugfix-artifact-canvas-test-su/status.yaml
+++ b/codev/projects/bugfix-1110-bugfix-artifact-canvas-test-su/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1110
 title: bugfix-artifact-canvas-test-su
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-06-28T05:52:25.996Z'
-updated_at: '2026-06-28T05:52:25.997Z'
+updated_at: '2026-06-28T05:52:58.629Z'
--- a/codev/projects/bugfix-1110-bugfix-artifact-canvas-test-su/status.yaml
+++ b/codev/projects/bugfix-1110-bugfix-artifact-canvas-test-su/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-06-28T06:00:05.286Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-06-28T05:52:25.996Z'
-updated_at: '2026-06-28T05:55:57.904Z'
+updated_at: '2026-06-28T06:00:05.287Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-1110-bugfix-artifact-canvas-test-su/status.yaml
+++ b/codev/projects/bugfix-1110-bugfix-artifact-canvas-test-su/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1110
 title: bugfix-artifact-canvas-test-su
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-06-28T05:52:25.996Z'
-updated_at: '2026-06-28T05:52:58.629Z'
+updated_at: '2026-06-28T05:55:57.904Z'
--- a/codev/projects/bugfix-1110-bugfix-artifact-canvas-test-su/status.yaml
+++ b/codev/projects/bugfix-1110-bugfix-artifact-canvas-test-su/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1110
+title: bugfix-artifact-canvas-test-su
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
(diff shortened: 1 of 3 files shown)


amrmelsayed:
## CMAP review (3-way)

| Model | Verdict |
|---|---|
| Gemini (agy) | APPROVE |
| Codex | APPROVE |
| Claude | APPROVE |

All three approved the one-line fix. Claude's review flagged the **identical** bare-assertion-after-`waitFor` race in the sibling test `surfaces adapter errors via onError and does not throw; failed list keeps prior markers (D2)` (line 133). The issue explicitly invited fixing such patterns as a ride-along ('the file is small'), so I applied the same one-line `waitFor` wrap there too (commit 846b53dc). 56/56 tests pass.

amrmelsayed:
## Architect Review

Low-risk BUGFIX. Two-line diff in `packages/artifact-canvas/src/components/__tests__/artifact-canvas.test.tsx`, exactly as scoped:

- Line 161 — the D2 watch-failure test originally reported. `expect(document.querySelector(...)).not.toBeNull()` is now wrapped in `waitFor(() => ...)`.
- Line 130 — the sibling D2 adapter-error test with the identical race. Same wrap. This is the ride-along the issue's Out-of-scope section explicitly invited ("could ride along if the fixer notices any") — picked up via the CMAP claude lane.

Three-way CMAP unanimous APPROVE (gemini + codex + claude). 56/56 tests pass. No production code changes; semantics preserved (the assertion still eventually succeeds, just with retry tolerance matching the sibling test's pattern).

**APPROVE.** Ready for the pr gate.

---
Architect review

Issue #777, opened 2026-05-19
consult --type {spec,plan,impl} resolves artifact from invoking workspace (not the builder's branch); impl-review diff base hardcoded to main


amrmelsayed:
## Problem

Two related artifact-resolution defects in the consult harness produce verdicts that don't reflect the artifact the architect actually intends to review.

### Defect A — workspace-rooted artifact resolution

`consult --type {spec,plan,impl} --issue N` resolves the artifact (spec/plan/PR diff) from the invoking process's **workspace root** rather than from the builder's branch. When run from an architect's checkout (typically the repo's integration branch), the consult reviews the version of the artifact that's on the integration branch, **not** the builder's reworked version on `builder/...`. Verdicts are therefore about the stale artifact, not the work-in-progress one the architect intends to review.

`commands/consult/index.ts`: `findPlanContent(workspaceRoot, issueId)` / `findSpecContent` / impl-diff resolution all read from `workspaceRoot` — wherever the architect's invocation is rooted. There is no flag/parameter to point the consult at a specific branch/ref/worktree.

### Defect B — `main` hardcoded as default integration branch (multi-layer, uneven distribution)

A **multi-layer** defect: the same assumption is fossilised in *both* the porch CLI implementation **and** the protocol instruction markdown. Fixing one without the other leaves the symptom intact. **Distribution across protocols is uneven** — PIR is roughly 4× more exposed at Layer 2 than its siblings.

#### Layer 1 — porch CLI / consult code (2 sites, protocol-agnostic)

- `packages/codev/src/commands/consult/index.ts:989` — `git merge-base HEAD main` (the `buildImplQuery` fallback).
- `packages/codev/src/commands/consult/index.ts:1297` — `git merge-base main origin/${pr.headRefName}` (the impl-review path).

This layer hits every protocol equally — any `--type impl` consult on a non-`main`-default repo produces false-positive "scope creep" verdicts: the diff sweeps in every commit that's on the integration branch but not on `main`, attributing them to the branch under review.

#### Layer 2 — protocol .md instruction files (per-protocol breakdown)

| Protocol | Hard-fail (literal shell commands) | Wrong-answer (review questions) | Convention | Layer-2 exposure |
|---|---|---|---|---|
| **PIR** | 5: `builder-prompt.md:85`, `prompts/implement.md:21,110`, `prompts/review.md:41,250` | 1: `consult-types/pr-review.md:25` | 2: `implement.md:136`, `review.md:242` | **High** |
[...]

PR #1114, opened 2026-06-29
[Bugfix #1113] Anchor consult --type integration diff on an integration-branch base

### LINKED ISSUE #1113: consult --type integration: no way to anchor the diff on the integration branch (gh pr diff inherits the PR's trunk base → 10k-line diffs overflow the reviewer)
## Summary

`consult --type integration pr <N>` can produce a huge diff (10k+ lines) that overflows a reviewer model's context, in repos that maintain a **long-lived integration branch ahead of the default branch** (e.g. an `ci`/`integration` branch well ahead of `main`). The integration review has **no way to anchor the diff on the integration branch** — it always inherits whatever base the PR targets on the host.

## Verified root cause (the reported "diffs against main" is one level off)

Traced through `packages/codev/src/commands/consult/index.ts` (source + installed dist agree):

- `--type integration` → `buildPRQuery(prId)` → `fetchPRDiff(prId)` → the `pr-diff` forge concept → **`gh pr diff <N>`** (GitHub) / `glab mr diff` (GitLab). That's the *only* diff source for this path — no local `git diff`, no `resolveDefaultBranch`, **no hardcoded `main`**.
- `gh pr diff <N>` is anchored on the **PR's base branch as recorded on the host** (merge-base of base↔head). So the diff balloons when the PR's base is the trunk (`main`) while the head was branched from an integration branch (`ci`) that is well ahead of trunk: "changes vs main" then legitimately swallow the entire `ci`-over-`main` delta.

So the literal claim "consult diffs against main instead of ci" is a symptom: consult diffs against the **PR's base**, which in this workflow is `main`, when the meaningful comparison is against the integration branch the work actually forked from.

The buildPRQuery path already writes the diff to a temp file and points the model at the path (not inlined), so this is **not** an inlining problem — the diff is genuinely too large because it's anchored on the wrong branch.

## Precedent (why integration is the odd one out)

The `impl` review type was already hardened for exactly this diff-base class:
- **#777** (CLOSED) — "impl-review diff base hardcoded to main"
- **#784** (CLOSED) — two-dot over-include

`impl` now computes the diff **locally** as `git diff origin/<baseRefName>...origin/<headRefName>` (three-dot, merge-base anchored on the PR's *actual* base), with both refs verified up front. The **`integration` path never received this treatment** — it still leans on `gh pr diff` with no base control.

## Prescribed fix (BUGFIX scope — design choice is specified, no plan gate needed)

Give the `integration` review path a diff-base override, reusing the verified machinery the `impl` path already has:

1. Add an explicit base override for `--type integration` — a `--base <ref>` flag (and/or read an integration-branch from `.codev/config.json`, e.g. `consult.integrationBranch`).
2. When a base override is provided, compute the diff **locally** as `git diff origin/<base>...origin/<head>` (three-dot, merge-base anchored), mirroring the `impl` path — verify both refs up front and fail loudly with an actionable `git fetch` hint if either is unresolvable (do **not** silently fall back to reviewing the checked-out tree — that's the cmap-3 degradation guarded against in the impl path).
3. **Default behavior unchanged**: with no override, keep `gh pr diff <N>` (the PR's host base) for back-compat.
4. Reuse the existing escaped-ref / three-dot helpers from the impl path rather than duplicating; keep the on-disk diff-file handoff (don't inline).

This lets a repo with a `ci`-ahead-of-`main` topology point integration reviews at `ci` and get a small, correct diff, while everyone else is unaffected.

**Out of scope** (note as follow-ups, do not implement here): auto-detecting the integration branch without configuration; changing the default base; GitLab parity beyond passing the override through the `pr-diff` concept if trivial.

## Repro

In a repo with an integration branch (`ci`) well ahead of `main`, with a PR whose head was branched from `ci` but which targets `main`:
```
consult --type integration pr <N>
```
→ the diff balloons to the full `ci`-over-`main` delta and overflows the reviewer model.

## Acceptance

- `consult --type integration pr <N> --base <ci-ref>` (or the configured integration branch) computes `origin/<ci-ref>...origin/<head>` three-dot and produces only the PR's actual change.
- Unresolvable base/head refs fail loudly with a `git fetch` hint — no silent degradation to the local checkout.
- No `--base` / no config → behavior identical to today (`gh pr diff`).
- A test covers the override path (small diff) and the no-override default (unchanged).

Reported via an adopter project running a `ci`-ahead-of-`main` integration workflow.

waleedkadous:
## Summary

`consult --type integration pr <N>` could produce a 10k+ line diff that overflows a reviewer model, in repos with a **long-lived integration branch ahead of the default branch** (e.g. a `ci` branch well ahead of `main`). The integration review had **no way to anchor the diff on the integration branch** — it always inherited the PR's host-recorded base.

This adds a base override for the integration path, reusing the verified machinery the `--type impl` path already has (#777/#784), so a `ci`-ahead-of-`main` repo can point integration reviews at `ci` and get a small, correct diff. Everyone else is unaffected.

## Root Cause

`--type integration` (both builder and architect context) → `buildPRQuery(prId)` → `fetchPRDiff` → the `pr-diff` forge concept → **`gh pr diff <N>`**. `gh pr diff` is anchored on the PR's base branch **as recorded on the host** (merge-base of base↔head). When a PR's base is the trunk (`main`) while the head was branched from an integration branch (`ci`) well ahead of trunk, "changes vs main" legitimately swallow the entire `ci`-over-`main` delta. The `impl` path was already hardened for exactly this diff-base class; the `integration` path never received that treatment.

## Fix

Give the integration path a diff-base override, mirroring the hardened `impl` path:

1. **`--base <ref>` flag + `consult.integrationBranch` config** (`.codev/config.json`). Precedence: flag → config → unset.
2. **`computeLocalPRDiff()`** — when a base override is present, fetch both refs, verify both `origin/<ref>` resolve, then compute the diff locally as **`git diff origin/<base>...origin/<head>`** (three-dot, merge-base anchored). Refs are passed as single argv elements (`execFileSync`) so branch names with shell metacharacters can't break out.
3. **Fails loudly** with an actionable `git fetch` hint if either ref is unresolvable — it never silently degrades to reviewing the checked-out tree (the cmap-3 degradation guarded against in the impl path).
4. **`buildPRQuery(prId, localDiff?)`** takes an optional local-diff override; PR info/comments are still fetched normally, and the on-disk diff-file handoff is preserved (no inlining).
5. **Default behavior unchanged**: with no flag/config, the integration review still uses `gh pr diff` (the PR's host base).
6. **Fail-fast**: `--base` is rejected outside `--type integration`. Config-load errors propagate (no silent revert to `gh pr diff`).

The local-diff path computes the diff with plain `git` (provider-agnostic), so it works for GitHub and GitLab alike without touching the `pr-diff` concept.

### Out of scope (follow-ups, per the issue)
Auto-detecting the integration branch without configuration; changing the default base; deeper GitLab-specific work.

## Test Plan

Regression suites `bugfix-1113-integration-base.test.ts` (helpers) and `bugfix-1113-integration-dispatch.test.ts` (wiring):

- **Override path (small diff)**: a git fixture with `main` → `ci` (well ahead) → `feature` (off `ci`); `computeLocalPRDiff(ci, feature)` three-dot yields **only** `feature.txt`, excluding the `ci`-over-`main` delta.
- **Bug demonstration**: anchoring on `main` sweeps in the whole `ci` delta (`ci-big.txt`) — the overflow source.
- **Command-path wiring**: drives the real `--type integration` dispatch (`resolveArchitectQuery` → `resolveIntegrationBase` → `computeLocalPRDiff` → `buildPRQuery`) with `--base ci` (forge `pr-view`/`pr-diff`/comments stubbed); asserts the query consumed the local three-dot diff (changed-file set = `feature.txt` only; the temp diff file contains the real change, **not** the stubbed `gh pr diff` output).
- **Fail-loud**: unresolvable base ref and unresolvable head ref both throw with a `git fetch` hint (no silent local-tree fallback).
- **`resolveIntegrationBase` precedence**: flag > config > undefined (no override → default `gh pr diff`, unchanged); malformed config propagates the error; `--base` works even with a broken config.
- **Guard**: `--base` rejected outside `--type integration`.

### Proven locally vs. deferred ⚠️

- **Proven here**: the diff math, the dispatch wiring (that `--type integration --base <ref>` computes and consumes the local three-dot diff rather than `gh pr diff`), the fail-loud ref checks, precedence/guard, and config error propagation. Full unit suite green (0 failures); `pnpm build` green.
- **Deferred (NOT run locally)**: the true end-to-end headline — that a *genuinely overflowing* PR now **fits** a reviewer model and returns a verdict with `--base ci`. codev itself has no `ci`-ahead-of-`main` topology, so this requires an adopter repo with the real topology. Recommend closing this against the real repro PR after merge.

Fixes #1113

### DIFF
--- a/codev-skeleton/resources/commands/consult.md
+++ b/codev-skeleton/resources/commands/consult.md
@@ -66,12 +66,26 @@ consult -m codex --protocol spir --type phase
 
 # Integration review
 consult -m gemini --type integration
+
+# Integration review anchored on a long-lived integration branch (#1113)
+consult -m codex --type integration --issue 42 --base ci
 ```
 
 **Options:**
 - `--protocol <name>` — Protocol: spir, aspir, air, bugfix, maintain
 - `-t, --type <type>` — Review type: spec, plan, impl, pr, phase, integration
 - `--issue <number>` — Issue number (required from architect context)
+- `--base <ref>` — **`--type integration` only.** Anchor the diff on this base branch (e.g. `ci`), computed locally as `git diff origin/<base>...origin/<head>` (three-dot, merge-base anchored). Use in repos with a long-lived integration branch ahead of the default branch so the review sees only the PR's actual change, not the whole integration-over-trunk delta. Unresolvable refs fail loudly with a `git fetch` hint (no silent fallback to the local checkout). Defaults to config `consult.integrationBranch`; with neither set, the integration review uses the PR's host-recorded base (`gh pr diff`), unchanged.
+
+**Config (`.codev/config.json`):**
+```jsonc
+{
+  "consult": {
+    // Repo-wide default base for `--type integration` (overridden by --base).
+    "integrationBranch": "ci"
+  }
+}
+```
 
 **Context resolution:**
 - **Builder context** (cwd inside `.builders/`): auto-detects project ID, spec, plan, and PR from porch state
--- a/codev/projects/bugfix-1113-consult-type-integration-no-wa/status.yaml
+++ b/codev/projects/bugfix-1113-consult-type-integration-no-wa/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-06-29T00:44:37.929Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-06-29T00:07:25.656Z'
-updated_at: '2026-06-29T00:27:45.096Z'
+updated_at: '2026-06-29T00:44:37.929Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-1113-consult-type-integration-no-wa/status.yaml
+++ b/codev/projects/bugfix-1113-consult-type-integration-no-wa/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1113
 title: consult-type-integration-no-wa
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-06-29T00:07:25.656Z'
-updated_at: '2026-06-29T00:07:25.657Z'
+updated_at: '2026-06-29T00:26:39.591Z'
--- a/codev/projects/bugfix-1113-consult-type-integration-no-wa/status.yaml
+++ b/codev/projects/bugfix-1113-consult-type-integration-no-wa/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1113
+title: consult-type-integration-no-wa
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-06-29T00:07:25.656Z'
+updated_at: '2026-06-29T00:07:25.657Z'
--- a/codev/projects/bugfix-1113-consult-type-integration-no-wa/status.yaml
+++ b/codev/projects/bugfix-1113-consult-type-integration-no-wa/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1113
 title: consult-type-integration-no-wa
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-06-29T00:07:25.656Z'
-updated_at: '2026-06-29T00:26:39.591Z'
+updated_at: '2026-06-29T00:27:45.096Z'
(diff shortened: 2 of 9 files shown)


waleedkadous:
## CMAP review (Gemini · Codex · Claude)

| Reviewer | Verdict | Confidence |
|---|---|---|
| Claude | ✅ APPROVE | HIGH |
| Gemini | ✅ APPROVE | HIGH |
| Codex  | ✅ APPROVE | HIGH |

**Codex** initially returned `REQUEST_CHANGES` twice; both findings were valid and have been addressed:

1. **Config errors were silently swallowed** — `resolveIntegrationBase()` wrapped `loadConfig()` in a blanket `try/catch`, so a malformed `.codev/config.json` (or legacy `af-config.json`) would quietly revert to `gh pr diff` — the exact overflow this fix prevents. Removed the catch so those errors propagate like every other `loadConfig` caller; the explicit `--base` flag still short-circuits before the read. Added regression tests (`43507a4b`).
2. **Skeleton doc not mirrored** — the `consult.md` update was applied to `codev/resources/commands/` but not the shipped `codev-skeleton/` template (two-tree mirroring rule). Mirrored it (`542b9048`). `packages/codev/skeleton/` is a gitignored build artifact, regenerated by `copy-skeleton`.

After both fixes, Codex re-reviewed → **APPROVE (HIGH), no issues**.

Verification: full unit suite green (0 failures), `pnpm build` green, porch Fix-phase checks (`npm run build`, `npm test`) green.

PR #1125, opened 2026-06-30
fix: Replace SSE eviction with rejection to prevent thundering herd

amustafa:
## Summary

- Replace evict-on-cap with reject-on-cap (503 + `Retry-After: 5`) to break the SSE reconnection cascade
- Raise `SSE_MAX_CLIENTS` from 50 to 200
- Add per-client max-age jitter (4–6 min range) to prevent synchronized eviction bursts after Tower restarts
- Send `retry: 5000` SSE directive to space out browser reconnections
- Remove per-connection INFO-level connect/disconnect logging (heartbeat every 30s is sufficient)

## Problem

When SSE client count exceeded the cap of 50, the oldest client was evicted (`sseClients.shift()` + `res.end()`). The evicted client auto-reconnected, pushing the count back to the cap, evicting another — creating a thundering herd loop at up to 800 conn/sec. Each short-lived TCP connection left a `TIME_WAIT` socket for ~60s, accumulating 2,900+ sockets and consuming ~10% of the ephemeral port range, causing intermittent `EADDRNOTAVAIL` for all localhost services.

Fixes #1124

## Test plan

- [ ] Connect 200 SSE clients → 201st receives `503` with `Retry-After: 5` (no eviction of existing clients)
- [ ] Verify `TIME_WAIT` sockets stay under 100 during normal operation: `ss -tn state time-wait dst 127.0.0.1:4100 | wc -l`
- [ ] Verify heartbeat logs show stable client count (no churn): `tail -f ~/.agent-farm/tower.log | grep 'SSE heartbeat'`
- [ ] Dashboard, VSCode, and terminal SSE connections work normally
- [ ] Start 10 clients simultaneously → they expire over a 2-minute window (jitter), not all at once

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1126, opened 2026-07-01
Fix SSE thundering herd causing ephemeral port exhaustion

### LINKED ISSUE #1124: Tower SSE thundering herd causes ephemeral port exhaustion on localhost
## Problem

The Tower server SSE endpoint (`GET /api/events`, port 4100) has a hard client cap of 50 (`SSE_MAX_CLIENTS` in `tower-server.ts`). When exceeded, the server evicts the oldest client to make room. The evicted client auto-reconnects, which pushes the count back to the cap, evicting another client — creating a thundering herd loop at up to ~800 connections/second.

Each short-lived TCP connection leaves a socket in `TIME_WAIT` for ~60 seconds. At peak churn this accumulates 2,900+ `TIME_WAIT` sockets against `127.0.0.1:4100`, consuming ~10% of the ephemeral port range (32768–60999). This causes intermittent `EADDRNOTAVAIL` / connection-refused errors for **all** localhost services — not just Tower.

## Root Cause Chain

1. SSE client count exceeds cap (50)
2. Server evicts oldest client (`sseClients.shift()` + `res.end()`)
3. Evicted client reconnects (browser EventSource ~3s, VSCode via `backoffDelayMs`)
4. Reconnect pushes count back to cap → another eviction → cascade
5. Each connection leaves `TIME_WAIT` socket (~60s TTL)
6. Ephemeral port pool (28,232 ports) depleted under sustained load
7. Other localhost services (gRPC, Postgres, Kafka, Redis) fail intermittently

## Evidence

| Metric | Value |
|---|---|
| Peak connect rate | 49,524 connects/min |
| TIME_WAIT sockets at diagnosis | 2,888 against 127.0.0.1:4100 |
| Port utilization at peak | ~10% sustained, higher in bursts |
| Log file size (2.5 days) | 1.18M lines, 774K SSE connect/disconnect |

## Code Locations

- **Cap + eviction logic**: `packages/codev/src/agent-farm/servers/tower-server.ts` — `addSseClient`
- **SSE handler**: `packages/codev/src/agent-farm/servers/tower-routes.ts` — `handleSSEEvents`
- **Age-based eviction (heartbeat)**: `packages/codev/src/agent-farm/servers/tower-server.ts` — heartbeat interval
- **SSEClient type**: `packages/codev/src/agent-farm/servers/tower-types.ts`

### 1. Reject instead of evict (critical)
Return 503 + `Retry-After` when at capacity instead of evicting oldest. Rejection is a dead end; eviction is a chain reaction.

### 2. Raise cap from 50 to 200
SSE connections are lightweight. 50 is too low for workstations with multiple builders + dashboard tabs + VSCode.

### 3. Send `retry:` SSE directive
`retry: 5000\n\n` spaces out browser reconnections.

### 4. Throttle SSE connect/disconnect logging
774K log lines in 2.5 days. Heartbeat every 30s is sufficient visibility.

### 5. Stagger max-age eviction with per-client jitter
Fixed 5-min max-age causes synchronized eviction bursts. Add ±1 min jitter (4–6 min range).

amustafa:
## Summary

- Replace evict-on-cap with reject-on-cap (503 + `Retry-After: 5`) to break the reconnect cascade
- Raise `SSE_MAX_CLIENTS` from 50 to 200
- Send `retry: 5000` SSE directive to space out browser reconnections
- Add per-client jitter (±1 min) to max-age eviction to prevent synchronized eviction bursts
- Remove per-client connect/disconnect INFO logging (774K lines in 2.5 days)

## Root Cause

The SSE endpoint evicts the oldest client when at cap (50). The evicted client auto-reconnects, pushing count back to cap, evicting another — a chain reaction at up to ~800 connections/second. Each short-lived TCP connection leaves a TIME_WAIT socket for ~60s, accumulating 2,900+ sockets and consuming ~10% of the ephemeral port range.

## Fix

**Reject instead of evict**: Return 503 + Retry-After when at capacity. Rejection is a dead end; eviction was a chain reaction.

**Raise cap**: 50 → 200. SSE connections are lightweight; 50 was too low for workstations with multiple builders + dashboard tabs + VSCode.

**Retry directive**: `retry: 5000` tells browsers to wait 5s before reconnecting instead of the default ~3s.

**Jitter**: Per-client `maxAgeJitterMs` in [-60s, +60s] prevents synchronized max-age eviction bursts.

**Logging**: Removed per-client connect/disconnect logs. Heartbeat still runs for dead-client detection.

## Test Plan

- [x] 3 regression tests added to `tower-routes.test.ts`
  - 503 rejection when at capacity (verifies status code, Retry-After header, body)
  - `retry: 5000` directive present in SSE stream
  - No cleanup listeners registered after rejection
- [x] All 87 tower-routes tests pass
- [x] Build passes
- [x] Net diff: 76 insertions, 25 deletions (well under 300 LOC guideline)

Fixes #1124

### DIFF
--- a/codev/projects/bugfix-1124-tower-sse-thundering-herd-caus/status.yaml
+++ b/codev/projects/bugfix-1124-tower-sse-thundering-herd-caus/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1124
+title: tower-sse-thundering-herd-caus
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-07-01T01:22:39.407Z'
+updated_at: '2026-07-01T01:22:39.408Z'
--- a/codev/projects/bugfix-1124-tower-sse-thundering-herd-caus/status.yaml
+++ b/codev/projects/bugfix-1124-tower-sse-thundering-herd-caus/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1124
 title: tower-sse-thundering-herd-caus
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-01T01:22:39.407Z'
-updated_at: '2026-07-01T01:22:39.408Z'
+updated_at: '2026-07-01T01:24:38.461Z'
--- a/codev/projects/bugfix-1124-tower-sse-thundering-herd-caus/status.yaml
+++ b/codev/projects/bugfix-1124-tower-sse-thundering-herd-caus/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1124
 title: tower-sse-thundering-herd-caus
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-01T01:22:39.407Z'
-updated_at: '2026-07-01T01:24:38.461Z'
+updated_at: '2026-07-01T01:35:15.992Z'
--- a/codev/state/bugfix-1124_thread.md
+++ b/codev/state/bugfix-1124_thread.md
@@ -28,3 +28,9 @@ Three regression tests added to tower-routes.test.ts:
 - No cleanup listeners registered after rejection
 
 All 87 tower-routes tests pass. Build passes.
+
+## CMAP review — Gemini finding addressed
+
+Gemini identified that browser `EventSource` transitions to `CLOSED` (readyState === 2) on a non-200 response (our 503) and does NOT auto-reconnect. The dashboard's `useSSE.ts` hook assumed auto-reconnect on all errors. Fixed by adding manual reconnection with 2-5s jitter when `readyState === EventSource.CLOSED`. VSCode extension and legacy `tower.html` already use fetch-based SSE with their own retry logic — no change needed there.
+
+Added regression test in `useSSE.reconnect.test.ts`. All 6 dashboard SSE tests pass.
--- a/codev/state/bugfix-1124_thread.md
+++ b/codev/state/bugfix-1124_thread.md
@@ -1,0 +1,30 @@
+# bugfix-1124 thread — SSE thundering herd / ephemeral port exhaustion
+
+## Investigate phase started
+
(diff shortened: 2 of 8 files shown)


amustafa:

**Risk**: Medium (Tower subsystem, but focused bugfix — 155 net lines, 8 files)

**Verdict**: APPROVE

Clean, well-scoped fix that addresses all five vectors of the thundering herd root cause. Key integration points verified:

- `addSseClient` type change (`void` → `boolean`) — single consumer (`handleSSEEvents`), test mock updated
- `maxAgeJitterMs` optional with `?? 0` fallback — backward compatible  
- Client-side `EventSource.CLOSED` handling correctly scoped to `useSSE.ts` — VSCode extension and `tower.html` use fetch-based SSE, unaffected
- Module-level `reconnectTimer` follows existing singleton pattern

CMAP results: gemini=APPROVE, claude=APPROVE (codex=AUTH_FAILURE, not a code concern).

No integration concerns. Please merge.

---
*Architect integration review*

waleedkadous:
This is fantastic — thank you for running it through the BUGFIX protocol. 🙏 It turned an already-good fix into a clearly better one, and it's worth spelling out exactly why this supersedes #1125, because the protocol earned its keep here:

**1. It caught a real correctness bug that #1125 would have shipped.** #1125's whole strategy is "reject at capacity with a 503" — but its dashboard `useSSE.ts` still had `onerror` commented "EventSource automatically reconnects on error; no action needed." That's false for a non-200 response: a browser `EventSource` goes to `CLOSED` (readyState 2) on a 503 and does **not** auto-reconnect. So under #1125, any dashboard tab that got rejected at the cap would go dark **permanently** until a manual reload — the fix would silently break the very clients it's meant to protect. CMAP (Gemini) flagged exactly this, and this PR fixes it with a manual, jittered reconnect on `CLOSED`. That's the difference between a fix that's correct on paper and one that's correct end-to-end.

**2. It's tested.** #1125 had a manual checklist; this has actual regression tests — the server-side 503/`retry:` behavior *and* a `useSSE.reconnect` test asserting the dashboard reconnects out of `CLOSED`. Those guard the fix (and specifically the bug above) against regressing.

**3. It's auditable.** The porch/CMAP trail (status.yaml, thread, the documented Gemini finding) means the reasoning is reviewable, not just the result — including the nice scoping note that the VSCode extension and legacy `tower.html` use fetch-based SSE with their own retry, so no change was needed there.

**4. Cleaner, too.** Storing per-client `maxAgeJitterMs` on the client (vs. #1125's inline computed max-age) reads better and documents intent.

Same sharp diagnosis and core fix as before — now correct, tested, and auditable. Merging this one. Thank you again, genuinely, for taking the extra pass. 🙏

waleedkadous (approved):
Superior to #1125 in every way — correctness (fixes the 503/CLOSED no-reconnect bug #1125 shipped), tests, and auditability. Full rationale in the PR comment. Thank you!

Issue #2, opened 2025-09-29
Context Window Optimization - AI Needs Project Knowledge Without Breaking Limits


waleedkadous:
Filed on behalf of Amr Elsayed.

## Problem
As projects grow, providing adequate context to AI without breaking context window limits becomes impossible. AI needs to understand APIs, database schemas, components, and existing patterns but can't hold all this information.

## From Amr
> "The trick is in how to provide proper context without breaking the context window during the planning phase"

## Requirements
AI should know at minimum:
- What does each API do?
- What is the purpose of each table/column in the database?
- What is each screen/page/component responsible for?
- How to use existing IP when building new features

## Proposed Solution
- Implement incremental context loading strategies
- Create summarized project knowledge base
- Develop selective context inclusion based on feature requirements

Issue #1, opened 2025-09-29
The Gestalt Problem - Features Developed in Isolation Without System Awareness


waleedkadous:
Filed on behalf of Amr Elsayed.

## Problem
Features are designed and developed in isolation without understanding the broader system context, leading to code duplication, inefficiency, and technical debt. Each feature operates as if it exists in a vacuum, unaware of existing services, components, and patterns.

## Impact
- Massive code duplication across features
- Unnecessary network requests  
- Recreation of existing functionality
- Growing technical debt with each new feature
- 500+ DAPPER commits showing consistent duplication patterns

## Evidence from Amr
> "After using DAPPER to develop many features, the issue that has been showing up consistently is that each feature is designed, developed, and tested in isolation (with limited system wide context)"

> "I have used DAPPER to push around 500 commits, and I have seen this pattern many times where I would end up in a situation where I would tell the agent, we already have x service or y component or z state that can be used"

## Proposed Solution
- Implement a system-wide context layer that documents existing services, APIs, components
- Create pre-planning research phase that maps existing functionality  
- Develop feature dependency analysis before implementation

Issue #1112, opened 2026-06-28
Unify builder conversation resume onto persisted session IDs (harness.session), retiring the jsonl-discovery heuristic


amrmelsayed:
## Context

#832 introduced a persisted, agent-neutral conversation-session mechanism for **architects**: a `session_id` column on the `architect` table, written at spawn (Tower generates a UUID and passes it via the new `HarnessProvider.session.newSessionArgs`), and read back at every revive surface to `--resume`. It was needed because named sibling architects (Spec 755) **share** `cwd = workspacePath`, which breaks the newest-jsonl-by-mtime heuristic.

**Builders** still resume via that jsonl-discovery heuristic (`findLatestSessionId` in `spawn.ts`, shipped in #831/#829). It works for builders because each owns a **unique worktree cwd** (`.builders/<id>/`), so "newest jsonl" is unambiguous. So this is **not a bug** — builders resume correctly today.

## Why unify anyway

Now that the robust mechanism exists, builders are the lone holdout on a weaker one:

- **Single mechanism over two.** One persisted-id path for all agent terminals beats maintaining jsonl-discovery for builders + stored-id for architects (consolidate duplicates / single source of truth).
- **More robust.** A stored id is immune to the heuristic's edge cases — a stray manual `claude` session run in the worktree, mtime races, or `--resume` picking an unintended lineage.
- **Reuse.** The `HarnessProvider.session` capability (`newSessionArgs`/`resumeArgs`) added in #832 drops straight into the builder spawn path; Codex/Gemini/OpenCode builders degrade gracefully (no `session` capability -> fresh spawn), same as architects.

## Proposed approach

1. Add a persisted `session_id` for builders (migration on the `builders` table, or reuse `terminal_sessions` — **open question**, see below).
2. At builder spawn (`spawn.ts`), generate a UUID, pass it via `harness.session?.newSessionArgs`, and store it.
3. On resume (`afx spawn --resume`, `afx workspace recover`), read the stored id and `--resume` via `harness.session?.resumeArgs`.
4. **Fallback chain:** stored id -> jsonl-discovery (kept as a legacy fallback) -> fresh. Unlike architects, builders need **no transitional capture command** — jsonl-discovery already resolves legacy builders correctly (unique cwd), so it's a clean, self-healing fallback with zero migration ceremony.

## Open questions

[...]

Issue #1094, opened 2026-06-23
afx send: detectCurrentBuilderId silently falls back to bare worktree name on state.db read failure → builder messages misroute to main


waleedkadous:
## Summary

`detectCurrentBuilderId()` (`packages/codev/src/agent-farm/commands/send.ts`) **silently** falls back to the bare worktree directory name when it cannot read the workspace `state.db`. That bare name (e.g. `bugfix-2461`) does not match any `builders.id` (canonical form `builder-bugfix-2461`), so Tower's affinity resolver (`lookupBuilderSpawningArchitect` → `undefined`) drops to the "non-builder sender → main first" branch. Net effect: **a builder's `afx send architect` is silently delivered to `main` instead of its spawning architect.**

This violates "fail fast, never implement fallbacks": a fatal environmental fault is laundered into a subtle, plausible-looking misroute that is very hard to diagnose.

## How it surfaced (real incident)

In a multi-architect workspace, builders spawned by `triage`/`cost`/`dreaming` had their `afx send architect` messages land on `main`. Tower log showed the bare sender:

```
Message sent: bugfix-2461 → architect (terminal efe2c590…)   # efe2c590 = main
Message sent: bugfix-2325 → architect (terminal efe2c590…)   # main
# vs. a healthy send: builder-experiment-2489 → architect …  (canonical id)
```

Trigger was a **Node ABI mismatch**: a `node` on PATH (ABI 127) different from the one codev's `better-sqlite3` was built for (ABI 147) made `new Database()` throw. The `catch { return worktreeDirName }` swallowed it. But the defect is general — **any** `state.db` open failure (corruption, permissions, ABI mismatch, transient lock) produces the same silent misroute.

## Root cause (the fallback in question)

```ts
let wsDb: Database.Database;
try {
  wsDb = new Database(dbPath, { readonly: true });
} catch {
  return worktreeDirName;   // <-- swallows fatal error, ships a non-canonical id
}
// ... and the no-match path also: return worktreeDirName;
```

We are unambiguously inside `.builders/<id>/` here (the regex matched). So "can't read the DB" is an **error condition**, not a "this isn't a builder" condition — but the function treats both identically.

## Proposed fix

[...]

PR #1127, opened 2026-07-01
Consolidate state.db tables into global.db (single user-global database)

### LINKED ISSUE #1118: Consolidate state.db tables into global.db (single user-global database) to eliminate the cwd-dependent fragmentation that causes "missing architect state after restart"
## Problem

`.agent-farm/state.db` is named and located as if it's workspace-local, but it actually holds rows from **every workspace Tower has ever interacted with** while parked in that directory. Combined with Tower's start-cwd determining which `state.db` file is the "active" one for a session, this produces a fragmentation pattern where:

- Tower running from workspace A → reads/writes `A/.agent-farm/state.db`
- During that session, any cross-workspace interaction (`afx send` from workspace B, dashboard switching to workspace C, VS Code extension calling Tower from workspace D) lands its rows in A's state.db.
- Reboot. Tower next starts from workspace B → reads/writes `B/.agent-farm/state.db`. Rows from previous session in A's file are now **invisible to the running Tower**, even though they're intact on disk.

User-facing symptom (verified against this user's machine): "**after a computer restart, some architects in my workspaces are missing their state / session data**." The architects whose rows happened to land in Tower's previous start-cwd's state.db are stranded; the architects whose rows are in the *current* Tower's state.db work fine. Hence "some" architects missing, not "all."

### Audit confirming the pattern

8 `state.db` files exist across a multi-workspace user's local checkouts:

| File | architect rows | distinct workspaces represented |
|---|---|---|
| `~/.agent-farm/state.db` | 1 | 1 (a builder worktree) |
| `~/repos/cluesmith/codev/.agent-farm/state.db` | 40 | **38** (codev + multiple sibling projects + 11 builder worktrees + 17 ephemeral test workspaces) |
| `~/repos/cluesmith/shannon/.agent-farm/state.db` | 4 | 3 (codev + shannon + a codev builder worktree) |
| `~/repos/insighttrail/kidscanspell/.agent-farm/state.db` | 4 | 4 (multiple unrelated workspaces) |
| `~/repos/bb/MPPS2/.agent-farm/state.db` | 2 | 2 (codev + shannon builder worktrees) |
| `~/repos/amrmelsayed/codev/.agent-farm/state.db` | 1 | 0 (pre-Migration-v11 row, no workspace_path) |
| `~/repos/insighttrail/autotoggl/.agent-farm/state.db` | 0 | 0 (empty file, lazy-init from a CLI command run there) |
| `~/repos/cluesmith/codev/worktrees/changelog/.agent-farm/state.db` | 0 | 0 (empty file, lazy-init from a CLI command run there) |

Workspaces appearing in *multiple* state.db files (the fragmentation symptom):
- `cluesmith/codev` → codev's DB + shannon's DB
- `cluesmith/shannon` → codev's DB + shannon's DB
- `cluesmith/codev-streamdeck` → codev's DB + kidscanspell's DB
- Multiple builder worktrees scattered across 3-4 different files

### Population affected

Verified against the spawn paths:

| User class | Affected? |
|---|---|
| Single-workspace user with manual `afx tower start` from project root | No |
| Single-workspace user with VS Code extension auto-start | No |
| Multi-workspace user with manual `afx tower start` from "wherever they happen to be" | **Yes** |
| Multi-workspace user with VS Code extension auto-start | **Yes** (the extension passes `cwd: workspacePath` to spawn, so whichever Codev project is opened first after a Tower-restart determines the location) |
| Local-install dev user | **Yes** (every `pnpm -w run local-install` runs `afx tower start` from the script's invocation cwd) |

This is a default-behaviour issue for any user with more than one Codev project on disk and any Tower-restart history (machine reboot, `pnpm -w run local-install`, manual `afx tower stop/start`, VS Code switching projects). It is not a developer-environment-only issue.

### Verified spawn-path code locations

- `packages/codev/src/agent-farm/commands/tower.ts` — `afx tower start` spawns with `cwd: process.cwd()`. Inherits the shell's cwd.
- `packages/codev/src/agent-farm/utils/config.ts` — `findWorkspaceRoot()` walks up from cwd looking for `.git` + `codev/`; falls back to cwd if not found (which is how `~/.agent-farm/state.db` got created when Tower was started from `$HOME` once).
- `packages/vscode/src/tower-starter.ts:47-51` — VS Code extension's `autoStartTower` spawns Tower with `cwd: workspacePath ?? undefined`. The VS Code workspace folder becomes Tower's cwd.
- `packages/codev/src/agent-farm/db/index.ts` — `getDb()` lazy-initialises state.db at `<workspaceRoot>/.agent-farm/state.db`, creating an empty file if absent. This is why workspaces where Tower never ran still have state.db files (CLI commands run from inside them lazy-created the file).

### Why two databases at all?

`state.db` and `global.db` exist as separate files for archaeological reasons, not principled design. `state.db` predates the system-wide Tower architecture: back when Tower was per-workspace, each workspace had its own daemon and its own genuinely-workspace-local file. When Tower's architecture shifted to a system-wide singleton, the right move would have been to relocate state.db to a user-global location to match the new scope. Instead, Bugfix #826 added a `workspace_path` column (Migration v11) so rows from different workspaces could coexist in the same file — a symptom-patching fix that left state.db's *location* workspace-local while making its *scope* effectively global.

`global.db` was added separately around Spec 0090 / TICK-001 for inherently-cross-workspace tables (terminal_sessions, port_allocations, file_tabs, cron_tasks, known_workspaces). The split has no principled boundary today — it's "tables added before the singleton transition" vs "tables added after." Both files are functionally user-global; only state.db lies about its scope via its file location.

## Proposed approach: collapse state.db tables into global.db (single user-global database)

All state.db tables (`architect`, `builders`, `utils`, `annotations`) move into `~/.agent-farm/global.db`. The per-workspace `state.db` file is retired. Tower keeps one connection to one file regardless of which workspace it's serving; rows are disambiguated by the existing `workspace_path` column (already in place since Migration v11, Bugfix #826).

### What changes

1. **All state.db tables migrate into global.db** in a new migration. global.db's migration system absorbs the schemas; a new global.db migration version becomes "absorb state.db tables (architect, builders, utils, annotations)".
2. **`getDb()` and `getGlobalDb()` consolidate** — `getDb()` always returns the global.db connection. The cwd-dependent state.db creation logic at `db/index.ts` is removed.
3. **`findWorkspaceRoot()` is no longer load-bearing for db location** — it stays for protocol/template resolution (the four-tier file resolver, unchanged), but the db path is now always `~/.agent-farm/global.db`.
4. **state.ts function signatures stay** — they already take `workspace_path` as their first argument (post-Bugfix-#826), which is now the row-disambiguator within the single shared DB rather than the path-resolver.
5. **Spawn-time writes go to the right workspace's rows** regardless of where Tower was started — no more cwd dependence at all.
6. **terminal_sessions ↔ architect/builder is a natural single-file join** going forward. Today it's a cross-file lookup; after the merge it's a single `SELECT JOIN`. The #1115 PID-parity follow-up benefits directly.

### What doesn't change

- `architect`, `builders`, `utils`, `annotations` table schemas (`workspace_path` column stays as the row-disambiguator; same role it has today, just inside the single shared DB).
- Migration v12's `session_id` column work — independent.
- `~/.agent-farm/global.db` location.
- The `known_workspaces` table in global.db (still the registry of workspaces Tower has ever touched).
- The per-workspace `.agent-farm/` directory — left in place (forward-compat for future per-workspace files; don't actively delete during migration).

### Migration strategy

**One-time on-disk migration** during the Tower restart that follows the upgrade:

1. **Discover all existing state.db files** by reading `known_workspaces` and checking each workspace's `.agent-farm/state.db`. Also include `~/.agent-farm/state.db` (the home-fallback that gets created when Tower was started from `$HOME`).
2. **Per file**: union its `architect`, `builders`, `utils`, `annotations` rows into global.db using `INSERT OR REPLACE` keyed on the primary key. Conflict resolution: latest-`started_at` wins. The `workspace_path` column preserves identity of rows from different workspaces.
3. **Dry-run available**: `afx tower start --dry-run-migration` lists every row that would be merged + flags conflicts (same row written to multiple state.db files with different content). `--apply-migration` (or running plain `afx tower start` after the upgrade) commits.
4. **Source files preserved**: do not delete state.db files during migration. Rename them to `state.db.pre-merge-<timestamp>` so users can recover if needed.
5. **Idempotent**: a marker row in global.db's `_migrations` table tracks completion. Subsequent Tower starts are no-ops on the migration check.

### Stale-row hygiene

The "self-cleaning on workspace deletion" benefit of per-workspace state.db files is replaced by an explicit prune command:

`afx prune-state` — removes rows from `architect`, `builders`, `utils`, `annotations` whose `workspace_path` is not in `known_workspaces`. Dry-run by default; `--apply` to commit. Handles the stale-row accumulation case (the 17 test-workspace rows in this user's audit are exactly what it removes). Run on user demand, not automatically — the per-workspace "free cleanup via rm" pattern was always opt-in in practice.

`afx workspace forget <path>` — removes a workspace from `known_workspaces` AND prunes its associated state rows in one command. The clean way to retire a workspace whose directory has been deleted or moved.

## Acceptance criteria

- `architect`, `builders`, `utils`, `annotations` tables exist in global.db (added via a new migration).
- `getDb()` returns the global.db connection from all callsites in state.ts.
- A spawn flow in workspace A writes its architect row to `~/.agent-farm/global.db` regardless of where Tower was started.
- Reboot scenario: stop Tower from workspace A, start Tower from workspace B, verify workspace A's architects (now in global.db) are readable from Tower running at B.
- Migration: existing state.db files are scanned, rows merged via INSERT OR REPLACE with latest-started_at-wins conflict resolution; source files renamed (not deleted).
- Migration is idempotent: re-running doesn't re-migrate.
- The cwd-dependent state.db creation logic is removed; `findWorkspaceRoot()` no longer drives db location.
- `afx prune-state` removes rows whose `workspace_path` is not in `known_workspaces`; dry-run by default, `--apply` to commit.
- `afx workspace forget <path>` removes a workspace from known_workspaces and prunes its rows in one command.
- Unit tests cover: cross-workspace row isolation via `workspace_path`; migration row-routing correctness on multi-source merge with conflicts; migration idempotency; prune-state correctness against known_workspaces; `workspace forget` end-to-end.

## Alternatives considered

1. **Per-workspace state.db** (each `<workspace>/.agent-farm/state.db` holds only its own rows; Tower opens N connections lazily, LRU-bounded). Gains: free cleanup on workspace deletion (rm the dir, state goes with it); portability (move the dir, state moves with it); self-cleaning of stale rows. Costs: N database handles requiring an LRU pool; per-callsite audit to thread workspace_path through every state.ts function; cross-workspace queries (dashboard's all-workspaces view, reconciliation passes) become iterations over `known_workspaces` opening each state.db on demand; the terminal_sessions ↔ architect/builder join stays cross-file; adds an architectural decision for future tables ("which workspace's DB?"). The cleanup and portability gains assume workflows (frequent `rm` of workspaces, moving workspace dirs) that aren't dominant; the self-cleaning gain is achievable cheaper via the `prune-state` command above. **Rejected** as more complex than the problem warrants.

2. **Just-move state.db to `~/.agent-farm/state.db`** (keep the two-DB split, just relocate the path). Fixes the cwd-dependence; doesn't address the deeper architectural debt that the state.db / global.db split was always arbitrary. New table additions would still face the "which DB?" question. **Rejected** as a half-measure.

3. **Force-migration on Tower startup with no opt-out / no dry-run**. The migration is straightforward and well-defined, but a hard-blocking migration on first launch under new code feels coarse. The proposed approach makes migration automatic with a dry-run preview available and sources preserved as `*.pre-merge-<timestamp>` files. Same net effect, gentler upgrade story. **Adopted** in the proposal above.

4. **Hybrid: per-workspace state.db AND user-global state.db with sync**. Two sources of truth; sync correctness becomes its own problem. **Rejected**.

## Out of scope

- Moving `terminal_sessions`, `port_allocations`, `cron_tasks`, `file_tabs` out of global.db. They're already correctly user-global; no change needed.
- Removing the `workspace_path` column from `architect` / `builders` / etc. It's still the row-disambiguator within the single shared DB; could be revisited as a v2 cleanup once the merge has lived for a while, but not for v1.
- Auto-pruning stale rows on every Tower start. `afx prune-state` is opt-in.
- Touching Migration v12's `session_id` work or any other in-flight migration.
- Per-workspace `.agent-farm/` directory cleanup. The directory stays for forward compat; let it die naturally if unused.

## Protocol

PIR. This is a significant infrastructure change with multiple design decisions (migration dry-run UX, prune-state semantics, conflict-resolution policy on merge, the `workspace forget` command shape) that benefit from plan-gate validation, plus dev-gate verification of:
- the multi-workspace reboot scenario on a running Tower
- the dry-run migration preview against a real machine with fragmented state.db files
- the prune-state command against synthetic known_workspaces/stale-row fixtures

before opening the PR.

## Related

- **Bugfix #826** — Migration v11 added `workspace_path` to the `architect` table primary key when cross-workspace contamination was first discovered. That was the symptom-patching response; this issue is the structural fix that retires the now-redundant per-workspace file location.
- **Spec 0090 / TICK-001** — added `global.db` for inherently-cross-workspace tables; this issue extends the same principle to the remaining state.db tables, which are also effectively user-global since Bugfix #826.
- **PR #1116 / Issue #832** — per-architect session UUID. Independent; layers cleanly on top of the single-DB consolidation.
- **#1115** — `afx status` builders PID parity. Benefits directly from the consolidation: the terminal_sessions ↔ builders join becomes a single-file query.
- **#1112** — Unify builder conversation resume on persisted session IDs. Independent but adjacent; both touch the same per-builder state surface.
- **Architect-row residue** — the 17 test-workspace rows + 11 builder-worktree rows in this user's codev state.db (audited in the conversation that spawned this issue) are what `afx prune-state` is designed to remove cleanly.

amrmelsayed:
# PIR Review: Consolidate state.db tables into global.db

Fixes #1118

## Summary

Retired the per-workspace `state.db` file and moved its four tables (`architect`, `builders`,
`utils`, `annotations`) into the already user-global `~/.agent-farm/global.db`. `getDb()` now
returns the single global connection, so architect/builder state no longer depends on Tower's
start-cwd — the root cause of "some architects missing their state after a restart." `builders`
was reshaped to be workspace-scoped (composite `(workspace_path, id)` PK), and a one-time,
marker-gated boot migration (plus a manual `afx db consolidate <path>` and an
`afx tower start --dry-run-migration` preview) moves legacy `state.db` files in — renaming
sources to `*.pre-merge-<ts>`, never deleting.

## Files Changed

(code, vs `main` merge-base)

- `packages/codev/src/agent-farm/db/schema.ts` (+92/−…) — `GLOBAL_SCHEMA` absorbs the four tables; `builders` reshaped with `workspace_path` + composite PK; `LOCAL_SCHEMA` retained as legacy-shape reference
- `packages/codev/src/agent-farm/db/index.ts` (−≈470 net) — `getDb()`/`getDbPath()` alias the global connection; removed `ensureLocalDatabase()` + the local v1–v12 migration ladder; added global migration v14
- `packages/codev/src/agent-farm/db/consolidate.ts` (+456, new) — reusable upsert-if-newer engine (`planMigration`/`applyMigration`), defensive legacy reads, `_consolidation` marker, strict boot one-off (`runBootConsolidation`)
- `packages/codev/src/agent-farm/db/migrate.ts` (−135, deleted) — dead state.json→SQLite migration
- `packages/codev/src/agent-farm/db/types.ts` (+1) — `DbBuilder.workspace_path`
- `packages/codev/src/agent-farm/state.ts` (+…/−…) — builder functions workspace-scoped (derive `workspace_path` from `worktree` on upsert; optional `workspacePath` on reads); `lookupBuilderSpawningArchitect` drops the per-file open
- `packages/codev/src/agent-farm/utils/workspace-path.ts` (+26, new) — single-source `normalizeWorkspacePath` (dedupe of 3 copies)
- `packages/codev/src/agent-farm/servers/tower-server.ts` (+16) — boot one-off caller
- `packages/codev/src/agent-farm/servers/tower-utils.ts` (±16) — re-exports the shared normalizer
- `packages/codev/src/agent-farm/servers/overview.ts` (+14/−…), `servers/tower-routes.ts` (+4/−…), `lib/builder-lookup.ts` (+12/−…), `commands/cleanup.ts` (+5/−…) — builder-read callsites scoped by workspace
- `packages/codev/src/agent-farm/commands/db.ts` (+48), `commands/tower.ts` (+54), `cli.ts` (+17) — `afx db consolidate <path>` + `afx tower start --dry-run-migration`
- Tests: `__tests__/consolidate.test.ts` (+246, new); `state.test.ts`, `spec-755-lookup-builder.test.ts`, `overview.test.ts` reworked for the single-DB model; `migrate.test.ts` (−264, deleted)
- Docs: `codev/resources/arch-critical.md`, `arch.md`, `lessons-learned.md` updated

## Commits

- `c05a2e0d` … `1f2e8fa9` — Plan phase (draft + gate-decision revisions)
- `539dc93b` / `de357092` Move state.db tables into global.db; reshape builders workspace-scoped
- `d44fc707` Consolidation engine + boot one-off + `afx db consolidate` + dry-run preview
- `3a2ee987` Update tests for single-DB model (state, lookup-builder, overview)
- `d32c0391` Consolidate engine tests + fix workspace derivation (lastIndexOf)
- `fc6f371a` Clean up stale state.db references in state.ts comments
- `f8823e53` Replace introduced ternaries with if/else (no-ternary preference)
- `65ed4788` Dedupe workspace-path canonicalization into one leaf module

## Test Results

- `pnpm build`: ✓ pass
- `pnpm test`: ✓ pass — full suite **3419 passed, 48 skipped, 0 failures** (agent-farm alone: 2014 passed); consolidate suite adds 8 new cases
- Manual verification (dev-approval gate): ran the new code against a **real snapshot** of this machine's fragmented data — `afx db consolidate` dry-run + apply on a copy of the real 40-row `codev` `state.db` migrated **40 architect rows across 38 workspaces** into a throwaway `global.db`, source renamed to `*.pre-merge-<ts>`, re-apply a graceful no-op. Verified `afx tower start --dry-run-migration` previews without spawning/mutating, and the v13→v14 migration path on a snapshot of the real `global.db` (existing `terminal_sessions`/`known_workspaces` preserved).

## Architecture Updates

Routed **HOT** (`arch-critical.md`): rewrote the always-injected state fact — "State lives in
`.agent-farm/state.db` + `~/.agent-farm/global.db`" → "single user-global `~/.agent-farm/global.db`
(Issue #1118 retired per-workspace state.db; architect/builders keyed by `workspace_path`)". This
is behavior-changing and cross-cutting, so it belongs in the capped hot tier (no displacement
needed — reworded an existing entry in place).

Routed **COLD** (`arch.md`): updated the "State inconsistency" quick-map row, the "State
Consistency" invariant, and the state-persistence section (two-databases → one `global.db`;
described the composite-PK reshape, the `consolidate.ts` engine, and migration v14). Left the
historical #826 v11-migration detail intact (still accurate about schema evolution).

## Lessons Learned Updates

Routed **COLD** (`lessons-learned.md` → Architecture): a `[From #1118]` entry — "when a
subsystem's scope outgrows its storage location, relocate the storage rather than patching the
scope with a column," plus the corollary that moving a table into a shared DB forces re-examining
its primary key (the `builders` id-collision finding), and the reusable migration techniques
(upsert-if-newer, defensive `PRAGMA` reads). Not routed HOT: the existing hot lesson "single
source of truth beats distributed state — consolidate duplicates rather than syncing them"
already captures the headline; this PR is a detailed instance of it, so it belongs in cold.

- **The `builders` reshape (`db/schema.ts`, migration v14, `state.ts`)** — the one structural
  change. Confirm every builder *read* that should be workspace-scoped is (`builder-lookup.ts`,
  `cleanup.ts`, `tower-routes.ts`, `overview.ts`, `loadState`), and that the security-relevant
  `lookupBuilderSpawningArchitect` filters by `(workspace_path, id)`.
- **`deriveWorkspaceFromWorktree` uses `lastIndexOf('/.builders/')`, not `indexOf`** — a bug the
  tests caught (a worktree path can itself sit under another `.builders/`, e.g. a builder testing
  from its own worktree). Same helper duplicated in `state.ts` and `consolidate.ts`.
- **Strict one-off marker semantics** (`runBootConsolidation`) — marks done unconditionally on
  first boot even if the active `state.db` is absent/empty; satellites are recovered via
  `afx db consolidate`. Confirm the marker + row-copy share one transaction.
- **Defensive legacy reads** (`consolidate.ts` normalizers) — pre-v11 architect (integer id, no
  `workspace_path`) and legacy builder rows are normalized without running the old migration
  ladder; enum values are clamped to satisfy the new CHECK constraints.
- **Schema-version tolerance is load-bearing, not incidental.** Because this PR deletes the
  local migration ladder (v1–v12), consolidation is the *sole* reader of legacy `state.db` files
  and never brings their schema up to date first — so it must tolerate **every** historical
  shape. This is why `readRows` uses `SELECT *` (never `SELECT <column>`) and every
  migration-added column is read as `row.<col> ?? null` (or synthesized). Concretely, a very
  common field shape is **v11-but-not-v12**: the `architect` table has `workspace_path` (#826)
  but no `session_id` column (#832 not yet rolled out). An explicit `SELECT session_id` would
  throw `no such column`; `SELECT *` + `?? null` yields `session_id = null` and migrates cleanly.
  Same for the builder columns added by v8/v9/v10 (`issue_number`, `spawned_by_architect`,
  `type` incl. `'pir'`). Pinned by tests: the `pre-v11` case and an explicit **v11-but-not-v12
  (no `session_id`)** regression test that would fail if any read ever became column-specific.
- **`db/index.ts` deletion is large** (−≈470) — it's the removed local-migration ladder, folded
  into `GLOBAL_SCHEMA` + v14. Worth a scan that nothing live still referenced it.

## How to Test Locally

- **View diff**: VSCode sidebar → right-click builder `pir-1118` → **View Diff**
- **What to verify** (maps to the plan's Test Plan — full recipe in `codev/plans/1118-*.md`):
  - Unit: `pnpm --filter @cluesmith/codev test` (state / consolidate / lookup-builder suites)
  - Safe preview: build the branch, then `afx tower start --dry-run-migration` (or
    `afx db consolidate <copy-of-state.db>`) — reads, no writes
  - Isolated apply: `NODE_ENV=test AF_TEST_DB=global-probe.db` against a copy of a real
    `state.db` / `global.db` — exercises migration v14 + data migrate without touching live state
  - Live: install from the **main** checkout (not the worktree) → Tower restart auto-migrates;
    then stop-from-A / start-from-B and confirm A's architects are visible (the fragmentation fix)

## Consultation (3-way verify, single advisory pass)

- **claude**: APPROVE — "implementation faithfully follows the approved plan across all dimensions."
- **codex**: REQUEST_CHANGES — caught **two real issues, both now fixed + regression-tested** (commit `d9828577`):
  1. **`clearRuntime()` wiped all workspaces' builders** — it did an unscoped `DELETE FROM builders`, and `afx workspace stop` calls it; on the shared `global.db` that deleted *every* workspace's builders, not just the stopping one. Fixed: `clearRuntime(workspacePath)` scopes the delete by `workspace_path` (threaded through `stop.ts`); `utils`/`annotations` (global, vestigial) are left untouched to avoid a cross-workspace wipe. Test: `clearRuntime(A)` leaves B's builders intact.
  2. **`afx db consolidate` repeat-run wasn't idempotent** — re-running on the renamed source `fatal`ed, and an already-`*.pre-merge-*` archive would re-migrate + double-rename. Fixed: missing source → friendly no-op; archived file → skip. Tests added.
- gemini/agy: not configured for the porch verify (a 2-way consult); a later ad-hoc 3-way
  re-run (iteration 2) added gemini (APPROVE).

**Iteration 2 (re-run after the iter-1 fixes) + a manual cross-workspace audit** surfaced two
more consolidation-fallout bugs, both fixed (commit `7f6ce330`):
  3. **`send.ts detectCurrentBuilderId` opened the RETIRED per-workspace `state.db`** to resolve
     a builder's canonical id for `afx send` (issue #1094 anti-spoofing). After migration renames
     that file, `afx send` from a worktree would break. This was the **last** direct `state.db`
     open (its siblings `lookupBuilderSpawningArchitect` and `overview.ts` were fixed in the main
     implementation; this one was missed). Found by the audit — **all three consult models missed
     it**. Fixed: read `global.db` scoped by `workspace_path`. Test rewritten incl. a same-id
     cross-workspace scoping case.
  4. **`afx db consolidate` dry-run wasn't side-effect-free** (codex iter2) — it called
     `getGlobalDb()`, which eagerly creates/migrates `global.db`. Fixed: dry-run opens `global.db`
     read-only (or in-memory if absent); only `--apply` uses the RW connection. Test added.
- **Known low-severity item**: `loadState()` returns `utils`/`annotations` unscoped (they have no
  `workspace_path` column). These tables are vestigial (no production producers — verified), so
  the exposure is a stale/empty read, not a live leak. Left as-is; noted for a future cleanup.
**Iteration 3** (final re-run: claude APPROVE, gemini COMMENT, codex REQUEST_CHANGES) — two
items, both addressed (commit `c81b8f0d`):
  5. **`send.ts` `.builders` regexes used lazy `.+?` (first-match)** — a *nested* worktree
     (`<repo>/.builders/a/.builders/b`) would resolve the outer builder. Same class as the
     `indexOf`→`lastIndexOf` fix; switched to greedy `.+` (last `/.builders/`) + regression test.
     Nesting is an unsupported anti-pattern (`afx spawn` from inside a worktree), so this is a
     consistency fix, not a normal-path bug. (Stale `detectCurrentBuilderId` docstring refreshed.)
  6. **No direct `runBootConsolidation` coverage** — added tests for the real boot path:
     first-boot migrate+marker+rename, marker-set no-op, and strict mark-done-when-absent.
- **Env note**: the codex consult initially failed because macOS 26 XProtect flagged the
  un-notarized `@openai/codex` vendor binary as malware and auto-deleted it (SIGKILL→ENOENT);
  restoring the binary + ad-hoc `codesign` unblocked it. Upstream/packaging follow-up, not a
  signal about this PR.

## Notes / Out of Scope

- **Pre-existing suite state**: scaffold tests (adopt/update/cold-tier/hot-tier/consult) and the
  terminal shellper-integration tests fail on a *stale build* (they need `pnpm build`'s
  copy-skeleton step / a live shellper) — verified they fail identically with this change stashed,
  so they are unrelated. After a full `pnpm build` the entire suite is green.
- **`session_id` for builders** deliberately excluded — that's #1112's charter (builders resume
  via unique-worktree mtime discovery; no shared-cwd sibling to disambiguate). Layers on cleanly
  as a later additive column.
- **`afx prune-state` / `afx workspace forget`** (named in the issue's acceptance criteria) were
  **cut** at the plan gate — stale rows are harmless under workspace-scoped reads; cleanup is a
  separable follow-up if ever needed.
- **`normalizeWorkspacePath` dedupe** collapsed 3 real copies into `utils/workspace-path.ts`; the
  bare `fs.realpathSync` calls in `tower-instances.ts`/`tower-routes.ts` are a different pattern
  (no fallback) and were intentionally left alone.


amrmelsayed:

Verified both codex-flagged fixes and the underlying consolidation independently against PR head. High-quality PR for a high-impact structural change.

### Codex fix #1: `clearRuntime` workspace-scoping — verified

`packages/codev/src/agent-farm/state.ts:422-441` — signature and semantics both changed correctly:

```ts
// before
export function clearRuntime(): void {
  db.prepare('DELETE FROM builders').run();     // unscoped — wipes ALL workspaces
  db.prepare('DELETE FROM utils').run();
  db.prepare('DELETE FROM annotations').run();
}

// after
export function clearRuntime(workspacePath: string): void {
  const db = getDb();
  const ws = canonicalize(workspacePath);
  db.prepare('DELETE FROM builders WHERE workspace_path = ?').run(ws);
}
```

Genuine regression that codex correctly caught: on the shared `global.db`, `afx workspace stop` on workspace A would previously have wiped workspace B's builders too. The fix scopes the delete by `workspace_path` (composite PK on the `builders` table makes this safe); intentionally leaves `utils`/`annotations` alone since they're global UUID-keyed and vestigial (no producers) — the alternative would have been a cross-workspace wipe of those tables too. `clearState()` (full-wipe) preserved for uninstall / nuke flows.

Call site at `commands/stop.ts:46` correctly threaded through: `clearRuntime(workspacePath)`.

**Regression test verified** at `state.test.ts` — creates two workspaces' builders, calls `clearRuntime('/workspace/aaa')`, asserts workspace B's builder survives. Pins the load-bearing behavior.

### Codex fix #2: `dbConsolidate` idempotency — verified

`packages/codev/src/agent-farm/commands/db.ts:178-190` — two guards added:

- **Missing source**: `if (!existsSync(sourcePath))` was `fatal(...)`, now `logger.info(... already migrated?)` and returns cleanly. Repeat-runs after a successful `--apply` renamed the source no longer error.
- **Already-archived file**: `if (/\.pre-merge-/.test(sourcePath))` skips with a message. Prevents double-migration of an already-consolidated archive (which would have harmlessly re-migrated all rows then double-renamed).

**Regression tests verified** at `db.test.ts:184-211` — two cases pinning both guard branches. Neither touches the DB (both return before opening it).

### Overall consolidation quality

Large but surgical: 29 files, +2046/-930 net. The structural shift is exactly what #1118 proposed:
- `db/index.ts` shrinks by ~600 lines (state.db creation + cwd-dependent path resolution removed)
- `db/consolidate.ts` (456 lines new) — migration engine with dry-run preview + apply
- `db/schema.ts` picks up the migrated architect/builders/utils/annotations tables into global.db's schema
- `state.ts` refactored (~192 lines net) — every helper now writes to the single shared DB, disambiguated by workspace_path
- `commands/db.ts` (new) — `afx db consolidate` and `afx db stats` CLI surface
- `commands/tower.ts` picks up +54 lines — boot-time one-off migration path
- `utils/workspace-path.ts` (new, 26 lines) — canonicalize helper deduplicated into a single leaf module (per your no-scattered-canonicalization preference visible in earlier PIRs)

Test suite: 2017 passed per builder report; the 246-line new `consolidate.test.ts` covers the migration engine end-to-end.

Docs updated: `codev/plans/1118-*.md` (336 lines), `codev/reviews/1118-*.md` (127 lines), plus `arch-critical.md`, `arch.md`, `lessons-learned.md` — the architecture story of "state.db is gone; global.db is the single source of truth" is now propagated through the hot/cold docs.

### Codex consult XProtect note

The builder's note about codex being blocked by macOS XProtect flagging its un-notarized binary is a separate packaging concern; doesn't affect this PR's mergability since codex DID complete for the review-time consult that caught the two real issues. Worth tracking as a packaging follow-up but not blocking.

### Recommendation

**APPROVE for the pr gate.**

The structural fix is what #1118 asked for, both codex-flagged issues are handled surgically with regression coverage, the migration story preserves rollback via archived `*.pre-merge-*` files, and the workspace-scoping fix closes a real cross-workspace data-loss risk that would otherwise have shipped with the consolidation.

---
Architect integration review

amrmelsayed:
## Architect Re-Review (post-audit + iter2)

Verified both additional fixes from commit `7f6ce330` against source.

### Audit-found (HIGH): `send.ts detectCurrentBuilderId` — verified

`packages/codev/src/agent-farm/commands/send.ts:113-160`. The pre-fix code opened the per-workspace `state.db` directly (`new Database(join(workspacePath, '.agent-farm', 'state.db'), { readonly: true })`), bypassing the singleton `getDb()`. Post-migration that file is renamed to `*.pre-merge-*`, so `afx send` from any builder worktree would throw `BuilderIdResolutionError` on the missing file — silently breaking Codev's own inter-agent messaging.

The fix repoints to `getGlobalDbPath()` (readonly), scopes the query by `workspace_path`:

```ts
// before
const rows = wsDb.prepare('SELECT id, worktree FROM builders WHERE worktree IS NOT NULL').all()

// after
const ws = normalizeWorkspacePath(workspacePath);
const rows = gdb.prepare('SELECT id, worktree FROM builders WHERE workspace_path = ? AND worktree IS NOT NULL').all(ws)
```

Uses the same `normalizeWorkspacePath` helper deduped into `utils/workspace-path.ts` earlier in the PR — consistent surface across every callsite that touches the workspace-path key. Anti-spoofing behavior preserved (`BuilderIdResolutionError` still fires with the clear "refusing to send with an unverified identity" reason, updated to name global.db).

**Test rewrite verified** at `spec-755-lookup-builder.test.ts` (199 lines updated in the diff): includes the same-id cross-workspace scoping case — a builder with id `X` in workspace A cannot be matched by a query from workspace B.

Notable: this was the **last** direct `state.db` open in the codebase, and it was **missed by all three consult models** in the earlier CMAP pass. Caught by the builder's own post-implementation cross-workspace audit. Strong quality signal about verification thoroughness — the audit surfaced what CMAP didn't.

### Codex iter2 finding (MED): `dbConsolidate` dry-run side-effect-free — verified

`packages/codev/src/agent-farm/commands/db.ts:193-215`. Pre-fix, the dry-run path called `getGlobalDb()`, which eagerly creates the target DB file and runs migration v14 (the state.db → global.db schema addition). This violated the "no writes" contract of `--dry-run` — the *preview* was itself performing the setup half of the migration.

The fix distinguishes the two paths:

```ts
if (options.apply) {
  db = getGlobalDb();                              // RW connection, runs migrations
} else if (existsSync(globalDbPath)) {
  db = new Database(globalDbPath, { readonly: true });   // RO — cannot write
  closeAfter = true;
} else {
  // In-memory fallback: no file to touch; every source row reads as "new".
  // Correct preview shape.
}
```

Read-only connection closed via `closeAfter` flag when dry-run finishes. In-memory fallback for the "no global.db yet" case correctly previews every row as new. Both paths preserve the preview UX.

**Test coverage added**: new `db-consolidate-command.test` (per the commit message) covers dry-run side-effect-freedom + the idempotency guards from iter1.

### Overall status of #1127 post-audit + iter2

- Original consolidation: verified in prior review comment
- Codex iter1 findings (clearRuntime scoping, consolidate idempotency): verified previously
- Codex iter2 finding (dry-run read-only): **verified now**
- Cross-workspace audit finding (send.ts direct state.db open): **verified now**
- Total test count: 2018 passed per builder report (+1 from iter1's 2017 — matches the new db-consolidate-command test)
- Iter2 CMAP verdicts: claude APPROVE, gemini APPROVE, codex flagged the dry-run (now fixed)

Honest note the builder made: `loadState` returns `utils`/`annotations` unscoped by workspace_path. Both tables are vestigial (no producers), so unscoped access is harmless today but is a latent future-issue if a producer ever gets wired up. Explicitly deferred as future cleanup rather than papering it in this PR — right call for scope discipline.

### Recommendation

**APPROVE — merge when ready per your team-feedback hold.**

The additional round hardened the PR meaningfully: `send.ts` was a real regression that would have shipped without the audit (Codev's own `afx send` breaks post-migration from any builder worktree), and the dry-run fix restores the preview UX contract. Neither introduces new risk; both are minimal, tested, and correctly scoped.

---
Architect integration re-review

waleedkadous:
## Architect CMAP — 3-way integration review

Thank you for this — it's an exceptionally thorough PIR. The root-cause writeup in #1118 is one of the clearest I've read, and the fact that the review iterations caught *real, severe* bugs (the `clearRuntime` cross-workspace wipe especially) is exactly what the multi-pass process is for.

I ran an independent 3-way integration review. **All three models: APPROVE, HIGH confidence, no merge blockers.**

|---|---|---|
| Gemini | APPROVE | HIGH |
| Codex | APPROVE | HIGH |
| Claude | APPROVE | HIGH |

### What the panel affirmed
- **This is a root-cause fix, not a patch** — relocating runtime state into the genuinely-global `~/.agent-farm/global.db` rather than adding another `workspace_path`-style workaround on a cwd-local file (the direction #826 went).
- **The `builders` composite-PK reshape** `(workspace_path, id)` is the correct security/compat move, mirroring the v11 treatment of `architect`.
- **The dangerous callsites were the right ones to audit** — `clearRuntime` scoping, the `send.ts` retired-db open, and dry-run side-effect freedom — and they're all fixed and covered by tests.
- **Migration design is genuinely safe**: strict boot one-off (marker + row-copy in one transaction), read-only dry-run preview, and `*.pre-merge-<ts>` rename (never delete) — a real rollback story.
- **Net simplification** — one DB, one connection, one migration chain; ~500 lines of dead local-migration ladder removed.

### Low-severity follow-ups (none blocking)
All safe to defer — noting them so they don't get lost:
1. **`utils` / `annotations` stay unscoped in `loadState()`** (Gemini + Claude). Harmless today (vestigial, no producers), but latent if a producer is ever wired up — a follow-up to add `workspace_path` or formally deprecate them would close the gap.
2. **The optional `workspacePath?` parameter** on the builder read functions is a mild footgun: a future callsite that forgets it silently gets cross-workspace results. A short TSDoc `@param` note ("omit → cross-workspace") would help the next person.
3. **Residual stale `state.db` wording** in a few untouched comments (Codex) — cosmetic.
4. **`consolidate.ts` goes inert** once the migration wave passes — no urgency (it's the recovery escape hatch), just a candidate for an eventual MAINTAIN pass.

### One pre-merge step on my side
Because this is the highest-blast-radius change we ship — it auto-migrates real `~/.agent-farm` state on the next Tower boot — I want to do one backed-up `local-install` smoke test against real data (stop-from-A / start-from-B, confirm A's architects survive) before it goes in. That's an architect responsibility, not a change request for you.

Really nice work. 🙏

waleedkadous (dismissed):
Approving on the strength of the 3-way CMAP above — Gemini, Codex, and Claude all APPROVE at HIGH confidence with no blockers. Thank you again for how thorough this was; the review iterations catching the `clearRuntime` cross-workspace wipe is exactly what the process is for.

Note: merge is still pending one backed-up `local-install` smoke test on my side, since the migration touches live `~/.agent-farm` state on first Tower boot — this approval just clears the review-required gate.

Issue #855, opened 2026-05-26
Monorepo layout: introduce apps/ for end-user surfaces


amrmelsayed:
## Goal

Adopt an `apps/` vs `packages/` split so end-user client surfaces (web UI, VS Code extension, future mobile app) are a discoverable peer set, distinct from shared libraries.

## Motivation

A mobile app is on the horizon. Today all six packages share `packages/*`. Adding `apps/mobile` while leaving the existing end-user clients under `packages/` is the worst option — a new contributor would see `apps/mobile`, `packages/vscode`, `packages/dashboard` and reasonably ask why three end-user clients live in different directories. Half-applied conventions cost cognition without paying for it. Either go all-in on the split or stay all-in on `packages/`.

### Move
- `packages/vscode` → `apps/vscode`
- `packages/dashboard` → `apps/web` (rename — see below)
- New `apps/mobile` lands fresh when built

### Unchanged
- `packages/codev`, `packages/core`, `packages/types`, `packages/config` — stay where they are. Not in scope for this proposal.

### Workspace glob
```yaml
packages:
  - packages/*
  - apps/*
```

### Why `apps/` vs `packages/` at all
End-user clients and shared libraries have different lifecycles, different release cadences, different ownership. Exposing the distinction in directory structure improves navigability and matches widely-used monorepo conventions (Turborepo, Next.js, Vercel templates).

### Why `web` (not `dashboard`)
- **Functional accuracy**: the package exposes the full Tower management surface (workspace overview, builder spawn, file browse, terminal attach, PR view, send messages) — far more than a dashboard
- **Symmetry**: `web` / `mobile` / `vscode` all describe **where the client lives in the user's life**
- **Future-proof**: marketing pages, settings, billing portals would still fit under `web`
- **Convention**: Turborepo's default app name
- **Package-name impact**: `@cluesmith/codev-dashboard` is `private: true` — renaming has zero external blast radius (referenced only internally via `workspace:*`)
- **User-facing brand stays separate**: the product surface can still be called "Dashboard" to users if desired; this is a directory decision, not a brand decision

### Why `mobile` (not `native`)
- All other directory names describe **surface** (where the client lives). `native` would describe **implementation tech**, breaking the axis.
- React Native is precisely the ambiguous case — branded "native" but isn't really. `apps/native` amplifies the confusion rather than resolving it.
- Stable under tech refactor: `mobile` stays accurate whether the implementation is React Native, Flutter, or future native Swift+Kotlin.
[...]

PR #1136, opened 2026-07-03
[Spec 1134] afx whoami + ship /arch-init command (multi-architect identity)

### LINKED ISSUE #1134: afx whoami + ship /arch-init command in codev (multi-architect identity)
## Summary

Two related improvements to support multi-architect workspaces:

1. **New `afx whoami` command** — reports, from Tower/global.db's perspective, who the current terminal's agent is:

   ```
   workspace: codev
   type: architect
   name: main
   ```

   For builders:

   ```
   workspace: codev
   type: builder
   name: builder-spir-984
   architect: main        # spawnedByArchitect, when known
   ```

   Should support `--json`. When identity cannot be determined (plain shell, unregistered terminal), say so clearly — fail loud, no guessing (see #1094 for why silent identity fallbacks are dangerous).

2. **Move the `/architect` slash command into codev itself, renamed `/arch-init`** — today it lives only in the architect's personal `~/.claude/commands/architect.md`. It should ship with codev (skeleton → all adopting projects) so any architect terminal in any codev workspace can run `/arch-init` to adopt its identity and recover state from `codev/state/<name>.md`.

   **`/arch-init` must use `afx whoami` for identity resolution** — replacing the current fragile `ps -p $PPID` process-ancestry matching against `afx status` output. If `whoami` resolves the identity, use it; an explicit `[name]` argument still overrides; if neither, stop and ask the human (never guess).

## Existing plumbing to build on

- Tower already sets `CODEV_ARCHITECT_NAME` in architect PTY env (`packages/codev/src/agent-farm/servers/tower-terminals.ts:654` and `:897`) — the natural signal for architect identity.
- `detectCurrentBuilderId()` in `packages/codev/src/agent-farm/commands/send.ts:111` resolves builder identity from cwd + global.db, scoped by `workspace_path`, failing loud (`BuilderIdResolutionError`) per #1094. `whoami` should reuse/share this, not duplicate it.
- Architect rows (name, pid, terminal id) live in the `architect` table in `~/.agent-farm/global.db` (#1118).

## Design notes / constraints

- Identity precedence for `whoami`: builder-worktree cwd match → `CODEV_ARCHITECT_NAME` env → unknown (exit non-zero with a clear message). Do not fall back to "main" implicitly.
- `/arch-init` content should be generalized (the current `~/.claude/commands/architect.md` is Shannon-workspace-flavored) — workspace-agnostic wording, same guardrails (never auto-approve gates, only touch your own builders, never cd into worktrees, stay on default branch at root).
- Ship `/arch-init` via `codev-skeleton/` so `codev init` / `codev update` installs it; mirror in our own instance. Remember the two-tree rule: change BOTH `codev/` and `codev-skeleton/` where applicable.
- `whoami` must work from both the main workspace root (architect) and inside a `.builders/<id>/` worktree (builder).

## Acceptance criteria

- `afx whoami` from an architect terminal prints workspace/type/name matching Tower's records.
- `afx whoami` from a builder worktree prints the canonical builder id (+ spawning architect when recorded).
- `afx whoami` in an unrecognized context exits non-zero with a helpful message.
- `/arch-init [name]` resolves identity via `afx whoami` (arg overrides), reads `codev/state/<name>.md`, and reports the resume summary.
- `/arch-init` ships in the skeleton and resolves in fresh installs.

waleedkadous:
## Summary

Adds `afx whoami` — a fail-loud identity diagnostic reporting the current terminal's agent (workspace, type, name) from Tower/global.db's perspective — and ships the `/arch-init` architect identity-adoption command with codev via the skeleton, replacing the fragile personal `ps -p $PPID` ancestry-matching command.

Closes #1134

## Changes

- **`afx whoami`** (`commands/whoami.ts` + cli.ts registration): precedence builder-worktree cwd match (reuses `detectCurrentBuilderId`, incl. the #1094 no-fallthrough rule) → `CODEV_ARCHITECT_NAME` → exit 1 with explanation. Never implicitly `main`. `--json` support; builders carry `architect:` (spawning architect) when recorded; workspace display name from `known_workspaces` with basename fallback; best-effort architect-row cross-check warns on stderr only. Strictly read-only: `lookupBuilderSpawningArchitect` gains an optional `db?` handle (default `getDb()`, existing callers unaffected). Works without Tower.
- **`/arch-init` skill** byte-identical in `codev-skeleton/.claude/skills/arch-init/` and `.claude/skills/arch-init/`: validated explicit `[name]` (architect-name rule, path-traversal guarded) or `afx whoami` resolution; STOP-and-ask when neither resolves; state recovery from `codev/state/<name>.md` excluding builder `*_thread.md` files; the four architect guardrails; workspace-agnostic.
- **Docs**: `afx whoami` sections in `resources/commands/agent-farm.md` and afx `SKILL.md` in BOTH trees (identical section content). Drive-by: three stale "state.db" wordings in `send.ts` now say "global.db" (post-#1118).

## Testing

- 20 new whoami unit tests (spec scenarios 1–8, 10, 11 + JSON shapes, fixture pattern from bugfix-774)
- Skill-text assertion tests (cross-tree byte equality, required/forbidden content) + first-ever `copySkills()` regression tests (install path for scenario 9)
- Full unit suite green: 3470 passed / 48 pre-existing skips
- Manual smokes: `afx whoami` from this builder worktree → `builder-spir-1134` + `architect: main`; from `/tmp` → exit 1 fail-loud; `/arch-init` skill confirmed loading in Claude Code
- CMAP: all three implement phases unanimously approved (Gemini/Codex/Claude)

## Spec
Link: codev/specs/1134-afx-whoami-ship-arch-init-comm.md

## Review
Link: codev/reviews/1134-afx-whoami-ship-arch-init-comm.md


waleedkadous:

Excellent work — this is a clean, well-composed change. Independent integration review (Claude): **APPROVE / HIGH confidence, no key issues**, on top of the unanimous CMAP at every phase.

Highlights from the integration pass:

- **Backward compatibility verified**: all 8 existing callers of `lookupBuilderSpawningArchitect` (2 production in `tower-messages.ts`, 6 in tests) use 1–2 args — the optional `db?` param is fully additive.
- **Resource lifecycle correct**: the read-only db handle closes on every return path; the `BuilderIdResolutionError` path never opens it at all.
- **Fail-loud semantics preserved end-to-end**: no fallthrough from a builder worktree to the env check (#1094 rule), no implicit `main` anywhere.
- **The `copySkills()` regression tests are a gift beyond scope** — first coverage for the whole skill-delivery pipeline, and the skill-text-assertion pattern (declarative artifact as the testable unit) is a reusable recipe.

Non-blocking notes for the record (both already scoped out deliberately, agreed):
- `describeStateDbOpenFailure` name still says StateDb — future cosmetic drive-by.
- whoami→send.ts command-to-command import — extract to a shared identity module when a third consumer appears.

waleedkadous:
@amrmelsayed I've requested your review on this one as the final pass — would value your opinion, especially since `afx whoami` reads directly from the `global.db` layout your state-consolidation work (PR #1127) put in place. Areas where your eye would help most:

- The read-only access pattern into `global.db` (`openReadonlyGlobalDb` + the optional `db?` handle on `lookupBuilderSpawningArchitect`) — does this fit how you'd want readers to consume the consolidated db?
- The identity precedence (builder-worktree cwd → `CODEV_ARCHITECT_NAME` → fail loud) and whether any multi-workspace edge cases from the consolidation are missed.
- Anything else that stands out.

No rush — the merge waits on your review. Thanks!

amrmelsayed:
## Review

Approving. I verified this against today's `main` rather than the July 3 base, since the branch predates #1149, #1150, and the #855 monorepo restructure: the merge is clean, `tsc --noEmit` on the merged tree is clean (confirming the `db?` param is fully additive against all current callers), and all four test files pass on the merged tree (59 tests). Schema references (`architect`, `known_workspaces`, and the #1118 composite-PK `builders`) all match the current `GLOBAL_SCHEMA`.

On the two areas you flagged:

**Read-only access into global.db**: this fits how I'd want readers to consume the consolidated db. Identity never depends on the handle (only enrichment does), the connection closes on every path including the fall-through to the fail-loud throw, and crucially whoami never triggers `getDb()`, so a diagnostic command can never take a write lock or kick off a migration mid-crash-recovery. That's the right property for the environments whoami serves. One note for later: the `db` default param creates mixed ownership semantics (singleton must not be closed, passed handle is caller-owned). Documented and fine at two consumers; when a third read-only consumer appears, promote `openReadonlyGlobalDb` into `db/index.ts` rather than copying it, same "third consumer" rule this PR already applies to the identity helpers.

**Identity precedence / multi-workspace edges**: handled correctly. The builder enrichment passes `workspaceRoot`, so the composite-PK lookup disambiguates same-id builders across repos, and both whoami queries normalize through `normalizeWorkspacePath`, which is the same helper `state.ts` aliases as `canonicalize`, so readers and writers agree on the symlink-dereferenced form. Two behavioral edges worth having on the record, neither a change request:

- cwd beats env: an architect who cd's into a builder worktree and runs `afx whoami` is reported as the builder. That's the spec's precedence, `/arch-init` handles it safely (`type: builder` → STOP), and "never cd into a builder worktree" is already a guardrail.
- Env inheritance: with `CODEV_ARCHITECT_NAME` inherited into subshells, whoami from an unrelated git repo reports `type: architect` with that repo's basename as workspace. The `rowMissing` stderr warning covers this adequately.

**One follow-up this PR surfaces (non-blocking, spec 1134 scoped state-file lifecycle out deliberately)**: `/arch-init` promotes architect state files (`codev/state/<name>.md`) from a private convention into a shipped concept, but codev currently takes no stance on their versioning: they're neither tracked nor gitignored today. Unlike builder threads (committed, land on main at merge), architect state files are inherently **per-person**: every team member working the same repo has their own `main` architect, so a committed `main.md` collides across members, and each `/arch-init` would resume from whoever pushed last. They should be gitignored while keeping builder threads tracked:

```
codev/state/*.md
!codev/state/*_thread.md
```

and that rule should be embedded across the codev CLI surface, not just documented: `codev init` and `codev adopt` scaffold it into the generated `.gitignore`, `codev update` backfills it for existing installs, and `codev doctor` warns when the rule is missing or an architect state file is already tracked. Will file a follow-up issue.

Also noting the `copySkills()` regression tests are first-ever coverage for the skill-delivery install path, and the cross-tree byte-equality test is the right enforcement for the skeleton/instance invariant.

Good to merge from my side.

amrmelsayed (approved):
Formal approval per my full review in https://github.com/cluesmith/codev/pull/1136#issuecomment-4998064325. Verified against current main (clean merge, clean typecheck, 59 tests passing on the merged tree). Good to merge.

PR #1143, opened 2026-07-06
[Bugfix #1142] Expose exitCode to cron conditions; let condition-true failure runs deliver alerts

### LINKED ISSUE #1142: tower-cron: conditions can't see exitCode (ReferenceError every run) and failure runs can never deliver alerts
## Symptom

`~/.agent-farm/tower.log` fills with a recurring WARN (every 15 min for months, survives Tower restarts):

```
[WARN] Condition evaluation failed for 'Service Health Check': ReferenceError: exitCode is not defined
```

The triggering task is a user health-check cron (`.af-cron/*.yaml`) with:

```yaml
condition: "exitCode != 0"
message: "Service Health Alert:\n${output}"
target: architect
```

## Root cause (two layers, both in `packages/codev/src/agent-farm/servers/tower-cron.ts`)

1. **`evaluateCondition()` (line ~293)** builds `new Function('output', 'return ' + condition)` — only `output` is in scope. Any condition referencing `exitCode` throws `ReferenceError`, which is caught at line ~251, logged as WARN, and forces `shouldNotify = false`. The docs/UX imply exit-code awareness; the evaluator doesn't provide it.

2. **Delivery gate (line ~256)** is `if (shouldNotify && result === 'success')`. A monitoring script that detects a problem and exits non-zero produces `result = 'failure'`, whose branch only logs a WARN ("infrastructure noise"). Consequence: **"alert me when this command fails" is inexpressible** — the alert is suppressed on exactly the runs that matter. (In the wild, task authors work around it by pushing notifications from inside the script.)

## Prescribed fix (design decided — carry it in BUGFIX)

1. **Capture the exit code** in `runCommand()`: in the `exec` callback, `error.code` (number) is the exit code when the process ran and exited non-zero. Resolve with `{ output, exitCode }` (0 on success) instead of rejecting for plain non-zero exits. Keep rejecting/flagging genuinely infrastructural errors (spawn failure, timeout/kill — `error.killed`/`error.signal`) as today's `failure` class.

2. **Expose `exitCode` to conditions**: `new Function('output', 'exitCode', 'return ' + condition)`, invoked with both. Conditions using only `output` are unchanged (backward compatible).

3. **Rework the delivery gate**: deliver when the condition (now able to see `exitCode`) evaluates true, regardless of the old success/failure split:
   - Task **with** a `condition`: delivery is decided by the condition alone (a non-zero exit is data, not noise).
   - Task **without** a `condition`: keep today's behavior (deliver only on exit 0) so existing no-condition tasks don't start alert-spamming on flaky commands.
   - True infrastructure errors (spawn fail/timeout) keep the existing WARN-only path, with `exitCode` reported as non-zero (e.g. 124/-1) so `exitCode != 0` conditions still fire if the author wants them to.

4. **State row semantics**: `last_result` in `cron_tasks` should record success/failure by exit code as before — only the *delivery* logic changes.

5. **Docs**: document the condition environment (`output: string`, `exitCode: number`) wherever `.af-cron` YAML is documented (afx skill + agent-farm.md), BOTH trees.

## Acceptance

- A task with `condition: "exitCode != 0"` produces no ReferenceError; delivers its message exactly when the command exits non-zero.
- A task with `condition: "output.includes('FAIL')"` behaves exactly as today.
- A task with no condition delivers only on exit 0 (unchanged).
- Unit tests cover: exitCode condition true/false, output-only condition regression, no-condition regression, timeout path.

waleedkadous:
## Summary

Fixes the two-layer bug in `tower-cron.ts` where (1) conditions referencing `exitCode` threw `ReferenceError` on every run, and (2) failure runs could never deliver alerts — making "alert me when this command fails" inexpressible.

## Root Cause

- `evaluateCondition()` built `new Function('output', 'return ' + condition)` — only `output` was in scope, so `condition: "exitCode != 0"` threw `ReferenceError` every run (12k+ WARN lines in a real tower.log). The catch forced `shouldNotify = false`.
- `runCommand()` rejected on any non-zero exit, so the exit code was never available as data.
- The delivery gate `if (shouldNotify && result === 'success')` suppressed delivery on exactly the runs a monitoring task cares about.

## Fix

- `runCommand()` resolves `{ output, exitCode }`. A plain non-zero exit (numeric `error.code`, no kill signal) resolves with that exit code — it's data, not an error. Spawn failures and timeout/kills still reject, reported as `exitCode -1` / `124` respectively.
- `evaluateCondition(condition, output, exitCode = 0)` exposes both variables to the condition. Output-only conditions are unchanged.
- Delivery gate: a task **with** a `condition` delivers exactly when the condition is truthy (including failed runs). A task **without** a `condition` keeps today's behavior — deliver only on exit 0 — so existing tasks don't start alert-spamming on flaky commands. Non-delivering failures keep the WARN log.
- `last_result` in `cron_tasks` still records success/failure by exit code — only delivery logic changed.
- Docs: the condition environment (`output: string`, `exitCode: number`) is now documented in the afx skill's `## afx cron` section and a new `### afx cron` section in `agent-farm.md`, in BOTH trees (instance + skeleton).

## Test Plan

- 8 new regression tests in `tower-cron.test.ts` (34 total, all pass):
  - `exitCode != 0` condition delivers on non-zero exit, no ReferenceError
  - `exitCode != 0` condition does not deliver on clean exit
  - no-condition task does not deliver on non-zero exit (regression)
  - output-only conditions unchanged alongside exitCode (regression)
  - timeout reported as exitCode 124 — condition can still fire; without a condition it stays WARN-only
  - `evaluateCondition` unit coverage for exitCode scope + default
- Full suite: 3440 passed, 48 pre-existing skips. Root `pnpm build` passes.

Fixes #1142

### DIFF
--- a/.claude/skills/afx/SKILL.md
+++ b/.claude/skills/afx/SKILL.md
@@ -153,7 +153,20 @@ afx cron enable <name>          # Enable
 afx cron disable <name>         # Disable
 ```
 
-There is NO `afx cron add` — create YAML files in `.af-cron/` directly.
+There is NO `afx cron add` — create YAML files in `.af-cron/` directly:
+
+```yaml
+name: Service Health Check      # required, unique per workspace
+schedule: "*/15 * * * *"        # required, cron expression (or @hourly/@daily/@startup)
+command: ./health-check.sh      # required, run via shell
+message: "Health alert: ${output}"  # required, ${output} = trimmed command output
+condition: "exitCode != 0"      # optional JS expression, see below
+target: architect               # optional, default architect
+timeout: 30                     # optional, seconds, default 30
+enabled: true                   # optional, default true
+```
+
+`condition` is a JavaScript expression with two variables in scope: `output` (string — the command's trimmed output) and `exitCode` (number — 0 on success, the command's exit code on non-zero exit, 124 on timeout, -1 on spawn failure). With a `condition`, the message is delivered exactly when it evaluates truthy — including on failed runs (e.g. `exitCode != 0` alerts when the command fails). Without a `condition`, the message is delivered only when the command exits 0.
 
 ## Other commands
 
--- a/codev-skeleton/.claude/skills/afx/SKILL.md
+++ b/codev-skeleton/.claude/skills/afx/SKILL.md
@@ -131,7 +131,20 @@ afx cron enable <name>          # Enable
 afx cron disable <name>         # Disable
 ```
 
-There is NO `afx cron add` — create YAML files in `.af-cron/` directly.
+There is NO `afx cron add` — create YAML files in `.af-cron/` directly:
+
+```yaml
+name: Service Health Check      # required, unique per workspace
+schedule: "*/15 * * * *"        # required, cron expression (or @hourly/@daily/@startup)
+command: ./health-check.sh      # required, run via shell
+message: "Health alert: ${output}"  # required, ${output} = trimmed command output
+condition: "exitCode != 0"      # optional JS expression, see below
+target: architect               # optional, default architect
+timeout: 30                     # optional, seconds, default 30
+enabled: true                   # optional, default true
+```
+
+`condition` is a JavaScript expression with two variables in scope: `output` (string — the command's trimmed output) and `exitCode` (number — 0 on success, the command's exit code on non-zero exit, 124 on timeout, -1 on spawn failure). With a `condition`, the message is delivered exactly when it evaluates truthy — including on failed runs (e.g. `exitCode != 0` alerts when the command fails). Without a `condition`, the message is delivered only when the command exits 0.
 
 ## Other commands
 
--- a/codev-skeleton/resources/commands/agent-farm.md
+++ b/codev-skeleton/resources/commands/agent-farm.md
@@ -478,6 +478,42 @@ afx tower status [options]
 
 ---
 
+### afx cron
+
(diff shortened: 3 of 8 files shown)


waleedkadous:
## Architect Review

Nice, tight execution — this carries the issue's prescribed design faithfully, and the docs work goes beyond the ask in a good way.

Low-risk change (production diff is 37/22 in one file, `tower-cron.ts`; the rest is 204 lines of regression tests + doc sync in both trees). Verified against the issue:

- `runCommand` resolves `{output, exitCode}` for plain non-zero exits (exit code is data), still rejects only for spawn failure and timeout/kill — with the coreutils-style `124`/`-1` mapping in the catch. ✓
- `evaluateCondition(condition, output, exitCode = 0)` — backward-compatible default, output-only conditions untouched. ✓
- Delivery gate: condition-bearing tasks deliver on the condition alone; no-condition tasks keep exit-0-only delivery (no alert-spam regression). ✓
- `last_result` still records success/failure by exit code. ✓
- Docs: full `.af-cron` YAML format + condition environment documented in the afx skill AND agent-farm.md, both trees — the skeleton's agent-farm.md previously had no cron section at all, so this fills a real gap. ✓

One intentional semantic change worth naming for the record (it's documented, and correct): an *output-only* condition that evaluates true on a failed run now delivers, where before failure suppressed everything. That's the point of the fix — "condition alone decides" — just noting it's a behavior change for pre-existing condition-bearing tasks, not a pure bugfix.

CMAP unanimous (3× APPROVE/HIGH), CI 6/6 green.

PR #1175, opened 2026-07-15
[Bugfix #1174] Keep childless architects visible in Agents view (architect axis)

### LINKED ISSUE #1174: vscode: architects disappear from Agents view (architect axis) when their last builder is cleaned up
## Problem

In the Agents view under the **architect axis**, an architect disappears from the tree the moment its last builder is cleaned up. The row is present while any builder is attached; the row vanishes when zero builders remain.

Concrete failure the user reported: architect had a builder → user cleaned up the builder (`afx cleanup --project N`) → the architect vanishes from the Agents view → user has to switch back to the Workspace view to find it (and the click affordance to open its terminal).

The workflow that surfaces this is common: user is looking at Agents to triage / monitor a specific architect's work, that architect finishes its last builder, the row disappears mid-session, next click to open the terminal is now two-view-hops away.

## Root cause

`packages/vscode/src/views/builders.ts` builds group headers from the builders list. Under the architect axis, groups are keyed by `spawnedByArchitect` values *observed in the current builders array*. Architects with zero builders contribute no key → no group header emitted.

The current behavior was documented as intentional in the v3.2.2 CHANGELOG:

> *"In architect mode only architects with in-flight builders appear as group headers — childless architects don't clutter the work view; the full architect roster lives in Workspace > Architects."*

The rationale was "workload views filter by activity; roster views enumerate membership." That principle is defensible for workload flows, but it misses the specific case where a row disappears out from under the user during a session (the architect existed a moment ago, and the user's mental model still expects it to be reachable from where they last saw it).

## Scope

**In scope**: architect axis only. Under the architect axis, enumerate every registered architect in the workspace as a group header — regardless of whether that architect currently owns any builders. Childless architects render as `MAIN (0)`, `REVIEWER (0)`, etc., with the same click-to-open-terminal affordance the populated headers already carry.

**Out of scope**: stage and area axes. Those group builders by builder-intrinsic properties (phase, area); there is no natural "empty stage" or "empty area" concept because stages and areas exist only as properties of builders, not as independently-registered entities. Those axes stay behavior-unchanged.

**Out of scope**: reshape of Workspace view's Architects section or the Agents view's overall shape. That's a separate larger design question. This bugfix does NOT remove or alter the Workspace view; both views continue to list architects. The redundancy under the architect axis is deliberate — this issue is about the workload view no longer losing rows out from under the user, not about consolidating the roster.

**Out of scope**: single-architect workspaces. Today the architect axis picker item is hidden when only one architect exists (no toggle available). This bugfix does not change that.

## Fix sketch

`packages/vscode/src/views/builders.ts` (the file that builds group headers):

- When the active grouping axis is `architect`, source the group-header set from the **architect roster** (via the existing state hook the Workspace view already uses), not from the builders array's `spawnedByArchitect` values
- For each architect, include the header even if the intersection of `builders.filter(b => b.spawnedByArchitect === architect.name)` is empty
- Child list for empty groups is `[]` — the group renders as a leaf-like row with a count of 0
- Header count `(0)` displays alongside populated counts `(2)`, `(1)`, etc. — same format
- The click-to-open-terminal command on the header (already bound to `codev.openArchitectTerminal` per the existing architect-axis logic) works identically for `(0)` groups

Rollup icon on the header (currently derived from `rollupGroupState(builders)`) — for empty groups, use a neutral / idle glyph rather than the blocked/idle/active tri-state; add a matching case to `worstBuilderState` or short-circuit before calling it.

## Acceptance

- With N architects and 0 builders, Agents view under architect axis shows N group headers, each `(0)`
- With N architects and builders spread across some of them, Agents view shows all N headers, each with its correct count
- Cleaning up the last builder of an architect keeps that architect visible in Agents view
- Under stage / area axes: no change in behavior
- Single-architect workspace: no change (axis picker still hidden when only one architect)
- Click on a `(0)` header opens the architect's terminal
- Rollup icon on empty groups is a neutral / idle glyph, not the yellow "attention" bell

## Scope discipline

- `packages/vscode/src/views/builders.ts` — grouping-provider change (source header set from architect roster under architect axis)
- Possibly `packages/vscode/src/views/builder-row.ts` — extend `worstBuilderState` / `rollupGroupState` for empty groups
- Tests: cover 0-builder architect + populated architect combinations under architect axis
- No package.json changes, no context-menu changes, no view-registration changes
- No changes to stage / area axes

## Related

- Design conversation on 2026-07-13 (in-session) about consolidating Workspace > Architects into Agents view. That reshape is separate and unfiled — this bugfix is intentionally narrower.
- v3.2.2 CHANGELOG entry that documented the "hide childless architects" behavior as intentional. That reasoning was defensible for one framing (workload views filter by activity) but this issue documents the user-facing cost that outweighs it (rows disappear mid-session).

amrmelsayed:
Fixes #1174

## Summary

Under the **architect axis** of the Agents view, an architect disappeared from the tree the moment its last builder was cleaned up. The row was present while any builder was attached and vanished at zero builders — so a common triage flow (watch an architect's work, its last builder finishes, click to open its terminal) lost the row out from under the user mid-session, forcing a two-view-hop back to Workspace > Architects.

This sources the architect-axis group-header set from the **registered-architect roster** rather than from the `spawnedByArchitect` values observed on the builders array. Every registered architect now gets a header, populated or not. Scope is architect-axis only; stage and area axes are unchanged.

## Root Cause

`packages/vscode/src/views/builder-grouping.ts` — `architectGrouping().group()` bucketed architects solely from `spawnedByArchitect` values present on the current builders array. A childless architect contributes no key → no group header. The v3.2.2 CHANGELOG documented this "hide childless architects" behavior as intentional; this issue documents the user-facing cost (rows disappearing mid-session) that outweighs that rationale for the architect axis.

## Fix

- **`builder-grouping.ts`**: `BuilderGrouping.group` gains an optional `roster?: readonly string[]`. `architectGrouping` seeds an empty bucket for every roster architect first, so childless architects still produce a header. A builder whose owner isn't in the roster (stale / non-live) still gets its owner header from the builder loop — no builder is ever dropped. Stage/area axes ignore the arg.
- **`builders.ts`**: passes the roster — `(data.architects ?? []).map(a => a.name)`, already on the overview cache the provider holds (no second fetch; `??` guards a pre-roster payload) — into both `group()` call sites. Empty groups render as leaf-like rows (`CollapsibleState.None`) rather than empty accordions, while still carrying the existing `codev.openArchitectTerminal` click command.
- **`builder-tree-item.ts`**: an all-zero rollup (only the architect axis produces one) short-circuits the worst-of tri-state — which would read all-zero as the green `active` glyph — and uses a neutral idle glyph (`circle-outline` / `disabledForeground`) with a `No builders` tooltip. Never the yellow attention bell.

Net diff ~88 LOC across 3 source + 2 test files.

## Test Plan

New/updated tests (`pnpm vitest run` → **627 pass**, 53 files; `check-types` + `lint` clean):

- `builder-grouping.test.ts` — every roster architect gets a header incl. childless ones; zero-builders-with-roster yields all-empty headers; a stale non-roster owner is never dropped; the no-roster back-compat path still shows only owners-of-builders.
- `builders-childless-architect-header.test.ts` (new, provider-level) — a childless architect renders a `(0)` leaf-like (`None`) header, carries `codev.openArchitectTerminal` with the right name, shows the neutral `circle-outline`/`disabledForeground` glyph (not the bell), and the stage axis ignores the roster (no architect-name headers).

## Acceptance (from the issue)

- ✅ N architects, 0 builders → N `(0)` headers
- ✅ N architects with builders spread across some → all N headers, correct counts
- ✅ Cleaning up an architect's last builder keeps it visible
- ✅ Stage / area axes: no change
- ✅ Single-architect workspace: no change (axis picker gating untouched)
- ✅ Click on a `(0)` header opens the architect's terminal
- ✅ Neutral / idle glyph on empty groups, not the yellow bell

### DIFF
--- a/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
+++ b/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1174
 title: vscode-architects-disappear-fr
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-15T01:30:06.203Z'
-updated_at: '2026-07-15T01:33:13.909Z'
+updated_at: '2026-07-15T01:38:38.103Z'
--- a/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
+++ b/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1174
 title: vscode-architects-disappear-fr
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-15T01:30:06.203Z'
-updated_at: '2026-07-15T01:30:06.203Z'
+updated_at: '2026-07-15T01:33:13.909Z'
--- a/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
+++ b/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-07-15T01:42:56.921Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-15T01:30:06.203Z'
-updated_at: '2026-07-15T01:38:38.103Z'
+updated_at: '2026-07-15T01:42:56.921Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
+++ b/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
@@ -6,11 +6,12 @@ plan_phases: []
 current_plan_phase: null
 gates:
   pr:
-    status: pending
+    status: approved
     requested_at: '2026-07-15T01:42:56.921Z'
+    approved_at: '2026-07-15T03:23:12.178Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-15T01:30:06.203Z'
-updated_at: '2026-07-15T01:42:56.921Z'
-pr_ready_for_human: true
+updated_at: '2026-07-15T03:23:12.178Z'
+pr_ready_for_human: false
--- a/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
+++ b/codev/projects/bugfix-1174-vscode-architects-disappear-fr/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1174
+title: vscode-architects-disappear-fr
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-07-15T01:30:06.203Z'
+updated_at: '2026-07-15T01:30:06.203Z'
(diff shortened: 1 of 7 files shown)


amrmelsayed:
## CMAP review (3-way, PR phase)

| Model | Verdict | Confidence | Issues |
|---|---|---|---|
| Gemini | APPROVE | HIGH | None |
| Codex | APPROVE | MEDIUM | None |
| Claude | COMMENT | HIGH | 1 non-blocking style nit |

**Claude's nit** (addressed in the follow-up commit): the `collapsibleState` assignment used a ternary; converted to `if/else` to match the project's no-ternaries convention. No behavior change.

All three confirm: root cause correctly identified, fix well-scoped to the architect axis (stage/area untouched), regression tests fail without the fix and pass with it. Full suite 627 pass; `check-types` + `lint` clean.

Issue #1062, opened 2026-06-17
Harden harness resolution: unrecognized override command defaults to claude harness (residual --resume crash-loop)


mohidmakhdoomi:
## Context

Follow-up from the architect integration re-review of #929 / PR #1059 (codex CMAP finding). #929 made `getArchitectHarness`/`getBuilderHarness` **override-aware**, so a *recognized* non-claude override command (`TOWER_ARCHITECT_CMD=gemini`, `--architect-cmd codex`, `--builder-cmd gemini`) now resolves its own harness instead of claude — closing the `<cmd> --resume <claude-uuid>` crash-loop for codex/gemini/opencode.

## Residual edge

`resolveHarness()` (`packages/codev/src/agent-farm/utils/harness.ts:~330-340`) auto-detects the harness from the command basename and, when nothing matches **and** there is no explicit `shell.architectHarness` / `shell.builderHarness`, falls back to `CLAUDE_HARNESS` — which implements `buildResume`.

So an *unrecognized* override command (e.g. `TOWER_ARCHITECT_CMD=bash`, or a wrapper script `my-cli.sh`) with no matching harness config still resolves the claude harness. With a stale Claude `.jsonl` for that cwd, `launchInstance` then builds `<unknown-cmd> --resume <claude-uuid>` — the same crash-loop class #929 set out to kill, shifted from gemini/codex to "unknown command."

## Severity / scope

- **Not a regression.** Pre-#929 the architect resume path read the claude session store unconditionally, so this unrecognized-command path was already vulnerable; #929 strictly improved things (fixed recognized CLIs) without worsening this sliver.
- **Pre-existing, broad behavior.** "Unknown command → default to `CLAUDE_HARNESS`" governs *all* harness capabilities (role injection via `--append-system-prompt` too), not just resume — so a proper fix is a design decision about unknown-command harness semantics, separable from #929.
- **Narrow exposure.** Requires an unrecognized override command + no explicit `*Harness` config + a stale Claude jsonl for the cwd. The recommended `.codev/config.json`-driven path is unaffected.

## Proposed fix (for discussion)

[...]

Issue #1018, opened 2026-06-09
Builder worktree write-guard: prevent writes anchored at the main checkout root


amrmelsayed:
## Problem

A strict-mode builder works in an isolated git worktree at `.builders/<id>/`, which is its git root. The `Write` and `Edit` tools require **absolute** paths. The builder synthesized an absolute path anchored at the repo's canonical root (e.g. `<repo>/codev/plans/...`), which drops the `.builders/<id>/` segment and points at the **main checkout's** working tree instead of the worktree.

Because the worktree is nested inside the main checkout, that wrong path is a real, writable directory, so the mis-write **succeeds silently**: the file lands in the main checkout. The error only surfaces later when `git add` from the worktree fails with a pathspec error (the worktree never received the file). The byte-identical trees at branch base also mean wrong-rooted *reads* succeed silently, masking the mistake until the first write.

Net effects: wasted builder turns, and (worse) silent pollution of the main checkout's working tree by a builder that is supposed to be isolated.

## Regression framing (read this first)

This is **not** a codev code regression, and a `git bisect` for a culprit commit will not find it. Two pieces of evidence:

1. It has occurred across multiple days, including before the most recent local install of the latest codev build, so it does not correlate with any recent codev change.
2. The most recent change to the spawn-prompt machinery inlines **worktree-relative** paths (`codev/plans/<id>-<slug>.md`, "open in the worktree"), so the prompt content pushes toward worktree-relative thinking, not main-root paths. It does not explain the bug.

The most likely root cause is **intrinsic path-synthesis behavior of the current builder model / CLI runtime**. Both `Read` and `Write` require absolute paths, so the builder must synthesize one. It synthesizes the canonical repo root it infers rather than its actual worktree cwd. Wrong-rooted reads then return identical content (byte-identical trees), giving the model no corrective signal, so it keeps reusing the bad prefix until a write to the worktree's git index exposes it.

Implication for the fix: this behavior will drift across model and CLI upgrades in both directions. Instructions, per-agent memory, and bisects do not hold against a moving runtime. **Only a deterministic guard holds across versions.** That is what this issue should deliver.

## Proposed fix (deterministic guard + backstop)

[...]

Issue #929, opened 2026-05-29
Support `codex` as an architect


mohidmakhdoomi:
> ## ⚠️ Rescoped 2026-06-16 — codex-only architect support (gemini retiring, see #778)
>
> **Final scope:** bring **`codex`** to parity with `claude` as a Codev architect, and **remove `gemini` from architect support** (the Gemini CLI is retired 2026-06-18, #778 — `GEMINI_HARNESS` stays for *builders* only). Conversation resume remains Claude-main-only.
>
> **Why not agy here:** the Antigravity CLI (`agy`) was evaluated as the gemini successor for the architect role, but it has **no system-prompt / model-instructions channel** — the architect role can only ride in agy's first *user* turn (`--prompt-interactive`), a weaker, visible, less-durable injection than claude/codex/gemini's out-of-band mechanisms. Rather than ship a degraded architect, **agy-as-architect is split into a follow-up: #1063.** (agy users still get project context for free — agy reads `AGENTS.md` natively.)
>
> ### Delivered on `builder/pir-929` (PR #1059)
> - Harness-agnostic resume **crash-loop fix** (`HarnessProvider.buildResume` seam) — engine-neutral.
> - **`codex` architect parity** (role via `-c model_instructions_file=`; fresh-launch + reconnect paths).
>
> ### Remaining delta (codex-only revision)
> - Remove `gemini` from architect support: `doctor` no longer affirms gemini as an architect (builder-only); remove the now-dead gemini `getArchitectFiles` + `.gemini/settings.json` gitignore entry, and any architect context-file seam left with no remaining consumer.
> - Tests + docs reflect **claude + codex** architects (drop gemini-architect cases; keep gemini *builder* tests).
> - Push to the same branch (PR #1059 auto-updates); re-CMAP at the PR; do not self-merge.
>
> The original request below is retained for history; its `gemini`-architect (and any agy) specifics are **superseded** by the codex-only scope above.

---

## Feature Request

Allow the OpenAI `codex` CLI and Google `gemini` CLI to run as Codev **architects** (the primary interactive agent that spawns and reviews builders), at parity with the default `claude` architect. Selectable via `shell.architect` / `shell.architectHarness` in `.codev/config.json` — exactly the config-driven mechanism builders already use to pick their CLI.

## Why / Current State

[...]

Issue #1189, opened 2026-07-16
Introduce packages/codev-sdk: client SDK for Tower (server/client dependency isolation)


amrmelsayed:
## Goal

Introduce `packages/codev-sdk` (`@cluesmith/codev-sdk`): the single client implementation of "how anything talks to Tower". Restructure the shared-package taxonomy so server and client dependencies are isolated at the package level:

- `codev-types`: wire contracts. The only package imported by both sides. Unchanged.
- `codev-core`: server-side implementation only (local-key auth issuance, homedir-derived path constants, and over time Tower internals).
- `codev-sdk` (new): Tower API client, WS/message-bus client, and the client-side pure helpers. Consumed by the web dashboard, the VS Code extension, the CLI's Tower-facing commands, and the future mobile app.

Invariant: `codev-core` and `codev-sdk` never import each other; both import `codev-types`.

## Motivation

`codev-core` is currently consumed from both server (Tower, CLI) and client (dashboard browser bundle, VS Code extension host) contexts. The boundary between its Node-only modules and its environment-agnostic ones is enforced only by convention (subpath exports plus reviewer vigilance). Two concrete problems observed:

1. `constants.ts` imports `node:path`/`node:os`, so a pure value like `DEFAULT_TOWER_PORT` is unreachable from any bundler that cannot resolve Node builtins (Metro for the planned mobile app fails loudly on it).
2. Nothing structural prevents a server-side need from adding a runtime dependency that every client then inherits.

Package-level separation makes the package manager enforce the boundary instead of review discipline. The mobile workstream (issue #1147, research in `codev/research/mobile/`) makes a third client environment (React Native / Metro) concrete, but the split stands on its own for the two existing clients.

## Import census (verified 2026-07-16)

Who imports each `codev-core` leaf today:

| Leaf | codev (CLI/server) | dashboard | vscode | Disposition |
|---|---|---|---|---|
| tower-client | CLI commands | - | yes | sdk (auth/transport injected) |
| auth | (via tower-client) | - | yes | issuance stays core; sdk takes injected getToken |
| workspace | yes | - | yes | sdk (URL grammar is contract-adjacent) |
| constants | yes | - | yes | split: pure values to sdk, homedir paths stay core |
| escape-buffer | - | yes | yes | sdk |
| reconnect-policy | Tower (tunnel-client, tower-websocket) | yes | yes | sdk owns it; core keeps a private copy (see exception) |
| agent-names | yes | - | yes | sdk |
[...]

PR #1203, opened 2026-07-19
Support Kimi Code CLI as a builder (PIR #1201)

### LINKED ISSUE #1201: Support Kimi Code CLI as a builder
## Summary

Add the Kimi Code CLI (`kimi`) as a supported **builder** harness, so `.codev/config.json` `shell.builder: "kimi"` / `builderHarness: "kimi"` produces a working builder. Scope is the builder MVI validated by spike task-Iptx; Kimi as an *architect* is explicitly out of scope (stage 2).

Spike findings (verdict: **Feasible with Caveats**, POC-validated end-to-end):
https://github.com/mohidmakhdoomi/codev/blob/builder/task-Iptx/codev/spikes/task-Iptx-kimi-code-cli-support.md
(reproducible probes: `codev/spikes/task-Iptx-kimi-poc.sh` on the same branch)

## What breaks today

- `detectHarnessFromCommand('kimi')` is unrecognized → falls through to the **Claude harness** (#1062): the builder script appends `--append-system-prompt` and a positional prompt, both of which `kimi` rejects (exit 1) → shellper restart loop.
- The false Claude harness also exposes Claude session capabilities, so a stale Claude `.jsonl` can route `--resume <claude-uuid>` into `kimi` (the pre-#929 crash-loop class).

## Validated approach (spike POC)

**Seed-session bootstrap** — solves role injection, initial-prompt delivery, and session identity in one pattern:

1. Seed: `kimi -p "<role + task briefing, ack-and-wait wrapper>" --output-format stream-json` in the worktree; capture the session id from the machine-readable `session.resume_hint` meta line.
2. Persist the id (`.builder-kimi-session`) and loop the TUI with `kimi -S <id> --yolo` — role/task context survives inner restarts.
3. Readiness barrier (required): script prints a `__CODEV_KIMI_SEED_DONE__ <id>` sentinel before exec'ing the TUI; Tower gates delivery on sentinel + grace, then sends a single-line `BEGIN` kick with **store-verified delivery** (poll the session's `state.json`; re-send Enter on timeout; retry once; loud warning after).

## Scope (builder MVI)

- [ ] `KIMI_HARNESS` in `harness.ts` + `detectHarnessFromCommand` recognizing `kimi`
- [ ] Provider-owned builder launch shape (seed + `-S` loop) branched in `spawn-worktree.ts`
- [ ] Readiness barrier + store-verified BEGIN delivery (per spike addendum — required scope)
- [ ] `buildResume` for `afx spawn --resume` (explicit-ID from the Kimi session store, `state.json.workDir` exact match; null → fresh-with-role fallback) + `verifyOwnership`
- [ ] `message-write.ts` per-harness Enter-delay knob (Kimi: 80ms fails, 1s works — bisect and pin)
- [ ] `codev doctor`: `kimi` presence/version, truthful auth heuristic (no billed probe) + `kimi login` guidance
- [ ] Docs: `arch.md` harness section, `.codev/config.json` examples, skeleton mirror where framework files change
- [ ] Tests per the spike's test matrix (incl. the #929-class regression: `kimi` + stale Claude jsonl must never yield `--resume <claude-uuid>` or `--append-system-prompt`)

## Out of scope

- Kimi as **architect** (needs async `seedSession` capability + async-buildable `CrashLoopFallback` per the spike addendum — stage 2, separate issue)
- ACP / `kimi server` adapter
- Write-guard parity (Kimi has no documented hook seam; must be documented as a caveat)

## Caveats (carry into spec/plan)

- Role rides a **user turn**, not a system prompt (same tradeoff that deferred agy, #1063).
- Reliance on undocumented surfaces (session store layout, `session_index.jsonl`, `session.resume_hint`): pin kimi ≥ 0.27.0 and add a session-store smoke probe.
- Documented-Kimi-behavior claims cite only the command reference: https://www.kimi.com/code/docs/en/kimi-code-cli/reference/kimi-command.html

## References

- Spike: task-Iptx (branch `builder/task-Iptx` on the fork above)
- Prior art: PR #1059 (codex architect, PIR #929); issues #832, #1145, #1149, #1150, #1062, #1018

mohidmakhdoomi:
# PIR Review: Support Kimi Code CLI as a builder

Fixes #1201

> **Re-integration notice (2026-08-09).** This PR was parked on two upstream blockers, both now merged (#1317/#1267, #1356/PIR #1233), and `main` moved 900+ commits — including **Spec 1313**, which made `afx send` mailbox-first. `kimi` itself moved 0.27.0 → 0.34.0. The branch has been merged up and the feature **substantially redesigned** against what changed underneath it. The delivery mechanism, the resume mechanism, and the pacing seam are all different from what earlier reviewers saw; the summary below describes the current design, and **"What changed since your last review"** near the end lists the deltas. Re-review from scratch rather than from memory.

## Summary

Adds the Kimi Code CLI (`kimi`, ≥ 0.33.0) as a supported **builder** harness — `shell.builder: "kimi"` / `builderHarness: "kimi"` / `--builder-cmd kimi` now produce a working builder instead of the #1062 false-Claude fallthrough (which appended `--append-system-prompt` and a positional prompt, both rejected by kimi, and could route a stale Claude `--resume <uuid>` into it).

Kimi has no system-prompt flag and takes **no positional prompt**, so role and task travel on two different channels:

- **Role → `--agent-file`** (kimi 0.31.0+). `getWorktreeFiles` writes an agent-definition file into the worktree whose body wraps the role around **`${base_prompt}`** — the token that interpolates kimi's own default system prompt, so the role **extends** rather than replaces it. This is the `claude --append-system-prompt` analogue.
- **Task → the Spec 1313 mailbox.** The generated launch script queues the task with `afx send`, and the render gate delivers it onto a verified-empty composer. Never a direct PTY write, so a boot screen, a busy line, or kimi's folder-trust dialog **holds** the message rather than corrupting or losing it.

Crash restarts resume with the documented, cwd-scoped `kimi -c` — no session id is ever baked into generated bash. Kimi as an *architect* remains out of scope (stage 2); `buildRoleInjection` throws and `doctor` warns, so misconfiguration fails loudly rather than falling through to claude flags.

## The finding that shaped the design

**`kimi -c` does not fail when there is nothing to continue.** It prints `No sessions to continue under "<cwd>"; starting a fresh session.` and starts one anyway — and that session never saw `--agent-file`, so it runs **silently roleless**. That is the #929 hazard class, arriving through a documented flag.

So the launch loop only takes `-c` after an inlined `node -e` store probe proves a conversation exists for this cwd, and the probe **fails closed**: any error (no store, unreadable dir, malformed JSON) exits non-zero and the loop relaunches fresh **with** the role, which is always safe. Tests **execute** that probe against fixture stores and cross-check its verdict against the TypeScript discovery it mirrors, so the hand-written snippet cannot silently drift from `findLatestKimiSessionId`.

The probe asks "**would `kimi -c` continue this?**", not "does a directory exist" — a distinction review had to teach me (see below). Kimi lists a cwd's sessions before continuing one, and that listing drops **archived** sessions and ids it does not recognize, so a session we call resumable but kimi skips lands right back on the roleless path. Both filters now apply on both sides of the mirror.

The same probe makes the script's entry self-configuring, so `afx spawn --resume` and a Tower-side terminal re-create need no second script shape — and a re-run never re-queues the task into a live conversation.

## ⚠️ Please look here first: two edits to shared, just-merged gate logic

`servers/render-gate.ts` is Spec 1313 code that landed recently. This PR touches its classifier in two places. Both are opt-in per profile, and **neither can change behaviour for claude, codex or agy** — argued below and pinned by tests.

### 1. The marker exemption follows the matched span instead of column 0

```diff
-      if (row === markerRow && col === 0) continue; // the marker glyph itself
+      if (col < markerEnd) continue; // the marker glyph itself
```

**Why:** kimi draws its composer inside a rounded box, so its prompt marker sits at **column 3** (`│ > `), not the row start. Under the column-0 rule the `>` glyph counts as user text, so a genuinely empty Kimi composer classifies `user-text` **forever** — i.e. holds all of its mail, permanently.

### 2. A profile may declare an upper bound for the composer region

`GateProfile` gains an optional `regionStartPatterns`. Kimi sets it to the box top; every other profile leaves it unset and keeps scanning from the marker row exactly as before.

**This one fixes a false CLEAN, and it was found by review, not by me.** `KIMI_MARKER` matches `│ >`, `findMarkerRow` takes the **last** match, and the scan started **at** that row. So a draft whose final line begins with `>` — a pasted quote, a markdown blockquote — puts the marker on the *continuation* row, leaving the real draft text **above** the scanned region. Measured on real kimi 0.34.0:

```
 ╭──────────────────────────────
 │ > implement the whole feature
 │   >
 ╰──────────────────────────────
```

→ classified `{clean: true, detail: "empty"}`. A queued message would then have been typed on top of unsent user input — the exact corruption Spec 1313 removes by construction. Committed as the `kimi-multiline-bare` fixture, captured rather than constructed.

The bound is **exclusive**, mirroring the region end, and that is load-bearing rather than stylistic: the box-top row's right corner `╮` is not in the classifier's ignorable-glyph set, so an inclusive bound counted it as user text and held every idle kimi composer forever. The fixture suite caught that on the first attempt.

**Why other profiles cannot reach it:** they declare no `regionStartPatterns`, so the region starts at `markerRow`; and since `findMarkerRow` returns the *last* match, no row below it can match either. The set of marker-matching rows in their region is exactly `{markerRow}` — the previous behaviour, by construction. A test asserts all three declare no region start, and that text above the composer still classifies clean for claude and codex.

**Why it is a no-op for every other app**, argued and then pinned by tests:

| profile | marker | span | effect |
|---|---|---|---|
| claude | `^[❯›]` | 1 | `col < 1` ≡ `col === 0` — literally the old rule |
| codex | `^[❯›]` | 1 | same |
| agy | `^> ` | 2 | extra cell is the space, already skipped by the whitespace rule that runs **before** the marker check |
| kimi | `^\s*│\s*>` | 4 | the case this exists for |

Guardrail tests (`render-gate.test.ts`, "marker-span exemption is a no-op…"): the exact span per shipped profile; a tightest-possible 1-char draft in the first cell the exemption could wrongly reach, per profile, all still `busy` (over-skipping is the only direction that could cause harm — a false CLEAN); proof that agy's span can never over-reach a typed character (`>x` doesn't match its marker at all); and a direct before/after demonstration that a span-2 kimi profile classifies the real 0.34.0 idle capture `user-text` while the shipped span-4 one classifies it clean. Every pre-existing claude/codex/agy fixture still passes unchanged.

## Undocumented-surface reliance (audited against kimi 0.34.0, 2026-08-09)

Two surfaces, both dated so the reliance can be re-checked on each Kimi major:

1. **Session store** `~/.kimi-code/sessions/wd_*/session_*/state.json`. This has **already drifted once**: 0.33.0 renamed `workDir` → `cwd`, moved timestamps from ISO strings to epoch ms, and dropped `lastPrompt`. Readers accept both shapes, and `codev doctor` asserts the load-bearing facts explicitly and **names** the one that broke rather than reporting "something changed".

2. **Workspace-trust record** `~/.kimi-code/workspace-trust/wd_<basename>_<sha256(root)[:12]>` → `{root, trustedAt}`. **This is the maintainer veto point, so here is the full argument.**

   kimi 0.33.0 added a startup *"Trust this folder?"* dialog. A builder worktree is always a brand-new directory; the dialog renders before any composer, and its only non-trusting option **exits kimi** — so an unattended builder would sit on it forever. The spawn path therefore pre-writes the trust record.

   - **No sanctioned bypass exists.** `kimi --help` has no flag; a full strings sweep of the 0.34.0 binary for `KIMI_*` env vars and for trust config keys found nothing (every "trust" hit was KaTeX/V8/OpenSSL noise). I looked for a supported knob first, as requested, and there isn't one.
   - **What trust actually gates is narrow**: whether project-level MCP servers (`.mcp.json`, `.kimi-code/mcp.json`) load from the folder. It does **not** gate tool execution or writes.
   - **Scope**: written only for a worktree Codev itself created, for a builder the human explicitly spawned, already running `--yolo`. It grants strictly less than launching the builder already authorized, and never touches a directory the user did not hand us.
   - **Fail-soft**: on any error the dialog simply appears, the gate holds the task message (no composer marker → `busy`), and mailbox escalation surfaces it. Never a silent misdelivery.
   - **Drift is detected, not silent**: `codev doctor` validates our derivation against kimi's **own** records (each record carries the `root` it was written for, so the expected filename is recomputable). A scheme change surfaces as a named warning instead of silently stranding every new builder on the dialog.

   **My reviewers split on this one, so you should see both sides rather than just my case.** Codex argued it is a real security boundary: `--yolo` governs approval of the agent's *tool calls*, whereas workspace trust governs whether repository-controlled MCP configuration is loaded and its processes started at all — so a builder spawned on a fork PR or other untrusted branch could have attacker-controlled MCP config loaded without a human ever seeing the decision. Claude reviewed the same code and reached the opposite conclusion: a `--yolo` builder in a Codev-created worktree already holds strictly more authority than MCP loading confers. I find Claude's reading more persuasive for the worktrees Codev creates, but Codex's fork-PR scenario is the case where the two arguments genuinely diverge, and that is a policy call I do not think is mine to settle.

   If you'd rather not ship the hash write at all, the fallback is that Kimi builders require one human keypress at first launch; say the word and I'll cut it. A narrower option, if you want the automation but not the blanket: refuse the pre-write when the worktree carries project-level MCP configuration.

## Corrections to claims this PR previously made

- **"Kimi has no hook seam, so #1018 write-guard parity is impossible" — obsolete.** Kimi documents blocking `PreToolUse` hooks (`[[hooks]]` in `config.toml`, exit code 2 blocks, 18 events as of 0.32.0). Parity is **achievable**. I have scoped it as follow-up rather than growing this PR further, but that is your call — say so and I'll add it here. Until it lands, a Kimi builder can write outside its worktree, and the docs now say exactly that.

### A. The version floor moves 0.27.0 → 0.33.0, which drops working installs

This is a real compatibility reduction and I want it visible rather than buried in a diff. The reasoning chain:

1. **`--agent-file` requires ≥ 0.31.0.** Below that the role does not inject *at all* and the builder runs silently roleless — the worst available failure mode, and not one a user would notice quickly.
2. **The folder-trust dialog appears at 0.33.0**, which is what the trust pre-write exists to handle. On 0.31–0.32 there is no dialog, so that machinery is inert — but it also means those versions are a genuinely *different* startup path from the one I exercised.
3. **0.33.0 is an engine boundary**: it made `agent-core-v2` the default. Every live measurement backing this PR — store shape, trust behaviour, the render-gate composer profile — was taken on 0.34.0, i.e. on that engine. 0.31–0.32 run the old engine and are **unmeasured**.
4. **Kimi ships weekly**, and its store has already renamed a load-bearing field once inside this PR's lifetime. Under that cadence a narrow, evidence-backed floor is safer than a wide compatibility claim I cannot stand behind.

So the floor sits at the oldest version the evidence actually covers, not the oldest that would nominally function. **If you would rather accept 0.31.0** (functional minimum, unmeasured) **or hold at 0.27.0** (maximum compat, definitely broken for role injection), say which and I'll change the one constant in `doctor.ts` plus the docs.

### B. The `.builder-kimi` marker is deleted — this is not a revert of your July fix

Your July REQUEST_CHANGES found a real bug: the **bare** launch shape (no role, no prompt) never persisted `.builder-kimi-session`, so an override-spawned bare Kimi builder (`--builder-cmd kimi` in a claude-configured workspace) fell through to claude's 80ms Enter and its mail was swallowed. That fix — `touch` the marker in the bare branch — was correct and shipped.

This PR removes the marker entirely, so I want to be explicit that the **property your finding protected is now stronger, not weaker**.

The marker was a *separate artifact* that every launch shape had to remember to write. That is a standing obligation, and the bare shape is precisely the one that forgot it — which is why your review caught a bug rather than a typo. Adding the missing `touch` fixed that instance; it did not remove the class. A future fifth launch shape could forget it again.

Pacing now reads the harness name out of the **generated `.builder-start.sh`**, which is generated *from* the resolved harness. There is nothing to remember: any shape that launches kimi necessarily names `kimi` in command position, because that *is* the launch. The obligation is discharged by construction rather than by discipline, and it is still override-proof for exactly your scenario — a `--builder-cmd kimi` spawn against a claude-configured workspace resolves `kimi`, because the script was generated from the override.

Tests keep your scenario pinned directly (`mailbox-pacing.test.ts`: "resolves kimi for the BARE launch shape too — the shape the old marker probe missed", plus an explicit override-proofness test), and `spawn-worktree.test.ts` pins that both generated shapes put `kimi` in command position, so a refactor that hid it would fail rather than silently degrade pacing.

## What changed since your last review

| Was | Now | Why |
|---|---|---|
| Seed bootstrap: `kimi -p` seed → capture `session.resume_hint` → pinned `kimi -S <id>` loop | Role via `--agent-file`; task via the mailbox | Drops 3 undocumented surfaces; role rides a **system prompt** instead of a user turn |
| `seed-kick.ts`: sentinel watcher + grace + `BEGIN` written straight to the PTY, verified via `state.json.lastPrompt` | **Deleted.** Spec 1313's render gate is the readiness barrier | A direct PTY write is exactly what Spec 1313 forbids; the gate already answers "is this composer empty?" |
| Resume: explicit `-S <discovered-id>` | `kimi -c` behind a fail-closed store probe | Documented flag; no undocumented id in generated bash |
| Pacing via a `.builder-kimi` marker file | Harness read from the generated `.builder-start.sh` | The marker obliged every launch shape to write one — **the bare shape didn't**, which was your finding last round. The launcher is generated *from* the resolved harness, so it cannot be forgotten or overridden away |
| `message-pacing.ts` + `seedKick` on `createTerminal` | `resolvePacingForSession` in `mailbox-wiring.ts`; `SeedKickRequest` removed from the SDK | Spec 1313 replaced the routes the old pacing hooked into — it was wired to nothing after the merge |

## What the 3-way review round found (and what it changed)

Before opening this for re-review I ran gemini, codex and claude over the post-merge delta, asking them to attack the shared gate edit hardest. **gemini APPROVE; codex and claude both REQUEST_CHANGES** — and they were right. Full dispositions are committed at `codev/projects/1201-*/1201-cmap-postpivot-dispositions.md`; the two blocking ones are worth stating here because they say something about where the risk in this PR actually lives:

- **The false CLEAN** described above. Claude reproduced it on a constructed screen and flagged honestly that it had no live kimi to confirm the real multi-row geometry. I measured it — kimi renders exactly that shape. Claude also proposed a second input (a marker row inside a second box *below* the composer); measured, that one is **not** reachable, because kimi's `/` menu renders as unclosed `│` rows with no `╰` beneath them, so anything inside it yields `no-region-end` → held. Both are now fixtures.
- **The store probe diverged from the TypeScript it mirrors**, and the cross-check test did not catch it because it compared two implementations of the *same* omissions. Codex found the dangerous direction: an `archived` session authorized `-c`, which kimi then refuses to continue, producing the silently-roleless session the guard exists to prevent. Claude found the safe-but-harmful direction: `readdirSync` on a stray non-directory threw `ENOTDIR` into the single **outer** try, so one `.DS_Store` in `~/.kimi-code/sessions/` disabled resume machine-wide, permanently and silently.

Also fixed from that round: shell metacharacters in a builder id or task path could execute when the launch script printed a recovery hint (all three reviewers, from different angles); a crash loop re-queued the same task every ~2s even though the mailbox persists held rows; and both drift probes reported healthy forever after a store migration, because "any record still matches" is satisfied by the pre-migration records.

**The common thread:** every one of these lives in a state a happy-path run does not produce — an empty composer and a clean store both behave correctly, which is precisely why three passing live demos missed them.

One more disclosure, since it affects how you should read the demo. Two demo steps were failing when I picked this back up, and the cause was the **demo**, not the product: its role told the model to prefix every reply with a token, which measures whether K3 honours a persistent output-format constraint rather than whether the role was injected. The `--agent-file` probe, run against a production-identical agent file, passed 7/7 including role survival across `kimi -c`. The demo now asks for a codeword instead — the same oracle the probe uses — and carries a comment explaining why, so the weaker check does not come back.

## Verification

- `pnpm build` clean; **full suite 4900 passed / 48 skipped / 0 failed**.
- **Live demo against real kimi 0.34.0, 7/7** (`codev/spikes/pir-1201-kimi-builder-demo.mjs`, runs the REAL dist modules): render gate classifies the live composer; `--agent-file` role honored in the interactive TUI; paced multi-line delivery submits; crash → store probe → `kimi -c` → role survives; probe fails closed on an empty store; trust pre-write idempotent.
- **Pivot validation, 7/7** (`pir-1201-kimi-agentfile-probe.mjs`): `--agent-file` injects in both `-p` and the interactive TUI; TUI start mints no session, the first message mints exactly one; `kimi -c` resumes with the role binding intact and mints no second session.
- **Gate fixtures are real 0.34.0 captures** (`pir-1201-kimi-gate-measure.mjs`), committed raw — they carry only throwaway `/tmp` paths. Idle → clean; draft, multi-line draft, the bare-`>` multi-line draft, `/` menu, `@` picker and the folder-trust dialog → busy (so a blind Enter can never confirm filesystem trust).

## Out of scope

Kimi as **architect** (stage 2); ACP / `kimi server` adapter; #1018 write-guard parity (now achievable — see above).


mohidmakhdoomi:

Contributor-side review summary for maintainers (this PR was developed under the PIR protocol on our fork; we do not merge — that call is yours).

**Process**: Plan and dev-approval were human-gated pre-PR. The dev-approval gate included a full-path live demo through a locally installed build — a real `afx spawn` of a Kimi builder through Tower showing (1) seed-session bootstrap with session-id capture, (2) sentinel-gated, store-verified BEGIN delivery, (3) multiline `afx send` at the pinned Enter delay, and (4) inner-restart context retention via `kimi -S` — plus `codev doctor`'s new kimi checks.

**Consultation (CMAP, single advisory pass)**: gemini APPROVE, claude APPROVE, codex REQUEST_CHANGES. The codex finding was real and was accepted + fixed in 732f04b8: seed-kick delivery confirmation was a substring match on `lastPrompt`, and the fresh-spawn seed prompt itself contains "BEGIN", so verification false-positived before the kick submitted — defeating the swallowed-Enter recovery. Fix is whitespace-normalized equality, with two pinning regression tests (both fail pre-fix), seed-kick suite 14/14, and a post-fix live demo re-run (5/5).

**Architect verification of the post-CMAP fix** (since PIR's single-pass consultation does not re-review fixes): I reviewed `confirmed()` in `seed-kick.ts` and the pinning tests directly — the equality predicate is correct, preserves the multiline-payload fallback (kimi flattens submitted newlines; normalization covers it), and the retry ladder (Enter re-send → one re-kick → loud warn) is intact. Known benign edge: a message raced into the session during the ~10s verify window causes one bounded re-kick, then a warning.

**Scope**: verified against issue #1201's builder-MVI checklist — no architect-parity changes (`tower-utils`/`tower-instances`/`tower-terminals`/`session-manager`/`architect.ts` untouched), no ACP adapter; kimi-as-architect fails loudly. Undocumented Kimi surfaces are labeled as observed, pinned to kimi >= 0.27.0 with a doctor smoke probe.

---
Architect integration review

waleedkadous (changes requested):
Outstanding work — this is one of the most disciplined external PRs we've received; the observed-vs-documented honesty around Kimi's store and the store-verified BEGIN ladder are exactly right. I'd like to see one change before merge: the bare launch path (no role, no prompt) doesn't persist `.builder-kimi-session`, so pacing resolution falls back to workspace config and an override-spawned bare Kimi builder gets Claude's Enter timing — the swallowed-Enter bug this PR fixes. Persisting a marker on every Kimi launch shape (plus a regression test for the override-spawn case, and softening the arch.md "exists iff Kimi-shaped" claim) closes it. I've added the `area/tower` label to #1201 for you.

mohidmakhdoomi:
## Architect Response — maintainer review addressed (2abd362a)

Thank you for the review — the finding is real and is fixed in 2abd362a.

**The gap, confirmed**: `KIMI_HARNESS.buildBuilderLaunchScript`'s bare branch (no role, no prompt — the `--builder-cmd kimi` override-spawn / `afx spawn --worktree` shape) returned a plain TUI loop that never wrote `.builder-kimi-session`. Seed and resume shapes both persist it, so pacing's marker-first probe held for them — but a bare builder fell through to the config-resolved harness and, in a claude-configured workspace, got the 80ms Enter that Kimi's paste detection swallows.

**Fix**: the bare branch now `touch`es the marker before the loop — empty, since there is no session id to pin. `touch` preserves a previously seeded id; an empty file keeps the seed `! -s` guard and `buildResume`'s empty-id fallthrough intact, so no other shape's semantics move.

**Regression tests** (all fail pre-fix where noted):
- `spawn-worktree.test.ts` — the override-spawn pin you asked for; notably this test previously asserted the marker's *absence* in the bare script, which is how the gap survived review. Flipped and failing pre-fix.
- `harness.test.ts` — same pin at the provider-script layer; failing pre-fix.
- `message-pacing.test.ts` — real-fs test: an empty marker (exactly what `touch` produces) must resolve Kimi pacing over a claude workspace config. Passes pre-fix by design — it pins the probe as existence-based so a future content-based "improvement" can't silently re-open the bare shape.

**Docs**: arch.md's pacing paragraph and the `message-pacing.ts` header now carry the softened claim — every Kimi launch shape persists the marker (existence-based probe), and the converse is explicitly *not* claimed: a leftover marker is a launch breadcrumb, not proof of a live Kimi session (staleness cost: a ~1s-slower Enter). The literal "exists iff Kimi-shaped" phrasing traced to the builder's thread journal; corrected there via an appended dated entry rather than rewriting history.

Suites: harness 75/75, spawn-worktree 86/86, message-pacing 8/8, seed-kick 14/14; package build clean.

---
Architect integration review

## CMAP results (iterated to convergence)

The maintainer-response change set was reviewed by a 3-way CMAP loop (gemini, codex, claude), re-run after every change until a full pass returned zero findings: iteration 1 reviewed the fix (2abd362a), iteration 2 the resulting test-hardening commit (642b1726), and iteration 3 the entire cumulative response diff (47d12ba9..1de55e13) as a holistic convergence check.

| Model | Verdict | Findings across iterations |
|---|---|---|
| gemini | APPROVE ×3 | Iter 2 NIT (accepted): the *pre-existing* `exit 1`-before-loop assertion lacked a `toContain('exit 1')` guard — removing `exit 1` would vacuous-pass (`indexOf` → -1, and -1 < any positive index) → fixed in 1de55e13. Iters 1 and 3: none. |
| codex | APPROVE ×3 | Iter 1 MINOR (accepted): tests pinned the `touch`'s presence but not that it stays **before** the `while true` loop → fixed in 642b1726 (ordering assertions at both test layers). Iters 2 and 3: none. |
| claude | APPROVE ×3 | Iter 1 NIT (thread-log phrasing) — explicitly no-action. Iter 2 NIT (accepted): the new ordering assertions lacked `toContain('while true')` — sound, but opaque failure diagnostic → fixed in 1de55e13; every ordering assertion now guards both substrings. Iter 3: none — additionally verified the no-race property (the `touch` completes before Tower can deliver a first send) and that the marker claims in the `KIMI_SESSION_FILE` JSDoc, `message-pacing.ts` header, and arch.md are mutually consistent. |

**Loop terminated** at iteration 3: unanimous APPROVE, zero findings, on the full change set. Suites green throughout (harness 75, spawn-worktree 86, message-pacing 8, seed-kick 14); package build clean.

waleedkadous:
Heads-up for your next iteration: PR #1244 (in review) changes the builder launch-loop contract — a deliberate agent exit (exit 0, no signal; see `isDeliberateExit()` in `shellper-protocol.ts`) now ends the loop with a keypress-gated relaunch instead of auto-respawning. The Kimi provider-owned loops in this PR replicate the old blind `while true` shape, so once #1244 lands, please adopt the shared `LAUNCH_LOOP_TAIL` (or its exit-code branch) in the Kimi script paths too — happy to point at the exact spots if useful.

mohidmakhdoomi:
## #1244 launch-loop contract adopted (bfc8d62f)

Per the heads-up above: merged `origin/main` (brings PR #1244) and adopted the shared `LAUNCH_LOOP_TAIL` in the Kimi provider-owned launch scripts.

**Change**: `LAUNCH_LOOP_TAIL` was module-local to `spawn-worktree.ts`; since `spawn-worktree.ts` already imports from `utils/harness.ts`, the constant moved there (exported, JSDoc noting why it lives in the harness module) — no circular import, and both Kimi loops (the pinned `-S` loop shared by the fresh/resume shapes, and the bare TUI loop) now interpolate it. A deliberate `exit 0` gates the relaunch on a keypress; nonzero/signal exits keep the auto-restart.

**Tests**: new `it.each` pins in `harness.test.ts` (fresh/resume/bare provider shapes) and `spawn-worktree.test.ts` (seeded + bare-override worktree shapes) assert `status=$?`, the exit-0 branch, the keypress gate, and the untouched crash path. Suites: harness + spawn-worktree 169/169, message-pacing + seed-kick 22/22, full package suite 3802 passed / 48 skipped; build clean.

**Live verification** (kimi 0.29.1, tmux PTY, the *actual* bare launch script generated from built `dist`):
- `/quit` in the TUI → exit 0 → `Agent exited at your request. Press Enter to relaunch…` — gate held with no respawn (observed >10s); pressing Enter relaunched the TUI via `continue`.
- `SIGKILL` on the `kimi-code` process → `Agent exited (code 137). Restarting in 2 seconds…` → auto-restarted, TUI back up.

**CMAP results** (gemini, codex, claude — parallel review of the merge-resolution + adoption change set; clean in one iteration):

| Model | Verdict | Findings |
|---|---|---|
| gemini | APPROVE | none — confirmed `status=$?` placement, `read -r \|\| exit 0` EOF semantics, all three Kimi shapes converted, no import cycle |
| codex | APPROVE | none — no remaining blind-restart Kimi loop; test coverage adequate across generic + provider shapes |
| claude | APPROVE | none — grep-verified the only `sleep 2`/restart echo in production code is inside `LAUNCH_LOOP_TAIL`'s crash branch, across all 8 interpolation sites; template-literal `$status`/`$SID` interpolation safe |

Ready for merge whenever you are.

waleedkadous:
Heads-up on the exit-handling contract I asked this PR to adopt — it just evolved, and I want you building against the current state rather than the deprecated one.

- PR #1244's `isDeliberateExit` semantics (clean exit → session ends dead) turned out to be wrong for architect terminals and were replaced in PR #1266 (issue #1264): for **architect** sessions, any clean exit now keeps the shellper/session alive and reruns the harness fresh (new conversation, no `--resume`); crashes still restart with recovery.
- **Builder** launch loops — which is what this PR's Kimi support runs as — are unchanged for now: clean exit still Enter-gates the relaunch per #1244. That half of the new spec is tracked in #1267 and hasn't been designed yet.

Practical upshot for this PR: mirror the current builder bash-loop behavior (Enter-gated relaunch on clean exit, auto-restart on crash) rather than the architect-side contract, and keep an eye on #1267 — when it lands, Kimi loops should move with it. Happy to review again whenever you've iterated!

mohidmakhdoomi:
Will wait for #1267 to be completed before updating this PR. Also once #1330 is merged to main, will need to update this PR to have Kimi work with the new `afx send` mailbox delivery.

mohidmakhdoomi:
Both blockers this PR was parked on have landed, so we're resuming work now. The branch currently conflicts with `main`; builder `pir-1201` is being resumed to bring it current. What changed upstream since the last update:

- **#1267 closed via #1317** (2026-08-01): a builder's **clean-exit** relaunch now reruns the harness **fresh** (new conversation, no `--resume`), matching the architect semantics from #1264. This supersedes the July guidance to mirror the #1244 Enter-gated contract, which this branch adopted on 2026-07-25.
- **#1356 (PIR #1233**, merged 2026-08-08): a builder's **crash**-restart now resumes the pinned session, driven by new `session` script-fragment capabilities on the harness interface (`newSessionScriptFragment` / `resumeScriptFragment` / `sessionLooksResumable`) and the shared `pinnedFresh` / `resume` loop forms in `spawn-worktree.ts`.
- **#1330 (Spec 1313**, merged 2026-08-08): `afx send` is now **mailbox-first** — messages persist to `global.db` first and deliver only onto a render-gate-verified empty prompt. Message writers never write a PTY directly and never force-inject; the response is `delivered` | `held`+reason.

Re-integration plan:

1. **Merge `main` into `builder/pir-1201`** (merge, not rebase). Nine files conflict: `codev/resources/arch.md`, `codev{,-skeleton}/resources/commands/agent-farm.md`, `discover-resume-session.test.ts`, `commands/spawn-worktree.ts`, `lib/tower-client.ts`, `servers/tower-cron.ts`, `servers/tower-routes.ts`, `commands/doctor.ts`.
2. **Add a measured Kimi render-gate profile.** Under Spec 1313, `gate-profiles.ts` delivers only to measured harnesses (claude, codex, agy); an unknown harness holds every message with `no-profile`. Without a Kimi composer profile, every `afx send` to a Kimi builder is held indefinitely — this is now a functional prerequisite for the feature, not polish. Measure Kimi's composer (marker, placeholder bounds) per the agy Phase-3 net-new-measurement precedent and register it keyed off `detectHarnessFromCommand`.
3. **Re-home the pacing hooks.** The send-path integration this PR added in `tower-routes.ts` / `tower-cron.ts` was rewritten by the mailbox pipeline (`mailbox-delivery.ts` / `cron-delivery.ts` / `message-write.ts`). The `.builder-kimi-session` marker-probe resolution and the Kimi Enter delay need to plug into the new delivery path.
4. **Reconcile the seed-kick with the no-force-inject invariant.** The sentinel + store-verified BEGIN kick writes the PTY directly, which the mailbox contract now forbids for message writers. Redesign it to route through the mailbox + render gate (the gate's empty-prompt verification likely subsumes part of the retry ladder); keep only the store verification that still earns its place, and document the decision.
5. **Align the Kimi launch loop with the current exit contract**: clean exit → fresh conversation (for Kimi that means a fresh seed bootstrap, since ids are seed-captured rather than minted upfront the way `--session-id` allows); crash → resume the pinned id (`-S`), mapped onto the #1233 `session`-fragment machinery where it fits, provider-owned where Kimi's id-capture difference requires it.
6. **`SeedKickRequest` placement**: the wire type this PR added to `packages/core/src/tower-client.ts` must land per the #1189 server/client split (types package), keeping the boundary tests green.
7. Then the usual bar: full build + test, re-run the live demo driver against real kimi, CMAP on the delta, refresh the PR description (its #1244-era loop language is stale), push.

Will report back here once the branch is green against current `main`.

mohidmakhdoomi:
**Kimi drift audit — the CLI moved from 0.27.0/0.29.1 (what this PR was built and live-verified against) to 0.34.0 (releases are ~weekly; 0.34.0 shipped 2026-08-06).** The pivotal change is 0.33.0 switching all CLI surfaces to the **agent-core-v2 engine by default** (`KIMI_CODE_LEGACY_FLAG=1` is the escape hatch). Findings below are verified against a real 0.34.0 install and store, not just release notes.

**Breaking — must fix during re-integration:**

1. **`state.json` renamed `workDir` → `cwd`** (verified in the live store; new sessions carry `version: 2`). `kimi-session-discovery.ts` hard-filters on `workDir` (`typeof parsed.workDir !== 'string' → null`), so on ≥0.33 every session parse returns null — seed id-capture, ownership verification, `--resume`, and the store-verified BEGIN all silently die. Fix: accept both fields (`workDir ?? cwd`), update the doctor drift probe to pin the current shape, and revisit the minVersion floor.
2. **Session ids are now `session_<uuid>`-prefixed** (verified: the `id` field in state.json and the `-p` resume hint both print the prefixed form; the hint also shows an undocumented `-r` shorthand). Verify what `-S` accepts on 0.34.0 and keep discovery's returned ids consistent with it.
3. **`lastPrompt` is absent from a fresh 0.34.0 `-p` session's state.json** (verified). The seed-kick verifier reads it to confirm BEGIN submitted. Re-check whether the v2 TUI still writes it on submit; if not, the verifier needs a different signal (`lastTurnReason` now exists; `updatedAt` remains deliberately untrusted). Reconcile with re-integration plan item 4 — if the BEGIN kick moves onto the mailbox + render gate, the gate's own confirmation may retire this verifier rather than it being rebuilt.
4. **0.33.0 added a folder-trust prompt at TUI startup.** Verified: `kimi -p` runs untrusted without prompting (seed path safe, no trust record created), but the pinned `kimi -S <id> --yolo` TUI loop in a fresh builder worktree will hit an interactive trust dialog before the composer exists — the BEGIN delivery would hang there. Trust records live at `~/.kimi-code/workspace-trust/wd_<basename>_<12hex>` as `{root, trustedAt}` (hash scheme undocumented). Find the sanctioned bypass (flag, config, or pre-trusting the worktree) and cover a fresh-untrusted worktree in the demo driver.

**Stale maintainer-visible claims to update:**

5. **Kimi now has documented hooks with a blocking `PreToolUse`** (exit code 2 blocks; `[[hooks]]` in config.toml; 18 events as of 0.32.0). The PR's "no documented hook seam — #1018 write-guard parity impossible" claim (PR description, arch.md, config docs) is obsolete. Recommend updating the claims now and proposing in this PR whether write-guard parity lands here or as a follow-up issue — maintainer's call on scope.
6. **The core design rationale needs rewording, not reversing**: 0.31.0 added plugin-contributed `systemPrompt` (effective on TUI and `-p`) and Markdown agent definitions via `--agent-file` (new-session only — cannot combine with `-S`/`--continue`). The seed-session bootstrap remains the right shape (it delivers role *and* task and pins a resumable id, which `--agent-file` can't while resume needs `-S`), but "Kimi documents no system-prompt flag" is no longer true and should become "why seed still wins."

**Re-verify on 0.34.0:**

7. The 1000ms Enter-delay pin was bisected on the 0.27.0 TUI; the v2 engine and 0.31.1's redraw reduction warrant re-running the probe-10 bisect.
8. The new render-gate composer profile (re-integration item 2) must be measured on 0.34.0 — note 0.33.0's TUI also no longer creates a session at startup.
9. `kimi -p` now waits for background tasks/subagents before exiting (0.34.0) — the seed one-shot can legitimately run longer; check sentinel grace/timeout headroom.

Given the weekly release cadence, make doctor's drift smoke probe assert the load-bearing shape explicitly (field names, id format) so the next rename fails loudly instead of silently degrading to fresh-with-role.

mohidmakhdoomi:
**Design direction update — superseding item 6 of the drift audit ("reword why seed still wins"): the seed+BEGIN mechanism can likely be retired entirely.** Evidence, verified live on kimi 0.34.0:

- `kimi --agent-file role.md -p "…"` injects the role (probe token returned exactly), and the flag works for the **main** session in both `-p` and interactive mode.
- Resuming that session (`kimi -S session_<uuid> -p "…"` — the prefixed id form is what `-S` accepts) returned the role token again. The docs confirm: the bound agent is the session's identity, fixed at first bind and restored automatically on resume — no flag needed.

The seed bootstrap existed because 0.27.0 had no role flag and no TUI-positional prompt, forcing role+task into a `-p` one-shot with ack-and-wait discipline, a captured id, and a BEGIN kick to start work. Both halves now have sanctioned homes:

- **Role** → `--agent-file` at launch (body should compose with `${base_prompt}` to *extend* rather than replace kimi's system prompt — the `--append-system-prompt` analogue).
- **Task** → an ordinary first message enqueued on the Spec 1313 mailbox at spawn, delivered onto a render-gate-verified empty composer. This also degrades gracefully at the 0.33.0 trust dialog: the composer never renders while the dialog is up, so the message is simply held, not lost (pre-trusting the worktree is still needed for unattended spawns).

Deleted outright under this design: `seed-kick.ts` (sentinel watcher, store-verified BEGIN, retry ladder), the seed `-p` bootstrap and `.builder-seed.txt`, the ack-and-wait discipline — and with them the `lastPrompt` dependency that 0.34.0 broke anyway. The session-discovery module shrinks to crash-resume id pinning, and `-c/--continue` (documented, cwd-scoped) may replace even that — it also self-heals after a clean-exit fresh relaunch, where a pinned id goes stale.

Remaining kimi-specific work is unchanged from the earlier plan: render-gate composer profile, Enter pacing, trust handling, and the clean-exit-fresh contract (#1317) — with one open design point: unlike claude, kimi cannot re-pass the task positionally on a fresh relaunch, so the loop needs a small Tower-side signal to re-enqueue the task via the mailbox.

Builder `pir-1201` is being redirected to validate the TUI half (composer behavior under `--agent-file`, gate measurement, trust) and implement this pivot. Flagging here so it can be vetoed early if the seed approach is preferred.

mohidmakhdoomi:
Reporting back as promised: the branch is green against current `main` and ready for re-review at head `4a7e2afe`.

Architect verification (independent of the builder's report): merge conflicts resolved (GitHub reports MERGEABLE); full suite 4900 passed / 0 failed; live demo 7/7 against real kimi 0.34.0; the post-pivot 3-way consultation ran with the shared render-gate edit flagged for hardest scrutiny — gemini APPROVE, codex and claude REQUEST_CHANGES, both blocking findings verified real, fixed, and re-tested (dispositions committed in-repo and summarized in the description).

The description was rewritten from scratch since the seed-era narrative no longer described the design. Three decisions are explicitly left open for you in it: the version floor choice (0.33.0, rationale chain included), the trust pre-write (codex and claude split on it — both arguments presented), and whether #1018 write-guard parity via kimi's now-documented blocking PreToolUse hooks lands in this PR or as a follow-up.

waleedkadous (changes requested):
First, the thing that matters most: this sat un-re-reviewed for 26 days after you reported it green on Aug 9. That was our failure, not yours — you did the work, including a drift audit most contributors would have skipped, and the evidence discipline here (live demo against the real CLI, doctor drift-probes weighted by recency, honest documentation of the trust write) is exemplary. The 3-way integration review (gemini APPROVE; codex and claude REQUEST_CHANGES, both high confidence) confirms the design is right; the changes below are mostly a consequence of main moving underneath the branch while it waited.

**Findings**
1. **Workspace-trust pre-write (security, verified against the branch).** `ensureKimiWorkspaceTrust` writes `~/.kimi-code/workspace-trust/wd_…` at spawn. Your arch.md note argues it grants strictly less than launching the builder already did — but the one thing the record gates (loading project-level MCP servers from `.mcp.json` / `.kimi-code/mcp.json`) is repository-controlled content, so a branch that commits an MCP config gets its processes loaded without the human ever seeing Kimi's dialog. That's the #1328 class. Refuse the pre-write when the worktree carries project-level MCP config, and/or gate it behind an explicit config opt-in.
2. **Stale against three rebuilt seams.** 1,603 commits behind and CONFLICTING: the write edge is now `submitMessagePaced` + in-lock precheck (#1365), gate details are consolidated under `MailboxGateDetail`/`isUnverifiableVerdict` (#1482), and marker anchoring moved to cursor-row/palette (#1474). Two of this PR's edits produce real post-merge defects against them — its new gate details bypass the #1482 consolidation and re-fork `isClassifierStuck` locally, so a stuck Kimi hold would render as "a human at the line" with no escalation. `launchLoopTail` is already on main.
3. **Unmeasured under verified delivery.** The 7/7 demo predates #1573/#1584 (echo-verification before `delivered`; zero re-writes). Kimi's echo behaviour is exactly the kind of thing that produced the #1583 loop — it needs measuring, not assuming.
4. **The approved plan describes the retired design** (seed-session / PTY-kick / `-S`, Kimi 0.27). The shipped architecture (mailbox + `--agent-file` + guarded `-c` + trust) was never re-approved, so the human-approved artifact no longer matches the code.
5. No `PreToolUse` write guard for Kimi builders (#1018 class) — documented, accepted as a follow-up by all three lanes.

**What happens next — we'll do it, not you.** Given the delay was ours and the conflicts are semantic (the delivery-path edits have to be re-derived against the converged code, not merged), I'm opening a PIR re-plan lane that builds on this branch: merge main, re-derive the write-edge and gate-detail integration, add the MCP-config refusal, update the plan to the shipped architecture for re-approval, and re-run the live demo under verified delivery. Your commits and authorship stay intact; the lane adds on top. If you'd rather drive it yourself, say so and I'll hand it back. Thank you for the patience — and for the audit that made this tractable.

waleedkadous:
Re-plan lane opened: #1620 (PIR — plan re-approval and dev-approval gates, since the approved plan is stale and the trust change is security-relevant). It builds on this branch directly; nothing here is rebased or rewritten.

waleedkadous:
Hi @mohidmakhdoomi — a heads-up and a request.

We are executing the re-plan described in #1620 directly on this branch: merging current `main` in (no rebase, no squash, your commits and authorship stay as they are), re-deriving the delivery-path integration against the write-edge, hold-verdict and marker-anchoring changes that landed while this sat, adding the workspace-trust refusals, and rewriting the 1201 plan to describe the shipped architecture. The 26-day gap that made this necessary is on us, not you.

One thing we cannot do on our side: we have no authenticated Kimi CLI here, and the CLI is now at 0.41.0, seven minors past the 0.34.0 you measured. So when our commits land, could you review the delta and re-run the live demo (`codev/spikes/pir-1201-kimi-builder-demo.mjs`) plus the render-gate captures on the current CLI? We will post a short, exact checklist here at that point so it is one pass for you rather than an open-ended ask.

Thank you for the original work — the design held up to a full 3-way review; only the branch had drifted.

mohidmakhdoomi:
@waleedkadous although https://github.com/MoonshotAI/kimi-code/issues/2767 remains unfixed (so `--agent-file` and `--agent <custom_name>` do not work), a potential workable approach is to create a Kimi agent profile (e.g. .kimi-code/agents/codev-builder.md) in the builder worktree, with `name: agent + override: true + ${base_prompt}` in the body → role present in the system prompt, Kimi's own default prompt interpolated ahead of it. This new agent profile replaces the default profile Kimi falls back to while keeping the role scoped to the builder worktree. This is documented behaviour (kimi's agents guide names it explicitly), not a back door. Then use launch the builder without `--agent` or `--agent-file` because the `agent` profile is the default Kimi cli binds anyway.

amrmelsayed:
@mohidmakhdoomi — closing out the two things we left hanging, with apologies for the 14-day gap; both answers are below and neither requires anything from you until we confirm one more maintainer-side step.

## 1. Your agent-profile workaround: accepted, with constraints

The worktree-scoped `.kimi-code/agents/codev-builder.md` (`override: true`, `${base_prompt}` interpolated ahead of the role) is an acceptable role-delivery mechanism while MoonshotAI/kimi-code#2767 keeps `--agent-file` broken. You are right that it is documented behavior, and worktree scoping keeps it from contaminating anything global. Constraints for the implementation:

- Generate it at spawn behind the existing role-injection seam (`getWorktreeFiles` / `buildScriptRoleInjection`), like `.builder-start.sh` — never committed, so the harness provider decides file-vs-flag per CLI version and the rest of spawn stays unchanged.
- One security check to include: the spawn-time write must overwrite-or-refuse if a file already exists at that path in the checked-out repo. A repository carrying its own `.kimi-code/agents/codev-builder.md` must not be able to replace the builder's role — that would be a prompt-injection door opened by exactly the `override: true` semantics we are relying on.
- Keep `--agent-file` as the preferred path so the workaround retires cleanly when the upstream bug is fixed.

Since only your side has an authenticated Kimi CLI to iterate against, we would prefer you implement it (it is small and sits behind the seam above) — but if you'd rather we write it blind and you just test, say so and we will.

## 2. The promised re-verification checklist

**Gated on one maintainer-side step first**: the branch is currently conflicting with `main`, and the delivery-path code moved under it again (bracketed-paste long-frame writes and composer-region gating shipped in 3.3.4). We will merge current `main` into this branch — no rebase, no squash, same as before — resolve the conflicts on our side, and confirm here. Do not start the pass before that confirmation, or you would be verifying against a write edge that is about to change.

Once we confirm (and the profile workaround from §1 is on the branch, whoever writes it):

1. `git pull` the branch; record `kimi --version` (we expect 0.41.x) and that you are authenticated.
2. `pnpm install && pnpm build` (the demo drives the real `packages/codev/dist` modules, not sources).
3. Run `node codev/spikes/pir-1201-kimi-builder-demo.mjs` — all five stages should pass: role reaches the interactive TUI, render gate holds busy / passes idle, paced >3-line delivery submits, crash resume keeps the role, empty-store fail-closed launches fresh WITH the role. Paste the full output.
4. Run `node codev/spikes/pir-1201-kimi-gate-measure.mjs` — re-capture the gate-profile measurements on 0.41 screens. If classification drifted since 0.34, adjust the `KIMI_PROFILE` constants and include before/after captures in the commit.
5. Post the results here (version, both outputs, any profile diffs).

That is the whole pass. After it, an independent reviewer-seat 3-way review runs on the final head, and then this merges. Thank you again for the patience and for the workaround analysis — it was the right call on the right mechanism.

Issue #1194, opened 2026-07-18
Gate approvals advance builders silently: notify the spawning architect from the porch approve path


amrmelsayed:
## Problem

The sidebar's Approve Gate button (and the gate toast, palette, and bare `porch approve` in a terminal) writes porch state directly. The spawning architect is never told, so its mental model of the builder goes stale: builders advance phases without the architect's knowledge, and the architect's review at the *next* gate starts from confusion instead of context.

All approval surfaces funnel through one place today: `runPorchApprove` in `apps/vscode/src/commands/approve.ts` shells out to `porch approve <id> <gate> --a-human-explicitly-approved-this`, and terminal users invoke the same CLI.

## Design (v2 — post team review, decision by Amr 2026-07-18)

Treat this as an information problem, not an authority problem. Gates stay human gates; the architect is never inserted into the execution path. Emit a notification from inside `porch approve` immediately after the gate write succeeds, so every surface (sidebar, toast, palette, CLI, future mobile) inherits it with zero per-surface work.

### Settled scope

1. **All four gates (spec-approval, plan-approval, dev-approval, pr), uniformly.** No per-gate policy. Self-adjusting: in the architect-driven path, pre-approved spec/plan frontmatter makes porch skip those phases, so notifications only fire on paths where the architect genuinely didn't know. Already-approved gates return early, so no double-fire.
2. **Typed gate-event frame ships in v1** (decision: Amr). The notification carries machine-parseable metadata alongside the human-readable banner, typed in `@cluesmith/codev-types` (wire contract only). This is the first system-typed bus frame and the seed of the structured activity events web + mobile need. Draft shape (field set open for team feedback, see below):

   ```ts
   interface GateEventMeta {
     kind: 'gate-event';
     gate: 'spec-approval' | 'plan-approval' | 'dev-approval' | 'pr';
     action: 'approved';
     projectId: string;          // "1191"
     builderId: string;          // "builder-pir-1191"
     workspacePath: string;
     source?: 'sidebar' | 'toast' | 'cli';
     routedTo: string;           // architect name the event was delivered to
     fallback?: boolean;         // true when rerouted to main
     artifactPath?: string;      // e.g. "codev/plans/1191-....md"
     timestamp: string;
   }
   ```

[...]

Issue #1147, opened 2026-07-07
Mobile v0 scope lock: LAN PoC (inbox + feed + chat + question cards + gate approval)


amrmelsayed:
## Summary

Lock the v0 scope for Codev Mobile before any spike or implementation work starts. v0 is a **LAN-only PoC** that proves the 30-second interaction model against a local Tower, with zero cloud dependency. Everything cloud-shaped is explicitly deferred.

Full design context: `codev/research/mobile/` (interaction-model.md, feasibility-2026-07.md, decisions/q1-q7). Ground-truthed against the repo on 2026-07-07.

## v0: IN scope (LAN PoC)

- `apps/mobile`: Expo + React Native + Expo Router, dev builds only (no store distribution)
- Connects directly to a local Tower over LAN (`BRIDGE_MODE`), inheriting Tower's existing perimeter model. Acceptable for a developer PoC; stated here as unacceptable to ship
- **Unified needs-you inbox**: pending gates and questions across workspaces (decisions/q5)
- **Activity feed**: rendered from `GET /api/overview` snapshots + the existing `/ws/messages` bus. No new Tower event log required for v0
- **Chat**: `POST /api/send` outbound, message-bus subscription inbound, markdown rendering
- **`AskUserQuestion` native picker cards**, once the Tower-side detection plumbing exists (see prerequisite below)
- **Gate approve/reject** from the inbox, once the Tower-side endpoint exists (see prerequisite below)
- **Presence**: definitive states (gate-blocked, pending question) as explicit attention rows; output-silence as a soft idle badge only (decisions/q6)
- **Desktop handoff**: "Continue on Desktop" opens the LAN dashboard URL (decisions/q1)
- `packages/tower-sdk`: extracted framework-agnostic data layer shared with the web dashboard (audit says ~70-75% ports; three known knots: URL/workspace scoping, SSE singleton, auth/storage injection)

## v0: explicitly DEFERRED

- **Push notifications** (APNs/FCM): cloud-gated, Phase 2 (decisions/q4, q7)
- **Auth / per-device tokens**: cloud-gated, Phase 2 (decisions/q2). No credential of any kind ships on a phone in v0
- **Offline outbound queue and question expiry semantics**: designed (decisions/q3) but implemented with the cloud phase
- **`codev://` deep-link handoff**: external dependency (codev-ide repo); browser URL is the v0 and permanent fallback
[...]

PR #1206, opened 2026-07-19
research: AskUserQuestion detection design comparison (co-design input for main)

amrmelsayed:
The design comparison owed to main for interaction-model §8.1: how Tower learns an agent called AskUserQuestion. Ground-truthed 2026-07-19.

Compares three mechanisms: **(A) harness-side hooks** — recommended; Codev already generates a per-builder PreToolUse hook at spawn (the #1018 worktree write-guard), so the seam, harness abstraction, and precedent all exist; PreToolUse = question_pending with the structured payload from tool_input, PostToolUse = question_resolved. **(B) PTY parsing** — rejected on ground truth (interactive ANSI menu, no semantic markers, no fixtures, inverts the repo's explicit-emission idiom). **(C) transcript tailing** — viable fallback (JSONL transcripts exist and Codev already discovers the files) but couples Tower to an undocumented harness-internal format; kept as a possible architect-side bridge.

Recommends a harness-neutral Tower contract (the hook is just the first emitter), reusing #1194's BusEventMeta envelope, with persistence per decisions/q3.

Four open questions flagged for main: architect hook coverage (builders get spawn-written hooks; architects don't), response *injection* being a separate harder problem than detection (v0 may ship detect-and-notify with answer-on-desktop), transport (dedicated endpoint vs #1194 path), and porch's reserved-but-unused `awaiting_input` field.

Related: #1147 (prereq 1), #1194 (envelope reuse), #1189 (consumer seams).

### DIFF
--- a/codev/research/mobile/decisions/askuserquestion-detection.md
+++ b/codev/research/mobile/decisions/askuserquestion-detection.md
@@ -1,0 +1,52 @@
+# Design comparison — `AskUserQuestion` detection (interaction-model §8.1)
+
+**Status**: Co-design input for main, not a decision. This mechanism serves web, VS Code, and mobile equally; ownership sits with main per the scope-lock's prerequisite 1 (#1147) and anti-pattern 9 ("do not design Tower's structured-event plumbing mobile-shaped").
+**Date**: 2026-07-19
+**Ground truth**: every repo claim below verified 2026-07-19 (file references inline).
+
+## The problem, restated
+
+To render a native question card, a surface needs the *structured* `{question, options[], multiSelect}` payload when an agent calls `AskUserQuestion`, plus a resolution signal when it's answered. Tower today does zero PTY content inspection (`pty-session.ts` `onPtyData`: timestamp + ring-buffer + broadcast, nothing else), so this is new plumbing whatever mechanism we pick.
+
+## Mechanism A — harness-side hooks (recommended)
+
+Claude Code hooks run a command on tool events, receiving the tool payload as JSON on stdin. Two hooks cover the whole lifecycle:
+
+- **`PreToolUse` matching `AskUserQuestion`** fires the moment the agent asks — the hook reads the structured question from `tool_input` and POSTs it to Tower: this *is* the `question_pending` event, with zero parsing.
+- **`PostToolUse`** fires when the question resolves — `question_resolved`, carrying the selection from the tool response.
+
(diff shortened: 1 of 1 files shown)


amrmelsayed:
## Main architect co-design review: rulings on the four open questions

Verified the load-bearing claims against source before ruling: the `buildWorktreeGuardFiles` / `getWorktreeFiles` seam exists as described; architect launch (`buildRoleInjection`) writes no settings; `awaiting_input` appears only in `porch/types.ts`, with no setter anywhere in porch source. Mechanism analysis is sound. **A confirmed as the recommendation**: harness-neutral Tower contract, the Claude hook as merely the first emitter, C kept in the back pocket. B rejected for the reasons given.

### Q1 (architect coverage): v1 is builders-only

Builders are where an unanswered AskUserQuestion silently stalls a lane; that is the high-value detection target. Architects are human-attended sessions in the main checkout, and every settings-injection option there collides with sessions we don't own: spawn-writing `.claude/settings.local.json` into the main checkout means writing to the human's own file (unlike a fresh worktree, where the write-guard precedent owns the file it creates). So: ship v1 on the builder spawn seam only. If architect coverage proves needed, two candidate bridges, in preference order: (a) a launch-flag-scoped settings file via `claude --settings <generated-file-outside-checkout>` (the flag exists, confirmed via `--help` today, but verify its merge semantics empirically before designing around it), routed through HarnessProvider like everything else; (b) Mechanism C as a read-only bridge. Neither blocks v1.

### Q2 (response injection): hard split confirmed

Detection ships as detect-and-notify; the answer happens at the desk in v0. Remote answering is its own spike with its own gate. Beyond the fragility you list, note: AskUserQuestion always offers a free-text "Other" option (arrow-keys-plus-enter synthesis can't cover it), multiSelect changes the key sequence, and a desk human can race the remote answer with the TUI as the only arbiter, so Tower-side compare-and-set protects Tower's record but cannot protect the terminal. The spike must define stale-answer semantics before any injection ships. Please amend interaction-model §8.3 accordingly once this lands.

### Q3 (transport): split, as you lean

- **Notification** (`question_pending` / `question_resolved`): rides #1194's system-sender `/api/send` path as the second member of the `BusEventMeta` union. The envelope, sender-type exemption, and open-enum discipline all reuse.
- **Resolution lifecycle** (answer submission, compare-and-set, expiry): dedicated endpoint. It is state mutation, not messaging, and doesn't belong in the messages route.
- **Backfill/reconnect**: REST read of Tower-held pending state per your q3-offline design.

Sequencing consequence: #1194 lands first (it is spawn-ready, awaiting Amr's go), and this rides its envelope. Design the union member here but don't spawn implementation ahead of #1194.

### Q4 (`awaiting_input`): don't couple

Ground-truth refinement: the field is not merely reserved. It came from plan 0087 (porch timeout/termination) together with `awaiting_input_output` / `awaiting_input_hash`, which exist for a resume guard (hash-compare the output file to detect whether the human resolved the blocker). The setter was never implemented, but the intended semantics are porch's resume-guard state machine, not a question lifecycle. Populating it from Tower's question events would conflate two state machines and hand porch state to a non-porch writer (porch owns `status.yaml`; nothing else writes it). Ruling: question lifecycle lives in `global.db`, owned by Tower. If overview or the sidebar wants a "question-blocked" badge, it consumes Tower state. Whether the dead porch fields should be deleted is a MAINTAIN-cycle question, not this design's.

### Disposition

Docs-only, well ground-truthed, good to merge on Amr's word. After merge, the amendment to interaction-model §8.3 (Q2) is the one follow-up owed.

---
Main architect co-design review

Issue #1149, opened 2026-07-08
Tower recovery: Claude architect crash-loops on stale --resume with no user escape


amrmelsayed:
## Problem

After a Tower recovery event (machine crash, `tower start` after downtime, network partition), some architect terminals enter a crash-loop displaying repeated "Could not resume session ..." errors. Users cannot cancel or dismiss the errors — the loop continues until Tower is externally killed or the offending row is manually removed from `~/.agent-farm/global.db`.

Reported by users 2026-07-08.

## Root cause (from subagent investigation 2026-07-08)

Both reconcile paths pre-compute `restartOptions.args` from the stored session id:

- `packages/codev/src/agent-farm/servers/tower-terminals.ts:661-690` (Phase 1 startup reconcile)
- `packages/codev/src/agent-farm/servers/tower-terminals.ts:898-925` (on-the-fly reconnect)

These `args` are handed to `SessionManager.setupAutoRestart` (`packages/codev/src/terminal/session-manager.ts:845-891`), which replays `session.options.args` verbatim after `restartDelay` (2s), up to `maxRestarts` (50 — `tower-instances.ts:532`).

If `--resume <uuid>` is unresumable (jsonl deleted, corrupted, or foreign), Claude Code prints `Could not resume session ...` and exits immediately. The exit re-triggers auto-restart with the same args. `startRestartResetTimer` fires unconditionally after each `spawn()` call (not after a "healthy" signal), so a fast-crashing process prevents the counter from ever resetting — count climbs monotonically to 50 (~100 seconds of errors) before `removeDeadSession` finally halts the loop.

**No escape hatch exists** — no env var, no CLI flag, no Tower HTTP endpoint. User must kill Tower or hand-edit `global.db` to break out.

Load-bearing context: the comment at `packages/codev/src/agent-farm/utils/harness.ts:83-87` explicitly notes that codex/gemini omit session id to avoid this class of crash-loop. Claude has the vulnerability.

## Fix sketch

Introduce a "recent failure" detector in `setupAutoRestart` (`packages/codev/src/terminal/session-manager.ts:845-891`):

1. Track restart timestamps + exit statuses on the session object
2. If ≥ 3 restarts fail within 30 seconds AND the failing args include `--resume`, strip `--resume <uuid>` from `session.options.args` on the next attempt
3. Log once: `"resume session <uuid> unrecoverable; retrying without --resume"`
[...]

Issue #1145, opened 2026-07-06
codev adopt / launchInstance: main architect hijacks an unrelated Claude Code session instead of spawning fresh


amrmelsayed:
## Problem

When Codev is initialized in a project that already had a Claude Code session running (`claude` binary already alive somewhere on the machine), the main architect launched by `codev adopt` (or by Tower's `launchInstance` on first workspace touch) does not spawn a fresh session — it appears to **hijack / attach to an existing, unrelated Claude Code session** as its starting point.

Reproduction:

1. Open a project directory in VS Code (or any terminal). Do NOT initialize Codev in it.
2. In some *other* context (a different terminal, a different project's session), Claude Code has an active session (persisted state on disk, potentially a running process).
3. In the target project, run `codev adopt` (or trigger auto-adopt via VS Code opening the folder with `codev.autoStartTower`).
4. Tower's `launchInstance` fires, creates the main architect terminal.
5. Observe: the main architect's opening state is **not a fresh session**. It picks up prior conversation context / role state from an unrelated Claude Code session, effectively hijacking it.

## Impact

- Architect boots in a context inherited from a different project — protocols, mental model, and any prior "you are the main architect for X" instructions bleed in.
- New workspaces don't start clean. The workspace's `codev/state/main.md` (per-architect notebook) and porch project state are correctly scoped, but the *conversation state inside the Claude terminal* is contaminated.
- Undermines the trust that "add-architect / adopt → fresh, purpose-scoped architect."

## Root cause hypothesis

Two candidate sites in the launch path:

- `packages/codev/src/agent-farm/servers/tower-instances.ts:392` — `launchInstance`. Spawns the main architect PTY. Investigate whether the launch options passed to shellper / Claude include a `--resume`, `--continue`, or session-id argument that reaches for the most-recent Claude session on the machine.
- `packages/codev/src/agent-farm/utils/harness.ts` — the Claude harness config. Look for `buildResume` / `sessionId` behavior that defaults to "most recent" when no explicit id is provided.

Specifically: Claude Code CLI has a `--resume` flag that reattaches to the most-recent conversation on the machine by default. If the launch command builds a `--resume` argument without validating that the session is scoped to this workspace, it'll grab whichever conversation Claude last touched globally.

[...]

Issue #1220, opened 2026-07-22
/arch-init: architect auto-state-saving — save to codev/state/<name>.md at checkpoints, then suggest /clear (supersedes #1212)


waleedkadous:
# /arch-init: architect auto-state-saving — save to codev/state/<name>.md at checkpoints, then suggest /clear

Supersedes #1212 (closed wontfix — we're switching approach from "add a save command / standing instruction" to **architect auto-saving built into `/arch-init` itself**).

## Design

`/arch-init` currently reads state at session start and never writes. This change makes the skill instruct a complete context-lifecycle loop:

```
/arch-init (recover) → work → auto-save at checkpoint → "good time to /clear" → human /clears → /arch-init (recover) → …
```

The state file stops being crash insurance and becomes the architect's deliberate memory-management mechanism: a **curated** summary written at a moment the architect chooses, at a resumable boundary — better than auto-compaction because both the timing and the content are intentional.

**Save-then-suggest ordering is the core safety property.** The agent cannot clear its own context and must not decide unilaterally to lose it. The skill instructs: save first, then tell the human it's a good time to `/clear`. The irreversible action stays behind a human keystroke, and accepting the suggestion can never lose anything because the save already happened.

## Skill-text requirements (add to BOTH trees' arch-init SKILL.md)

1. **When to save** — at resumable boundaries, judged by the architect: a gate approval, a PR merge, a completed investigation, the end of a long tool-heavy stretch. NEVER mid-task: the state file must describe a point a fresh session can pick up from.
2. **Write format = read format** (symmetry with the Spec 1134 minimum contract): *rewrite* the current-state/open-loops section in place; *append* a short dated log entry. The recovery flow reads "banner + most recent dated section", so that is exactly what a save must leave behind.
3. **Compaction discipline** — the file must stay readable in one screen: prune stale dated sections when appending; the state file is a summary, not a transcript.
4. **The suggestion is advisory, never nagging**: e.g. "State saved to `codev/state/main.md` — good time to `/clear` if this session is feeling heavy." Suggest only right after a save.
5. **Content guardrails**: no secrets, no transcript dumps; open loops + resume instructions + current focus.

## Why in /arch-init rather than a separate /arch-save skill

Self-reinforcement: every recovery re-reads the skill and re-primes the save habit — the durability gap named in #1212 (recovery assumes a file nothing instructs anyone to write) closes automatically after each architect's first `/arch-init`.

## Deliverables

[...]

PR #1225, opened 2026-07-22
[Bugfix #1224] Skip architect session resume when a live process holds the id

### LINKED ISSUE #1224: Tower/shellper: architect session-ID collisions + non-recovering crash loops after workspace restart
Reported by an external adopter workspace (4 incidents in one session, 2026-07-22), after a full workspace restart: Tower restarted, 8 architects re-added via `afx workspace add-architect`. Details scrubbed; adopter can supply shellper PIDs/timestamps on request.

## Symptom A — session-ID collision crash loop

An architect terminal crash-loops with:

```
Error: Session ID <uuid> is already in use
```

The shellper restarts its claude child forever; every child dies instantly. Two distinct holder cases observed:

1. **Stale pre-restart shellper** for the *same architect name* still held the session ID (survivor from before the Tower restart).
2. **Unrelated foreground claude process** on the user's own tty (started with wrong auth) had claimed the session ID.

## Symptom B — wedged shellper / registry divergence

- After the holder is killed, the shellper does **not** recover — children keep dying within seconds even with the session free.
- The architect **silently deregisters from Tower** while its shellper process stays alive crash-looping. `afx workspace remove-architect` then fails with `not found` while the zombie process persists.
- A third architect silently vanished from the registry with **no crash loop observed at all**.

## Recovery that worked (both collision cases)

1. Kill the wedged/stale shellper PID manually.
2. `afx workspace add-architect --name <name>` fresh.
3. Re-run `/arch-init`.

No state lost (state files + builders unaffected) — but the procedure is entirely manual and non-obvious.

## Suggested fixes (from the reporter)

- **(a) Mint-or-reclaim session IDs atomically on add-architect** — never reuse a persisted session ID without verifying no live holder; or always mint fresh.
- **(b) Crash-loop breaker in shellper** — detect repeated instant-exit of the child (N failures in M seconds), give up loudly, and deregister cleanly instead of looping forever.
- **(c) Make deregistration and process death atomic** — today both divergence directions occur: a registry row without a live process, and a live process without a registry row.

## Notes

- Related prior art: the verified-orphan shellper cleanup procedure (#1007) exists precisely because shellper processes can outlive their registry rows; this report shows the inverse direction too, plus the session-ID reuse hazard that turns a stale survivor into an active denial of the architect slot.

waleedkadous:
## Summary

After a workspace restart, re-adding architects via `afx workspace add-architect` could put an architect terminal into a permanent crash loop:

```
Error: Session ID <uuid> is already in use
```

The shellper restarted its `claude` child forever, every child dying instantly. This fixes the root cause: an architect launch would resume a persisted conversation session id **without checking whether a live process already held it**.

Fixes #1224.

## Root Cause

`resolveArchitectLaunch` (`packages/codev/src/agent-farm/servers/tower-utils.ts`) decides whether to resume a persisted architect conversation. Since #1145 it resumes a stored session id whenever the session jsonl **exists on disk** (`sessionIsOwned` → `verifySessionOwnership`).

That existence check answers *"does the transcript exist?"* but not *"is someone using it right now?"* — and a **held** session's jsonl exists *precisely because the holder is writing to it*, so the #1145 guard is guaranteed to pass in exactly the collision case. Two live holders were reported:

1. A **stale pre-restart shellper**'s `claude` child for the same architect name (shellpers survive Tower restarts by design).
2. An **unrelated foreground `claude`** the user started by hand.

In both cases the launch bakes `claude --resume <id>`, claude sees the id already locked and exits with *"Session ID is already in use"*, and the shellper auto-restarts into the same collision forever. (This is the reporter's suggested fix **(a)**: never reuse a persisted session id without verifying no live holder.)

## Fix

- **`sessionHasLiveHolder(sessionId, {list?})`** — scans `ps -A -o args=` for a live process launched with `--session-id <id>` or `--resume <id>` (both space- and `=`-joined forms). The match is **flag-anchored**, not a bare substring: a session id is short in tests and a raw substring coincided with unrelated paths in the process table; the flags are exactly how a holder is spawned. On any scan failure (`ps` missing/timeout) it returns `false` — the guard is **purely additive**, diverting to a fresh session only on *positive* evidence of a holder, and never making an un-held resume worse than before.
- **`resolveArchitectLaunch`** resume branch now requires `sessionIsOwned(...) && !hasLiveHolder(...)`. A held id mints a **fresh** session (with a `WARN` log) instead of baking a colliding `--resume`. Ownership is still checked first, so the `ps` scan is skipped for stale ids. `hasLiveHolder` and `log` are injectable seams.
- **`resolveArchitectRestart`** threads the logger, so the shellper restart-bake path inherits the guard for free.
- **`tower-instances.ts`** passes `log: _deps.log` at the `addArchitect` and `launchInstance` main-path call sites, so the divert-to-fresh decision is diagnosable in Tower logs.

This is **observation-only** — it never touches the holder. That is deliberate: holder case 2 is the user's own foreground terminal, so killing-to-reclaim would be wrong. Minting fresh is universally safe for both cases, and the architect ends up with a working conversation instead of a crash loop.

### Scope note (what this PR intentionally does *not* do)

The issue also lists suggested fixes (b) and (c):
- **(b) crash-loop breaker in the shellper** already largely exists — #1149's `maybeApplyCrashLoopFallback` (3 failing exits in 30s → swap to a fresh launch). This PR removes the collision at its source so that backstop isn't reached in the reported path.
- **(c) atomic deregistration ↔ process death** (Symptom B: a live shellper without a registry row, or a row without a process) is a separate, more architectural concern. Recommend a follow-up issue rather than expanding this BUGFIX.

## Test Plan

New cases in `tower-utils.test.ts`:
- **Held owned id → mint fresh, not `--resume`** (asserts no collision baked + a `WARN` log). Fails without the fix.
- No-holder owned id → still resumes as before (no regression).
- Ownership short-circuits the scan for a stale id (holder probe not invoked).
- `sessionHasLiveHolder` units: `--session-id`/`--resume` holders match, an unrelated id doesn't, and a scan failure returns `false` (never throws).

Validation: `porch check` — build ✓, tests ✓ (full suite). `tsc --noEmit` clean.

### DIFF
--- a/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
+++ b/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1224
 title: tower-shellper-architect-sessi
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-22T21:01:48.295Z'
-updated_at: '2026-07-22T21:06:31.211Z'
+updated_at: '2026-07-22T21:24:27.488Z'
--- a/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
+++ b/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1224
+title: tower-shellper-architect-sessi
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-07-22T21:01:48.295Z'
+updated_at: '2026-07-22T21:01:48.295Z'
--- a/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
+++ b/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-07-22T21:29:44.736Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-22T21:01:48.295Z'
-updated_at: '2026-07-22T21:24:27.488Z'
+updated_at: '2026-07-22T21:29:44.736Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
+++ b/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1224
 title: tower-shellper-architect-sessi
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-22T21:01:48.295Z'
-updated_at: '2026-07-22T21:01:48.295Z'
+updated_at: '2026-07-22T21:06:31.211Z'
--- a/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
+++ b/codev/projects/bugfix-1224-tower-shellper-architect-sessi/status.yaml
@@ -6,11 +6,12 @@ plan_phases: []
 current_plan_phase: null
 gates:
   pr:
-    status: pending
+    status: approved
     requested_at: '2026-07-22T21:29:44.736Z'
+    approved_at: '2026-07-22T22:25:19.645Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-22T21:01:48.295Z'
-updated_at: '2026-07-22T21:29:44.736Z'
-pr_ready_for_human: true
+updated_at: '2026-07-22T22:25:19.646Z'
+pr_ready_for_human: false
--- a/codev/state/bugfix-1224_thread.md
+++ b/codev/state/bugfix-1224_thread.md
@@ -63,3 +63,16 @@ short-circuits the scan, and 4 `sessionHasLiveHolder` unit cases incl. scan-fail
   before the full build shows 8 pre-existing env failures (missing `dist/` + copied skeleton
   artifacts — adopt/update/consult/tier-materialization/consolidate/session-manager integration);
   all clear once porch's build check emits those artifacts. None touch changed code.
+
+## PR
+
+PR #1225 opened (`Fixes #1224`), mergeable. CMAP:
+- Codex: APPROVE (HIGH, no issues)
+- Claude: APPROVE (HIGH, no issues) — confirmed TOCTOU windows are safe (mint-fresh or #1149 backstop)
+- Gemini: skipped non-blocking (couldn't emit a `--type pr` VERDICT in this worktree; known lane limitation)
+
+No REQUEST_CHANGES. Requested the `pr` gate via `porch done`; awaiting human approval before merge.
+
+Note for follow-up: consult's project auto-detect fails from a builder worktree that carries the
+full `codev/projects/` tree ("Multiple projects found"); had to pin `--issue 1224 --project-id
+bugfix-1224`. Worth a separate issue if it recurs.
--- a/codev/state/bugfix-1224_thread.md
+++ b/codev/state/bugfix-1224_thread.md
@@ -1,0 +1,65 @@
+# bugfix-1224 — Tower/shellper architect session-ID collisions + non-recovering crash loops
+
+Protocol: BUGFIX (strict). Issue #1224.
+
+## Investigate
+
+### Symptom A root cause (the headline: session-ID collision crash loop)
+
+`resolveArchitectLaunch` (`packages/codev/src/agent-farm/servers/tower-utils.ts`) decides
+whether to resume a persisted architect conversation. Since #1145 it resumes a stored
+session id whenever the jsonl **exists on disk** (`sessionIsOwned` → `verifySessionOwnership`).
+
+It never checks whether a **live process is already holding** that session id. When one is
+(the report's two cases: a stale pre-restart shellper's claude child, or an unrelated
+foreground claude), it bakes `claude --resume <id>` anyway → claude dies instantly with
+`Error: Session ID <uuid> is already in use` → shellper auto-restarts → dies again → crash loop.
+
+The existence check answers "does the transcript exist?" but not "is someone using it right now?"
+— and a *held* session's jsonl exists precisely because the holder is writing to it, so the
+#1145 guard is guaranteed to pass in exactly the collision case.
+
+### Fix (reporter's suggestion (a)): verify no live holder before resuming
+
+Before resuming a stored id, positively confirm **no live process holds it** (scan the process
+table for the id in argv — the same observable technique as the #1007 orphan-shellper cleanup).
+If a live holder is found, mint a fresh session instead of colliding. This is universally safe
+for both holder cases: we never touch the holder (critical — case 2's holder is the user's own
+foreground claude, must not be killed), and the new architect gets a working fresh conversation.
+The crash loop never starts.
+
+### Scope decision
+
+- Fix (a) is the isolated root-cause fix for Symptom A. Implementing it.
+- Fix (b) "crash-loop breaker in shellper" already largely exists (#1149
+  `maybeApplyCrashLoopFallback`, 3 failing exits in 30s → swap to fresh). Noting in PR.
+- Fix (c) "atomic deregistration/process death" (Symptom B registry divergence) is a separate,
+  more architectural concern — will recommend a follow-up issue rather than expand scope here.
+
+## Fix
+
+`packages/codev/src/agent-farm/servers/tower-utils.ts`:
+- New `sessionHasLiveHolder(sessionId, {list?})` — scans `ps -A -o args=` for a live process
+  launched with `--session-id <id>` / `--resume <id>` (both space- and `=`-joined). Match is
+  **flag-anchored**, not a bare substring — a bare substring false-positived on the short synthetic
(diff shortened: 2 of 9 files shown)


waleedkadous:
## CMAP review results

|---|---|---|---|
| **Codex** (gpt-5.4-codex) | ✅ APPROVE | HIGH | None |
| **Claude** (opus) | ✅ APPROVE | HIGH | None |
| **Gemini** (agy) | ⚪ skipped (non-blocking) | — | Lane could not produce a `--type pr` VERDICT in this worktree (known limitation); no findings |

Both completing reviewers approved with HIGH confidence and no requested changes. Claude specifically flagged the TOCTOU windows (holder dies/appears between check and launch) as safe — either mints fresh unnecessarily or falls back to the existing #1149 crash-loop backstop — and confirmed the guard is purely additive (scan failure → `false` → no regression).

waleedkadous:
Excellent root-cause work — the #1145-guard analysis ("a held session's jsonl exists *precisely because* the holder is writing it") is exactly right, and the fail-open, observation-only design is well judged. I'd like to see three changes before merge; the third is an approved scope expansion, not creep.

**1. Detect shellper parents too.** A remnant shellper's own argv carries the session id in JSON form — `"--session-id","<id>"` (quote-and-comma joined; verified against a live process table today). In a crash-looping remnant the claude child is dead most of the time (<8s alive per respawn), so the current space/`=` needles miss exactly the incident-1 holder class during most of its life. Adding the JSON-quoted forms of both flags to the needle set makes detection reliable instead of timing-dependent.

**2. Mint-or-**reclaim**, not mint-only.** When the live holder is verifiably *our own* superseded shellper for the same architect identity, abandoning the conversation is a worse outcome than necessary. In that case: kill the stale instance (whole process group, SIGTERM with SIGKILL escalation — the `PtySession.kill` pattern) and **resume** the session. Keep mint-fresh for every other holder — a foreign process may be the user's own terminal and must never be touched. The identification must be airtight before any kill: a `shellper-main.js` process whose config argv carries the **same session id** AND the **same workspace cwd** AND the **same architect name**, and which is not the instance being launched. If any of those can't be positively confirmed, fall through to mint-fresh. Please cover the never-kill-foreign branch in tests explicitly.

**3. Fix symptom B in this PR as well.** Two concrete legs:
- **Reconcile on add/launch**: `add-architect` (and launch) for a name that already has a live shellper for the same identity should reattach to or reap that instance rather than spawning a duplicate that fights it for the session — this is the lifecycle hole that manufactured both collisions.
- **Registry/process consistency**: a shellper that gives up (the #1149 crash-loop breaker) should deregister cleanly, and `remove-architect` should be able to clear a live-process-without-registry-row zombie instead of failing "not found" while the process loops on.

For the wedge-after-free (incident 3's key datapoint: children kept dying <8s with the session demonstrably free), please capture and log the dying child's stderr/exit reason as part of this work — if the actual cause surfaces and is separable, document it in the PR rather than chasing it blind.

Practical pointers: the full forensic timeline (PIDs, lstart times, session ids, the manual recovery that worked, plus today's live `main`-architect recurrence and its different recovery path) is in the comments on #1224. The reporting workspace has offered to run live diagnostics on the next occurrence if that helps.

waleedkadous:
## Iteration 2 — addressed the three review asks

Thanks for the detailed review. All three landed on this branch (commits `f28f3300`, `f7d13245`).

### 1. Detect shellper parents (JSON-argv needles)
`sessionIdNeedles` now also matches the shellper parent's config-JSON form `"--session-id","<id>"` / `"--resume","<id>"` (alongside the child's `--session-id <id>` / `--resume <id>` / `=` forms), and the scan uses `ps -ww` to avoid truncating the large config blob. This catches the incident-1 remnant *shellper* whose claude child is dead between crash-loop respawns.

### 2. Mint-or-**reclaim** (never touch foreign)
New `servers/architect-session-holder.ts`:
- `classifyArchitectSessionHolder` splits holders into **reclaimable-own** vs **foreign**. Reclaimable requires an airtight identity: a `shellper-main.js` process whose JSON config carries the **same session id** AND **same cwd** (symlink-tolerant) AND **same `CODEV_ARCHITECT_NAME`**, and isn't self. Anything else that holds the id — a bare `claude`, a shellper for a different identity, an unparseable match — is **foreign**.
- `reconcileArchitectSessionHolder`: foreign → `{foreignHolder:true}` (mint fresh, holder untouched); our own superseded shellper(s) → `reapShellpers` (process-group SIGTERM → poll → SIGKILL) then resume; none → resume. A foreign holder *alongside* an own one still refuses to reap (foreign wins).
- Wired into `addArchitect` and `launchInstance`'s `main` path (main recovers via `workspace start`, so that's main's reclaim site).

The **never-kill-foreign** branch is covered explicitly (`reap` asserted *not* called for a foreign holder, and for foreign-alongside-own).

### 3. Symptom B
- **remove-architect** now reaps a **live-process-without-a-registry-row zombie** (`findOwnArchitectShellpers` by identity) instead of returning `not found` while the process loops.
- **crash-loop give-up** (`SessionManager` maxRestarts): reaps the shellper husk (process-group SIGTERM) so give-up leaves no row-gone/process-alive zombie, and captures the dying child's exit code/signal + shellper stderr tail.

**On the wedge-after-free** (incident 3: children dying <8s with the session demonstrably free): claude's own error surfaces through the PTY data / terminal ring buffer, not the shellper's stderr, so give-up now logs the exit code/signal + shellper stderr *for diagnosis* and reaps the husk so the loop can't persist. The definitive cause of that specific datapoint is captured-for-diagnosis rather than claimed-fixed — flagging honestly per your "document rather than chase blind." The reporter's offer to run live diagnostics on the next occurrence would confirm it.

### Validation
Full build ✓ · full unit suite **3582 passed / 48 skipped / 0 failed** · `tsc --noEmit` clean. New tests in `architect-session-holder.test.ts` and a give-up husk-reap test in `session-manager.test.ts`.

Re-running CMAP now and will re-request the `pr` gate.

waleedkadous:
## CMAP iteration 2 + responses

|---|---|---|---|
| **Claude** (opus) | ✅ APPROVE | HIGH | None |
| **Codex** (gpt-5.4-codex) | 🔶 REQUEST_CHANGES | HIGH | reap SIGKILL race; PR artifacts |
| **Gemini** (agy) | ⚪ skipped (non-blocking) | — | lane can't emit a `--type pr` VERDICT here |

### Codex #1 — `reapShellpers` didn't confirm death after SIGKILL → **fixed** (commit `5c9b4cb9`)
Real bug. After escalating to SIGKILL the function returned immediately, so the reclaim caller could resume `claude --resume <id>` while the holder was still in the kernel's exit path and hadn't released the session lock — re-opening the very collision race the reap exists to close. Added a `killGraceMs` post-SIGKILL confirmation poll and a test asserting it keeps polling until death.

### Codex #2 — `status.yaml` / `thread.md` in the PR → **rebuttal** (deferring to you)
Both are there by design, not stray artifacts I introduced:
- `codev/projects/.../status.yaml` is committed by **porch's own `chore(porch)` commits** — builders must never touch it, and every prior merged bugfix (e.g. #1220) carries the identical `chore(porch)` commits on `main`. Removing it would mean reverting porch's state-machine commits.
- `codev/state/bugfix-1224_thread.md` is committed per the **builder-role contract** ("default disposition is COMMIT"; the thread ships to `main` as the cohort situational-awareness record, alongside `codev/reviews/`).

Happy to strip either if you'd prefer, but I didn't want to unilaterally revert porch commits or drop the thread.

### Self-caught fix (also in `5c9b4cb9`)
`resolveArchitectRestart` (the shellper restart-bake path) was self-detecting the very shellper being reconnected — its child's argv holds `--resume <id>`, so the live-holder check matched *itself* and would bake fresh restart args on every healthy reconnect, dropping conversation continuity on the next child crash. Fixed by disabling the holder check on that path (`hasLiveHolder: () => false`); collision-avoidance lives at the add-architect / main-launch reconcile layer plus the #1149 runtime fallback.

**Validation after fixes:** full build ✓ · full unit suite **3583 passed / 48 skipped / 0 failed** · `tsc --noEmit` clean.

Re-requesting the `pr` gate.

amrmelsayed:

High-risk band (Tower core, ~600 added lines) — reviewed the full production diff against source, on top of the PR's own two CMAP iterations (codex REQUEST_CHANGES → SIGKILL-confirmation race fixed with the `killGraceMs` poll; then codex APPROVE HIGH + claude APPROVE HIGH; gemini lane skipped, known limitation). My verdict: **APPROVE**.

What I verified independently:

- **The own-vs-foreign classification is airtight in the right direction.** `isOwnArchitectShellper` requires shellper-main.js marker + `CODEV_ARCHITECT_NAME` match + symlink-tolerant cwd match, and every unparseable/partial match degrades to *foreign* (never touched). Kill requires positive identity proof; scan failure degrades to today's behavior. The asymmetry (easy to be foreign, hard to be reclaimable) is exactly right for a code path that kills processes.
- **The restart-bake opt-out (`hasLiveHolder: () => false`) is correct and necessary** — at bake time the holder is this shellper's own child by construction; holder-checking there would fresh-mint on every healthy reconnect and destroy conversation continuity. Good catch; the #1149 fallback remains the runtime backstop on that path, as documented.
- **The give-up husk-reap in session-manager closes the Symptom-B divergence I flagged on the issue** (deregistered-but-alive shellper was previously unreapable by ANY path — killOrphanedShellpers skips live sockets, shellper-main waits forever for a SPAWN). Give-up now kills the process group before `removeDeadSession`, and `remove-architect` gains the identity-keyed zombie reap for husks that predate this fix. Both divergence directions now have an owner.
- **Mint-or-reclaim ordering is safe**: `reapShellpers` resolves only after confirmed death (incl. post-SIGKILL poll), and only then does `resolveArchitectLaunch` bake the resume.

Two advisory notes, neither blocking:

1. **Adoption-vs-reclaim ordering assumption.** `launchInstance`'s main-path reconcile runs inside the `!entry.architects.has('main')` branch, so a healthy surviving shellper is protected only because Tower-startup adoption registers it before a workspace-start can reach this code. That ordering holds today; if launch ever races reconciliation, the reclaim would kill a healthy adoptable shellper. Worth a one-line comment at the call site naming the dependency.
2. **Version-dependence of the collision itself** (from my issue comment): on claude 2.1.216 two interactive instances share a `--session-id` without error, so the instant-exit crash loop the incident showed is claude-version/mode-dependent. The fix stays correct regardless — reaping superseded own shellpers and refusing to share a held id are right even where claude tolerates the collision (two writers on one jsonl is its own corruption risk). Just don't encode the 'Session ID is already in use' string/exit behavior as a test contract.

---
Architect integration review

PR #1234, opened 2026-07-23
Codev process fleet drives macOS memory-pressure kills: shellper husk sweep + fleet observability

### LINKED ISSUE #1227: Codev process fleet drives macOS memory-pressure kills: stranded shellper husks accumulate per restart + no idle-fleet memory policy
## Evidence (2026-07-23, this machine — the recurring 'window terminated unexpectedly (killed, code 9)' report)

The VS Code window kills are **macOS jetsam (memory-pressure SIGKILL) with codev's process fleet as the dominant memory load**. Hard evidence, not inference:

1. **JetsamEvent-2026-07-22-164753.ips** (system diagnostic, same day as the first window-kill report): the top-resident-memory ranking at the event is dominated by claude CLI processes — 8+ processes at 400–580MB each, self-titled by version (`2.1.207`, `2.1.209`, `2.1.214`), i.e. long-lived survivors weeks old. Chrome (~580MB) and WindowServer round out the list. Under this pressure macOS SIGKILLs large renderers — the VS Code window (code 9) is the visible casualty.
2. **Live process census** (same machine, 2026-07-23): 41 shellper-main processes running, only 20 registered in Tower. 18 claude processes totaling ~6.3GB RSS.

## Finding 1 — stranded shellper husks accumulate in waves that date-match Tower restarts

19 of the 21 unregistered shellpers are **childless husks** (~34MB each, ~657MB total), in two clean age cohorts:

- 6 husks aged ~6d15h → stranded **2026-07-16**, the v3.2.3 release-day restarts
- 13 husks aged ~3d19–20h → stranded **2026-07-19**, the #1198 live-deploy-incident / local-install churn day

This is empirical confirmation of the #1224 comment finding: a deregistered-but-alive shellper is never reaped by ANY path (`killOrphanedShellpers` skips responsive sockets; `shellper-main` idles forever awaiting a SPAWN). **Every Tower restart/update strands a batch and nothing ever cleans them** — the leak grows monotonically with update cadence, which is why the symptom 'keeps recurring and correlates with updates'. PR #1225's give-up/remove-architect reaps prevent NEW husks in its covered paths but nothing reaps the EXISTING population or strays outside those paths.

## Finding 2 — the standing fleet itself is a multi-GB design reality with no policy

The tracked-and-legitimate fleet: ~13 architect claude processes across 5 active workspaces, resumed-alive since ~2026-07-13 (auto-restart + #832 resume keeps them permanent), at 175–583MB each ≈ **~4GB standing**, plus active builders (e.g. 5 shannon builders ≈ 2.1GB at census time). An always-on multi-workspace fleet of ~20 claude processes is inherently ~5–6GB; on a laptop sharing memory with Chrome/VS Code this guarantees chronic pressure. There is currently no idle policy at all.

## Proposed work

1. **Husk sweep (bug-level):** a reconciliation pass that reaps childless, unregistered shellper-main processes older than a grace period — as a Tower-startup step (extending `killOrphanedShellpers`'s policy: a responsive socket should not protect a process that is BOTH unregistered AND childless for N hours) and/or a `codev doctor`/`afx` maintenance command so operators can recover without hand-rolled `ps` diffs. Must respect multi-Tower-instance socketDir scoping.
2. **Idle-fleet memory policy (design-level, likely separate spec):** options include architect hibernation (kill the claude child of an idle architect after N hours, keep the shellper + stored session id for instant resume — the infrastructure for this mostly exists post-#832/#1224), per-workspace stop-on-idle, or at minimum surfacing fleet memory in `afx status`/dashboard so the cost is visible.
3. **Observability:** `afx status` (or doctor) should report total fleet RSS and flag unregistered shellpers — this census took manual `ps`/registry joining that the tooling should own.

## Relation to existing work

- #1224 / PR #1225: fixes husk-stranding in the give-up and remove-architect paths; this issue covers the existing population + paths outside those + the sweep.
- #1215 documented the window-kill as 'no causal link found, likely OS OOM' — correct on mechanism, but the census shows codev's own fleet is the primary load; this issue corrects that record.
- #1217 (web decoration DOM accumulation) remains a separate, minor renderer-side contributor.

amrmelsayed:
# PIR Review: Shellper Husk Sweep + Fleet Memory Observability

Fixes #1227

## Summary

macOS memory-pressure kills (jetsam) were traced to Codev's own process fleet: shellper wrapper processes whose PTY child already exited ("husks") were permanently unreapable, because the existing `killOrphanedShellpers()` unconditionally protects any shellper whose socket still responds — and a husk's socket always responds. This PR adds a second, stricter sweep (unregistered AND childless AND aged past a grace period) on three triggers (Tower startup, an hourly in-process timer, and an on-demand `afx tower sweep-husks` CLI command with dry-run/`--apply` semantics), plus fleet-wide RSS and unregistered-shellper-count observability in `/health` and `afx status`. Idle-architect hibernation (the issue's proposal item 2) was explicitly excluded per architect scoping and is flagged as a follow-up spec.

## Files Changed

- `packages/codev/src/agent-farm/servers/process-census.ts` (+44 / -0) — new
- `packages/codev/src/agent-farm/servers/shellper-husk-sweep.ts` (+173 / -0) — new
- `packages/codev/src/agent-farm/commands/tower-sweep-husks.ts` (+92 / -0) — new
- `packages/codev/src/agent-farm/servers/tower-server.ts` (+36 / -0)
- `packages/codev/src/agent-farm/servers/tower-routes.ts` (+137 / -1)
- `packages/codev/src/terminal/session-manager.ts` (+21 / -2)
- `packages/core/src/tower-client.ts` (+55 / -0)
- `packages/codev/src/agent-farm/lib/tower-client.ts` (+3 / -0)
- `packages/codev/src/agent-farm/cli.ts` (+16 / -0)
- `packages/codev/src/agent-farm/commands/status.ts` (+29 / -2)
- `packages/codev/src/agent-farm/__tests__/process-census.test.ts` (+79 / -0) — new
- `packages/codev/src/agent-farm/__tests__/shellper-husk-sweep.test.ts` (+190 / -0) — new
- `packages/codev/src/agent-farm/__tests__/shellper-husk-sweep.e2e.test.ts` (+177 / -0) — new
- `packages/codev/src/agent-farm/__tests__/tower-routes-husks.e2e.test.ts` (+142 / -0) — new
- `packages/codev/src/agent-farm/__tests__/tower-sweep-husks.test.ts` (+117 / -0) — new
- `packages/codev/src/agent-farm/__tests__/status-fleet-observability.test.ts` (+132 / -0) — new

16 files changed, 1438 insertions(+), 5 deletions(-).

## Commits

- `ff25c080` [PIR #1227] Plan draft
- `ce24667f` [PIR #1227] Plan revised
- `7c01cf97` [PIR #1227] Add process-census + shellper-husk-sweep predicate
- `6a622dbe` [PIR #1227] Wire startup + hourly husk-sweep triggers into Tower
- `f4a9ee56` [PIR #1227] Add husk preview/sweep routes + fleet observability in /health
- `af130187` [PIR #1227] Extend TowerClient with husk preview/sweep + fleet health fields
- `73895847` [PIR #1227] Add afx tower sweep-husks CLI command
- `d2af62fc` [PIR #1227] Surface fleet RSS + unregistered shellper count in afx status
- `69bb42f3` [PIR #1227] Correct stale give-up-reap comment re: husk sweep eligibility

## Test Results

- `porch check` (build + tests): ✓ both pass
- `tsc --noEmit` clean across `packages/core` and `packages/codev`
- Unit tests: 98+ passing across `process-census.test.ts`, `shellper-husk-sweep.test.ts`, `tower-sweep-husks.test.ts`, `status-fleet-observability.test.ts`, plus the full existing `session-manager.test.ts` regression suite (79 tests, unmodified behavior)
- E2E (real Tower process, real `ps`/signals, no mocks):
  - `shellper-husk-sweep.e2e.test.ts` — creates a genuine husk (deletes its `terminal_sessions` row, kills its PTY child directly), confirms the periodic timer reaps it on the next tick; separately confirms graceful shutdown clears the new interval without hanging
  - `tower-routes-husks.e2e.test.ts` — confirms `GET /api/shellpers/husks` lists a genuine husk without touching it, `/health` reports it under `unregisteredShellperCount`, and `POST /api/shellpers/husks/sweep` then actually reaps it
  - Existing `shellper-cleanup.e2e.test.ts` regression suite: unmodified, still passing
- Manual: `afx tower sweep-husks` (preview and `--apply`/`-y`), `afx status` (human + `--json`) reviewed at the `dev-approval` gate

## Architecture Updates

Added a "Husk sweep (Issue #1227)" paragraph to `codev/resources/arch.md`'s "Orphan Session Detection" section — the existing text there stated shellpers are "never killed" during the SQLite sweep, which is no longer the complete picture now that a second, stricter sweep exists for exactly the case that policy leaves permanently unreaped. Routed to the **cold** tier (`arch.md`), not `arch-critical.md`: this is subsystem-specific mechanism detail, not a cross-cutting fact that should change unrelated implementation decisions, and the hot tier's existing "Agent Farm Internals" map entry already points here.

## Lessons Learned Updates

Added one entry to `codev/resources/lessons-learned.md` under "Debugging and Root Cause Analysis" (cold tier — the section already holds this exact class of JS falsy-coercion footgun, e.g. the `[From 0107]` `body && body.name` entry): `parsed || fallbackDefault` silently discards a legitimate `0` override, since `0` is falsy in JS. I hit this for real — `SHELLPER_HUSK_GRACE_MS=0` (set by my own E2E test for an immediate-eligibility grace period) silently reverted to the 1-hour default, and only a real running-process assertion caught it, not a unit test mocking the parse. Fixed with `Number.isNaN(parsed) ? fallbackDefault : parsed` for both the grace-period and sweep-interval env vars (`tower-server.ts`, `shellper-husk-sweep.ts`'s `resolveHuskGraceMs`).

- **Design choice: a sibling function, not an edit to `killOrphanedShellpers`.** I deliberately did not modify the existing, well-tested `killOrphanedShellpers()` in place — its "responsive socket protects" behavior is still correct and needed for the reconnectable case; the new `sweepShellperHusks()` is strictly additive and covers only the gap. Rationale is documented in the plan's Risks & Alternatives section.
- **The `0 || default` bug** (see Lessons Learned above) — worth a close look at `resolveHuskGraceMs()` in `shellper-husk-sweep.ts` and the parallel fix in `tower-server.ts`'s `huskSweepIntervalMs` parsing, since this class of bug is easy to reintroduce.
- **Codex review (iteration 1) found a real defect, fixed in `3a5083e3`: `process-census.ts` originally used `execFileSync`, called synchronously from the `/health` HTTP handler — blocking Tower's entire event loop (all open terminals' WebSocket traffic) for the duration of every `ps` call.** This is a previously-fixed anti-pattern in this exact codebase (`lessons-learned.md:160`), reintroduced by this PR. Rewrote `listProcessCensus()` to use non-blocking `execFile` (matching `session-manager.ts`'s existing async convention), added a regression test (`process-census.test.ts`: `'never calls the synchronous execFileSync API'`) that fails against the pre-fix code. Full detail and a second, related fix (a redundant double `ps` scan in `handleHuskPreview`, also flagged independently by claude's own review) in `codev/projects/1227-codev-process-fleet-drives-mac/1227-review-iter1-rebuttals.md`. Per PIR's single-pass design this fix was **not** independently re-reviewed by either model — please verify it at this gate.
- **`afx tower sweep-husks` UX** mirrors `afx workspace recover` exactly (dry-run default, `--apply`, `-y`/`--yes`, same `confirm()` helper) per explicit architect direction mid-plan-review — not an independent design choice, so please check it actually matches that precedent rather than just looking reasonable in isolation.
- **`fleetRssKb` scope**: sums every in-scope shellper *and its direct children* (the full process-group tree), computed from one `ps` scan, regardless of DB-registration state — so a not-yet-swept husk still counts toward the visible number. This is explicitly advisory/observational, not used anywhere as a decision input (confirmed in conversation during the dev-approval gate review).
- **PID-reuse defense**: `computeRegisteredShellperPids` validates each DB row's `shellper_pid` against the live process's actual start time (2000ms tolerance, same constant `SessionManager.reconnectSession` already uses) before counting it as registered — a stale row whose PID was reused by an unrelated process correctly falls through to "unregistered."

## How to Test Locally

- **View diff**: VSCode sidebar → right-click builder pir-1227 → **Review Diff**
- **Run dev**: `afx dev pir-1227`
- **What to verify**:
  - `afx tower sweep-husks` with no flags previews only (nothing killed); `--apply` alone prompts for confirmation; `--apply --yes` reaps without prompting
  - `afx status` shows `Fleet RSS` and `Unregistered Shellpers` lines when Tower is running; `afx status --json` carries a `fleet: { rssKb, unregisteredShellperCount }` field (explicit `null`s, not omitted keys, when Tower is down)
  - `GET /health` on a running Tower includes `fleetRssKb` and `unregisteredShellperCount`
  - Tower starts and stops cleanly with the new interval in place (`SIGTERM` exits promptly, no hang)

## Flaky Tests

None encountered.


amrmelsayed:

High-risk band (Tower core, ~950 production+test lines) — full production-diff review, with focused verification of the post-consult fix commit since PIR's single-pass consultation left it unreviewed. **APPROVE.**

**Scope compliance (architect-scoped at spawn): exact.** Items 1+3 delivered — three sweep triggers (startup after reconciliation, hourly timer via `SHELLPER_HUSK_SWEEP_INTERVAL_MS`, on-demand `afx tower sweep-husks` with dry-run-default/`--apply` semantics) plus fleet RSS and unregistered-shellper observability in `/health` and `afx status`. Item 2 (hibernation) explicitly excluded and flagged as follow-up spec. Plan discipline was clean.

**The sweep predicate is safer than the issue demanded.** Beyond the required unregistered∧childless∧aged conjunction: (1) `computeRegisteredShellperPids` validates each DB row's PID against actual process start time (±2s, matching `reconnectSession`'s tolerance) so a stale row whose PID was recycled neither shields the recycled process nor miscounts as registered; (2) socketDir scoping uses the trailing-slash marker technique from `findShellperProcesses` (no `/run`-vs-`/run2` prefix collisions); (3) unknown start time → skip (fail-safe); (4) `resolveHuskGraceMs` NaN-checks rather than `|| default` so a deliberate zero survives. Kill delegated to #1225's `reapShellpers` (SIGTERM→poll→SIGKILL→confirm) rather than reimplemented. Both orderings of the startup-sweep-vs-reconciliation race degrade safely: stale rows over-protect (caught next hourly pass), and freshly-deregistered live shellpers are inside the 1h grace.

**The unreviewed fix commit (3a5083e3) — verified directly, it's correct.** codex's finding was real (`execFileSync` on `/health` = the lessons-learned.md:160 event-loop-blocking anti-pattern, would freeze every terminal's WebSocket for the `ps` duration). The fix: `listProcessCensus` is now callback-form `execFile` wrapped in a hand-rolled promise — with a documented reason for avoiding `util.promisify` (execFile's custom-promisify symbol shape vs test mocks) — resolving `[]` on any failure, 5s timeout, 16MB buffer. The redundant double-`ps` in the husk preview is eliminated via the census-as-array seam so `/health` takes one snapshot for both RSS display and husk detection. Regression tests pin the callback-form call, the timeout option, and never-rejects semantics.

**Field validation:** the predicate this ships is the same one hand-verified on this machine yesterday — 19 husks identified by exactly these conditions, killed with zero collateral (2026-07-23 census on #1227).

---
Architect integration review

Issue #1239, opened 2026-07-25
agent-farm: migration backups (~/.agent-farm.bak-*) are never reaped


waleedkadous:
Found during a disk audit (2026-07-25): a `~/.agent-farm.bak-20260702-113930` directory — created by a state migration on 2026-07-02 — was still on disk three weeks later at **18 GB** (a full copy of the agent-farm dir, dominated by the logs it duplicated). Nothing ever deletes these backups.

Proposal:
- Migrations that create a `.bak` should record it and reap it after a verification window (e.g. 7 days of successful operation on the migrated state), or
- `codev doctor` should at minimum surface existing `.bak` dirs with age + size so the human can decide.

Related: the no-rotation issue filed alongside this one — the .bak was so large mostly because the logs inside it had never rotated.

waleedkadous:
## Completion statistics

Shipped in #1242 (merge commit `25b2ef9c`, merged 2026-07-25T12:41Z).

| | |
|---|---|
| Protocol | AIR (Implement → Review) |
| Iterations | 1 — no rework rounds |
| Started → complete | 09:40Z → 12:42Z (3h 1m wall clock) |
| Active build time | ~12 min (09:40Z → 09:52Z, when the pr gate was requested) |
| Human gate wait | 2h 45m (09:52Z → 12:37Z) |
| CMAP | codex APPROVE/HIGH, claude APPROVE/HIGH — no key issues from either |
| Required CI | 6/6 green |

### Diff

| File | Lines |
|---|---|
| `packages/codev/src/lib/migration-backup-audit.ts` | +188 (new) |
| `packages/codev/src/__tests__/migration-backup-audit.test.ts` | +182 (new, 14 tests) |
| `packages/codev/src/commands/doctor.ts` | +45 −1 |
| `codev/resources/commands/codev.md` + skeleton mirror | +10 |
| `codev/state/air-1239_thread.md` | +66 |
| **Total** | **490 insertions, 1 deletion** |

Implementation LOC excluding tests and the builder thread: **243** — inside AIR's 300-LOC threshold, so no escalation was needed.

### Testing

- 14 new unit tests; full suite **183 files / 3655 tests passed**, 3 files / 48 tests skipped (pre-existing)
- No flaky tests encountered; nothing skipped by this change
- End-to-end: ran the built `codev doctor` against a real home directory carrying the actual leftovers. It found all five, with sizes cross-checked against `du -sh`

### Notes for the record

[...]

PR #1247, opened 2026-07-25
[AIR #1239] Carry post-merge thread addendum to main (docs-only)

waleedkadous:
Docs-only follow-up to #1242 (issue #1239). Touches exactly one file: `codev/state/air-1239_thread.md`. No code, no tests, no behavior change.

## Why this needs its own PR

The #1242 merge carried the builder thread to main **as of the pr gate**. Everything learned during and after the merge could not ride that same merge — under AIR the thread lands on main *via* the PR, so anything generated at or after merge time is structurally unable to be in it. Left alone, that content lives only on `builder/air-1239` and disappears at worktree cleanup.

The knowledge with the shortest path to being lost is precisely the knowledge generated last, which is why the addendum now says so explicitly rather than just carrying the notes.

## What's added

- **`porch done` timing for AIR's terminal `pr` phase.** `transition.on_complete` is `null`, so running it while the PR is still open flips the protocol to complete against reality. Wait for the merge.
- **`mergeStateStatus: BLOCKED` with all checks green is the required-review rule, not CI.** Easy to misread as a failing check and start debugging a green build.
- **Porch pushes its own gate-approved commit,** so the follow-up `git push` is a no-op reporting "Everything up-to-date" — not a failed push.
- **The porch "protocol complete" commit lands after the merge** and therefore never reaches main. Expected, not a missed push.
- **The cross-PR `formatBytes` collision** (see below) and the general lesson it teaches.

## Note: this PR's CI will be red, and not because of this PR

`main` does not currently build, and I found it while branching off main to open this. Reported to the architect separately:

```
src/commands/doctor.ts(27,3): error TS2300: Duplicate identifier 'formatBytes'.
src/commands/doctor.ts(36,3): error TS2300: Duplicate identifier 'formatBytes'.
```

#1242 (this project) and #1243 (air-1238, session logs) merged three minutes apart. Both added a `formatBytes` helper that `doctor.ts` imports — mine from `lib/migration-backup-audit.js`, theirs from `agent-farm/servers/session-log-sweep.js`. Two separate import statements, no textual conflict, clean auto-merge, duplicate identifier on the combination. Each PR's required checks were green against its own base, and each base predated the other, so nothing in the per-PR gate ever compiled the two together.

I have deliberately **not** fixed it here. It spans two projects and the right fix — alias one import, or consolidate the two duplicate helpers into one shared util — is an ownership decision rather than a builder's unilateral call. This PR is docs-only on purpose; folding a cross-project build fix into it would muddy both.

Once main is green this branch should go green with it; it changes one markdown file under `codev/state/`.

### DIFF
--- a/codev/state/air-1239_thread.md
+++ b/codev/state/air-1239_thread.md
@@ -64,3 +64,55 @@ pre-existing drift is out of scope here.
 PR-phase checks green (`pr_exists`, `e2e_tests`). `porch gate 1239` registered
 the **pr gate — waiting for human approval**. Not running `porch approve`;
 that's the human's call.
+
+## Closed out
+
+Waleed approved the pr gate at 12:37Z (relayed by the architect). Ran
+`porch approve 1239 pr --a-human-explicitly-approved-this`; porch pushed the
+gate-approved commit itself, so the follow-up `git push` was a no-op
+confirming parity — worth knowing so a sibling doesn't think the push failed.
+
+All 6 required CI checks green. Note `mergeStateStatus` stayed `BLOCKED` with
+every check passing — that's the required-review rule, not CI, and it's what
+the architect's admin-merge clears. Don't misread it as a failing check.
+
+I held `porch done` until the merge actually landed: AIR's `pr` phase is
+terminal (`transition.on_complete: null`), so advancing while the PR was still
+open would have flipped the protocol to complete against reality. Merged
+12:41Z (`25b2ef9c`), then `porch done 1239` → PROTOCOL COMPLETE. Issue #1239
+auto-closed on merge; completion stats posted as an issue comment per the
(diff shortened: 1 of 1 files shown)


waleedkadous:
Closing unmerged after discussion with Waleed — the content is good, but the vehicle sets the wrong precedent. Post-merge learnings should flow builder → architect via afx send (which you already did, exemplarily), and the architect routes the durable bits into lessons-learned/issues at the next MAINTAIN pass. 'Every builder opens a post-merge thread PR' doesn't scale to the cohort. Your two lessons (AIR terminal-phase timing; 'green CI is evidence about the PR's base, not about main') are captured and will land in codev/resources/lessons-learned.md. Thank you — the loss-analysis in this PR's framing is exactly why nothing was actually lost.

PR #1250, opened 2026-07-25
[Bugfix #1077] Pre-flight agy auth cache to stop the OAuth tab burst

### LINKED ISSUE #1077: consult: pre-flight auth-state cache for agy to prevent N-spawn browser-tab burst on unauthenticated state
## Problem

When the `agy` (Antigravity CLI) binary is **not currently authenticated**, every Codev consult invocation that hits the gemini lane spawns a fresh `agy --print` process, and that process opens an OAuth browser tab via macOS `open` *before* it prints the OAuth URL to its stderr. Codev's consult code (`packages/codev/src/commands/consult/index.ts:843-848`) watches the early stream for `AGY_OAUTH_MARKERS` and kills agy with a non-blocking COMMENT skip the moment it sees one. That detection is correct semantically, but it fires *after* the browser tab has already opened. The kill closes the barn door once the horse is out.

Across N concurrent CMAP-3 consults (the architect-builder workflow runs them in parallel by design), N tabs open in rapid succession. On 2026-06-19 a user reported ~12-15 stranded OAuth tabs in a single Chrome window. The screenshot at `tmp/screenshots/agy-oauth-tab-burst-2026-06-19.png` in the architect worktree shows the tab strip filled with stacked "Sign in with Google: Choose an account to continue to Google Antigravity" tabs.

## Why the user-side workaround does not stop the bleed

Running `agy` interactively once and completing the OAuth in one tab does write a persistent token, but the *other* in-flight consults that have already spawned agy children before that token landed cannot benefit. Each one is racing toward its own `AGY_OAUTH_MARKERS` detection. The user observes new tabs continuing to appear until the entire CMAP-3 burst has played out.

## Where the spawn pressure compounds

Three amplifiers, each independently verified in the code:

1. **CMAP-3 = 3 separate `codev consult -m <model>` process spawns per round** (`packages/codev/src/commands/porch/next.ts:529-530`). Only the gemini lane spawns agy, so one agy spawn per CMAP-3 round per consult phase.
2. **Multiple consult phases per builder.** PIR runs plan-approval consult + review consult = 2 agy spawns per builder. SPIR adds spec consult = 3 spawns. AIR / BUGFIX = 1 spawn at review.
3. **Porch re-emits the same consult tasks when review files look "partial"** (`packages/codev/src/commands/porch/next.ts:544`, "Partial reviews, still waiting. Emit same consultation tasks (idempotent)."). If gemini's COMMENT-skip output is malformed (e.g. because the kill was racy and stdout was cut mid-banner), porch re-emits all 3 consult commands. The builder runs them again. The gemini one re-spawns agy. Another tab opens.

With 5 to 6 active builders mid-cycle, each running 1 to 2 consult phases, with the occasional re-emit, the math reaches the observed 12-15 in the visible burst.

## What this issue tracks

A session-scoped **auth-state cache + pre-flight** that short-circuits `runAgyConsultation` when the cache says agy is unauthenticated, so the second and subsequent consult calls in a burst do not spawn fresh agy children.

The skip-with-COMMENT contract that ships today is already the right *semantic* behavior; this issue is about doing the skip *before* the agy spawn, not after.

## Proposed fix

Add a cross-process, filesystem-backed auth-state cache to `packages/codev/src/commands/consult/`. The shape:

1. **Cache file:** `~/.cache/codev/agy-auth.json`. Stores `{ state: 'auth' | 'unauth', checkedAt: <epoch-ms>, agyBinPath: '<resolved-path>', agyAuthMtime: <epoch-ms> }`. Permissions 0600.
2. **TTL:** 5 minutes for `state: 'unauth'`. Short enough that a user who completes OAuth in a separate terminal sees the cache invalidate promptly. Longer TTL (e.g. 30 min) for `state: 'auth'`, since authenticated probes are non-intrusive.
3. **Pre-flight check in `runAgyConsultation`** (before `spawn(bin, args, ...)`). Read the cache:
   - If `state: 'unauth'` and the TTL has not expired: emit the COMMENT skip immediately, with reason `"agy unauthenticated (cached); run agy once to sign in"`. **No spawn, no browser tab.**
   - If `state: 'auth'` and the TTL has not expired: proceed with the spawn as today.
   - If the cache is missing or expired: probe agy with a fast `--print-timeout 5s` call. The probe itself may open at most 1 tab if unauth (irreducible; this is the upstream agy behavior). Persist the result with `agyAuthMtime` set to the mtime of agy's own credentials file if one can be located. Then either skip (if unauth) or proceed (if auth).
4. **Cache invalidation on successful auth.** When the user completes OAuth in any tab, agy writes its credentials to disk. If we can identify the credentials path (candidates: `~/.antigravity/credentials.json`, `~/Library/Application Support/Antigravity/credentials.json`, the LevelDB at `~/Library/Application Support/Google/Chrome/Default/IndexedDB/https_antigravity.google_0.indexeddb.leveldb`), watch its mtime: if it changed since `agyAuthMtime`, invalidate the cache and re-probe on the next consult call. As a fallback if the credentials path is opaque, accept the TTL as the recovery window.
5. **Cross-process safety.** Multiple `codev consult` processes started in parallel must not all race to probe simultaneously (that would defeat the point). Wrap the probe in an exclusive file lock (`flock`-style on `~/.cache/codev/agy-auth.lock`) with a short wait. The lock holder probes; others read the cache the moment the holder releases.

## Acceptance criteria

- [ ] A new module (e.g. `packages/codev/src/commands/consult/agy-auth-cache.ts`) exposes `checkCachedAgyAuth()` and `recordAgyAuthState(state, agyBinPath)`.
- [ ] `runAgyConsultation` calls `checkCachedAgyAuth()` before `spawn`. On `unauth` (fresh), it calls `settleSkip` immediately with reason mentioning the cache.
- [ ] On a fresh process with no cache file, the first invocation probes agy with `--print-timeout 5s` and writes the result.
- [ ] Concurrent `codev consult` processes do not all probe agy simultaneously: a file lock serializes the probe; non-holders read the cache when the lock releases.
- [ ] Cache TTLs: 5 min for `unauth`, 30 min for `auth`. Both overridable via env (`CODEV_AGY_AUTH_CACHE_TTL_UNAUTH_MS`, `CODEV_AGY_AUTH_CACHE_TTL_AUTH_MS`) for tests.
- [ ] When the user completes OAuth in another window during the unauth-TTL window, the next consult call after the TTL expires re-probes and finds `auth` (no manual cache clear required).
- [ ] Smoke test: spawn 5 parallel `codev consult -m gemini` invocations against an unauthenticated agy. Verify at most 1 spawn of agy happens, not 5; the other 4 short-circuit on the cache.
- [ ] Smoke test: with valid `auth` cache, a `codev consult -m gemini` invocation completes normally (no behavior regression).
- [ ] `codev doctor`'s `verifyAgy` (`packages/codev/src/commands/doctor.ts:380`) reads from the same cache when fresh, instead of probing on every doctor run. (Optional but obvious extension.)
- [ ] Updates: brief mention in the consult lane docs in CLAUDE.md / AGENTS.md and the `consult` commands reference under `codev/resources/commands/consult.md`.

## Alternatives considered

- **Don't cache; serialize agy spawns with a global lock.** Replace step 5 of the proposal with a process-wide lock: only one `runAgyConsultation` may have agy spawned at a time, all others wait. Eliminates the burst, but adds latency to the legitimate case and does not stop a single unauth spawn from opening 1 tab per N consult phases. Worse user experience than the cache for the same problem.
- **Detect unauth without spawning at all by stat'ing agy's credentials file.** Cheapest possible pre-flight, no spawn ever needed. Blocked on: agy's credentials path is not officially documented, the LevelDB form is non-trivial to read from Node, and any path change in a future agy release breaks the heuristic silently. Keep this as a hint to invalidate the cache, not as the only check.
- **Add a `codev.consult.gemini.enabled` setting and let the user toggle the lane off.** Manual workaround, not a fix. Useful as a separate escape hatch but does not address the burst for users who want the gemini lane working.
- **Wait for upstream agy to add a non-interactive auth probe.** Tracked at the agy CLI repo, no ETA. Codev's pre-flight cache is the right defense in depth regardless.

## Out of scope

- The upstream agy behavior of opening a browser tab before printing the OAuth URL. That is the root cause of "one spawn opens one tab," but Codev cannot fix it directly. The cache reduces N-spawn-bursts to at most 1 spawn per TTL window, which is the achievable mitigation.
- Migrating away from agy to a different gemini backend. Out of scope; the gemini lane swap to agy was the deliberate decision recorded in #778 / PR #988.
- Changes to the COMMENT-skip non-blocking semantics in porch's CMAP-3 verdict aggregation. Today's behavior is correct: skip + non-blocking advance. This issue only changes how soon the skip fires.
- Other agy hardening (sandbox isolation, off-topic reviews, malformed verdicts). Tracked under #1051, #1032, #1020.

## Why BUGFIX (or AIR)

The cache + pre-flight is well-bounded: one new module, two integration points (`runAgyConsultation` and the optional `doctor.verifyAgy`), small set of acceptance criteria. The behavior change is invisible to the user when agy is authenticated and dramatic when it is not. Could ship as BUGFIX (it fixes the N-spawn bug) or AIR (small additive feature). Lean: BUGFIX. PIR's plan-approval gate is overhead for this scope.

## Related

- **#1020**, harden agy integration (sibling: silent drop-out when agy emits empty stdout). This issue addresses a different agy failure mode (unauth burst) but shares the broader "agy integration robustness" theme. Could fold as a sub-task; the recommendation is to keep them separate so they ship independently.
- **#1051**, agy runs in empty sandbox + can execute commands.
- **#1032**, agy goes off-task on structured reviews (emits no VERDICT).
- **#778 / PR #988**, the original gemini-to-agy lane swap. Background for why agy is the backend.
- `packages/codev/src/commands/consult/index.ts:614-862`, the agy runner + skip detection.
- `packages/codev/src/commands/doctor.ts:380-425`, the doctor `verifyAgy` probe (precedent for the spawn-with-timeout pattern).
- `packages/codev/src/commands/porch/next.ts:529-544`, the CMAP-3 dispatch and the partial-review re-emit logic that amplifies the burst.

## Reference: tab-burst screenshot

The architect captured a screenshot of the stranded OAuth tab strip in Chrome on 2026-06-19; it is preserved at `tmp/screenshots/agy-oauth-tab-burst-2026-06-19.png` in the architect worktree (gitignored, not pushed). About 12-15 "Sign in with Google" tabs are visible at once, all pointing at `accounts.google.com/v3/signin/accountchooser?access_type=offline&client_id=10710060...` for Google Antigravity. The architect will attach the image to this issue via the web UI as a visual reference.

waleedkadous:
## Summary

The gemini consult lane spawned `agy` before it could know `agy` was signed out, and every spawn opens an OAuth browser tab. A CMAP round is N independent `consult` processes, so N tabs stranded — 12-15 in the reported case.

This adds a cross-process auth-state cache consulted **before** the spawn. Unauthenticated `agy` is now spawned **at most once per TTL window** instead of once per consult.

Measured with a fake `agy` that logs every invocation (spawn counts are observed, not mocked):

| Scenario | Before | After |
|---|---|---|
| 5 parallel `consult -m gemini`, unauthenticated | 5 spawns / 5 tabs | **1** |
| 5 parallel, authenticated | 5 | **5** (unchanged — every lane is real work) |
| unauth verdict already cached | 5 | **0** |

## Root Cause

`runAgyConsultation` (`packages/codev/src/commands/consult/index.ts`) spawned `agy` unconditionally, and `AGY_OAUTH_MARKERS` detection lives in the child's stdout/stderr handler — i.e. strictly *after* the spawn. Upstream `agy` opens the OAuth tab before printing the URL those markers match, so the kill can never prevent the tab. The skip semantics were already correct; only the timing was wrong.

A module-level memo cannot help: each `consult` invocation is a separate OS process, so the state has to be shared through the filesystem.

Amplifiers confirmed in code: CMAP-N dispatches one process per model (`porch/next.ts:529-531`), builders run 1-3 consult phases each, and the partial-review re-emit (`porch/next.ts:544`) re-issues the same consult commands.

## Fix

New module `packages/codev/src/commands/consult/agy-auth-cache.ts`:

- Verdict in `~/.cache/codev/agy-auth.json` (0600 in a 0700 dir, honours `XDG_CACHE_HOME`), keyed to the resolved agy binary, written via temp-file + rename so a concurrent reader never sees half a document.
- TTLs: 5 min `unauth` (the sign-in recovery window), 30 min `auth`. Both env-overridable.
- A `flock`-style lock (`openSync` `'wx'`, with stale-lock reclaim) elects **one prober** per cold window. Losers poll the *cache*, not the lock — the prober holds it for its whole run, which may be minutes, but publishes the verdict within seconds.
- Best-effort credential-mtime hint invalidates a verdict early when agy rewrites its credentials. agy's credential path is undocumented and was not locatable on this machine, so the TTL is the real recovery mechanism and the hint is a bonus.

Two integration points: `runAgyConsultation` pre-flights before `spawn`, and `doctor.verifyAgy` both reports a fresh verdict without probing and records the verdicts its own probe produces.

### Deliberate deviation from the issue's proposal

The issue proposes a dedicated `--print-timeout 5s` probe on a cold cache. Implemented instead: **the real consult run doubles as the probe**, publishing the verdict as soon as it is knowable (OAuth marker → `unauth`; real output, or 3s of marker-free silence → `auth`). This avoids a second agy spawn per TTL window in the authenticated common case, and keeps marker detection in one place rather than duplicating it in a path that could classify differently. The observable guarantee the ACs ask for is unchanged.

### Failure bias

Only positive marker evidence records `unauth`, and a waiter that never sees a verdict **proceeds anyway**. Both choices push failures toward "one unnecessary browser tab" (the status quo) and away from "silently skipped a working reviewer", which would corrupt a CMAP round.

## Test Plan

`packages/codev/src/commands/consult/__tests__/agy-auth-cache.test.ts` — 25 tests. The headline case drives the real lane (`_runAgyConsultation`) against a fake `agy` that logs every invocation, so a 5-wide unauthenticated burst is asserted to produce exactly **one** spawn. A companion case asserts **3** spawns with the cache disabled, which is what proves the burst test measures the cache rather than something incidental.

Also covered: per-state TTL expiry, bin-path and credential-mtime invalidation, corrupt/unknown entries, lock election, stale-lock reclaim, sign-in recovery after the unauth TTL, and that an authenticated burst still runs every lane.

- Full suite: **3720 passed**, 0 failed (`npm test`), build green. Both verified again by porch's own `fix` phase checks.
- Cross-process smoke (5 real `consult` processes, not in-process): 1 spawn unauthenticated, 5 authenticated with all 5 reviews delivered.
- Recovery verified end-to-end: unauth cached → sign in → wait out the TTL → next call re-probes, finds `auth`, delivers a review. No manual cache clear.

### Note on test isolation

The cache is a **user-global** side effect. Unpinned, the suite wrote a verdict into the developer's real `~/.cache/codev`, and one `doctor` case's verdict made a sibling case stop spawning the thing it asserted on (one genuine failure, caught and fixed). The consult/doctor suites now pin `CODEV_AGY_AUTH_CACHE_DIR` into their temp dirs, and the module is inert under `VITEST` unless a cache dir is explicitly set — the guard is what keeps *future* tests from silently re-acquiring the hazard. Verified: a full suite run no longer creates `~/.cache/codev`.

## Scope

Net diff is ~960 lines across 12 files, above BUGFIX's ~300 guideline. The composition: ~310 lines of new module (roughly half comments), 74 lines across the two integration points, ~370 lines of tests, ~58 of docs. Shape is what the guideline actually targets — one new module plus two call sites, no architectural change, no new abstractions imposed on callers. Flagging it rather than quietly ignoring it; happy to split the docs out if you'd prefer a smaller diff.

## Notes

- `packages/codev/src/commands/doctor.ts` also carries the `formatBytes` alias from the hotfix in #1248, which arrived via a `main` merge — not a change of mine.
- Docs: `consult.md` (mechanism, recovery, env vars), mirrored into `codev-skeleton/`; CLAUDE.md / AGENTS.md gemini-lane bullet, kept byte-identical.

Fixes #1077

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -339,7 +339,11 @@ Use sequential numbering with descriptive names (no leading zeros):
 **DEFAULT BEHAVIOR**: Consultation is ENABLED by default with:
 - **Gemini** via the **Antigravity CLI (`agy`)** for deep analysis (the retired Gemini CLI's
   replacement; OAuth/subscription, agy's default model — no pinned model id). Skips non-blockingly
-  if `agy` is missing/unauthenticated.
+  if `agy` is missing/unauthenticated. An unauthenticated `agy` is spawned **at most once per TTL
+  window** rather than once per consult: the verdict is cached across processes in
+  `~/.cache/codev/agy-auth.json`, because each spawn opens an OAuth browser tab before Codev can
+  detect the missing login (#1077). Sign in with `agy` in any terminal and the lane recovers on its
+  own; see `codev/resources/commands/consult.md` for the TTL/opt-out env vars.
 - **GPT-5.4 Codex** (gpt-5.4-codex) for coding and architecture perspective
 
 To disable: User must explicitly say "without multi-agent consultation"
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -339,7 +339,11 @@ Use sequential numbering with descriptive names (no leading zeros):
 **DEFAULT BEHAVIOR**: Consultation is ENABLED by default with:
 - **Gemini** via the **Antigravity CLI (`agy`)** for deep analysis (the retired Gemini CLI's
   replacement; OAuth/subscription, agy's default model — no pinned model id). Skips non-blockingly
-  if `agy` is missing/unauthenticated.
+  if `agy` is missing/unauthenticated. An unauthenticated `agy` is spawned **at most once per TTL
+  window** rather than once per consult: the verdict is cached across processes in
+  `~/.cache/codev/agy-auth.json`, because each spawn opens an OAuth browser tab before Codev can
+  detect the missing login (#1077). Sign in with `agy` in any terminal and the lane recovers on its
+  own; see `codev/resources/commands/consult.md` for the TTL/opt-out env vars.
 - **GPT-5.4 Codex** (gpt-5.4-codex) for coding and architecture perspective
 
 To disable: User must explicitly say "without multi-agent consultation"
--- a/codev-skeleton/resources/commands/consult.md
+++ b/codev-skeleton/resources/commands/consult.md
@@ -170,6 +170,32 @@ Configure auth:
 - Gemini (`agy`): **OAuth / subscription** — run `agy` once and sign in (no API key). If `agy`
   is missing or unauthenticated, the gemini lane skips non-blockingly (the run proceeds without it).
 
+#### agy auth-state cache (unauthenticated tab burst)
+
+An unauthenticated `agy` opens an OAuth browser tab *before* it prints the URL
+Codev detects, so the lane cannot suppress the tab after spawning — only by not
+spawning. Since a CMAP round is several independent `consult` processes, this used
+to strand one tab per consult (#1077).
+
+Codev keeps a shared auth verdict in `~/.cache/codev/agy-auth.json` (0600,
+honours `XDG_CACHE_HOME`). The first call probes; the rest read the verdict and
+short-circuit, so **an unauthenticated agy is spawned at most once per TTL
+window** instead of once per consult. A file lock keeps parallel consults from
+probing simultaneously. `codev doctor`'s agy check shares the same cache.
+
+Verdicts expire on their own — sign in with `agy` in another terminal and the lane
+recovers within the unauth TTL, no cache clearing required. The cache only ever
+suppresses a lane on positive "signed out" evidence; anything ambiguous falls
+through to a normal spawn.
+
+| Variable | Default | Purpose |
+|---|---|---|
+| `CODEV_AGY_AUTH_CACHE_TTL_UNAUTH_MS` | `300000` (5 min) | How long a "signed out" verdict is trusted — the sign-in recovery window. |
+| `CODEV_AGY_AUTH_CACHE_TTL_AUTH_MS` | `1800000` (30 min) | How long a "signed in" verdict is trusted. |
+| `CODEV_AGY_AUTH_CACHE_WAIT_MS` | `6000` | How long a consult waits for another process's in-flight probe before proceeding anyway. |
+| `CODEV_AGY_AUTH_CACHE_DIR` | `~/.cache/codev` | Cache location (mainly for tests). |
+| `CODEV_AGY_AUTH_CACHE_DISABLE` | unset | `1` restores the always-spawn behaviour. |
+
 ### Claude auth: subscription vs. metered API
 
 `consult -m claude` runs on the Claude Agent SDK. When `CLAUDE_CODE_OAUTH_TOKEN`
--- a/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
+++ b/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1077
 title: consult-pre-flight-auth-state-
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-25T12:49:13.351Z'
-updated_at: '2026-07-25T13:10:27.468Z'
+updated_at: '2026-07-25T13:11:31.792Z'
--- a/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
+++ b/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-07-25T13:15:16.670Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-25T12:49:13.351Z'
-updated_at: '2026-07-25T13:11:31.792Z'
+updated_at: '2026-07-25T13:15:16.670Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
+++ b/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1077
 title: consult-pre-flight-auth-state-
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-25T12:49:13.351Z'
-updated_at: '2026-07-25T12:49:13.351Z'
+updated_at: '2026-07-25T13:10:27.468Z'
--- a/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
+++ b/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
@@ -6,11 +6,12 @@ plan_phases: []
 current_plan_phase: null
 gates:
   pr:
-    status: pending
+    status: approved
     requested_at: '2026-07-25T13:15:16.670Z'
+    approved_at: '2026-07-25T13:23:32.434Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-25T12:49:13.351Z'
-updated_at: '2026-07-25T13:15:16.670Z'
-pr_ready_for_human: true
+updated_at: '2026-07-25T13:23:32.435Z'
+pr_ready_for_human: false
--- a/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
+++ b/codev/projects/bugfix-1077-consult-pre-flight-auth-state-/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1077
+title: consult-pre-flight-auth-state-
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-07-25T12:49:13.351Z'
+updated_at: '2026-07-25T12:49:13.351Z'
--- a/codev/resources/commands/consult.md
+++ b/codev/resources/commands/consult.md
@@ -170,6 +170,32 @@ Configure auth:
 - Gemini (`agy`): **OAuth / subscription** — run `agy` once and sign in (no API key). If `agy`
   is missing or unauthenticated, the gemini lane skips non-blockingly (the run proceeds without it).
 
+#### agy auth-state cache (unauthenticated tab burst)
+
+An unauthenticated `agy` opens an OAuth browser tab *before* it prints the URL
+Codev detects, so the lane cannot suppress the tab after spawning — only by not
+spawning. Since a CMAP round is several independent `consult` processes, this used
+to strand one tab per consult (#1077).
+
+Codev keeps a shared auth verdict in `~/.cache/codev/agy-auth.json` (0600,
+honours `XDG_CACHE_HOME`). The first call probes; the rest read the verdict and
+short-circuit, so **an unauthenticated agy is spawned at most once per TTL
+window** instead of once per consult. A file lock keeps parallel consults from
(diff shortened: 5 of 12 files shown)


waleedkadous:
## CMAP-3 review

| Reviewer | Verdict | Confidence | Key issues |
|---|---|---|---|
| Gemini (agy) | **APPROVE** | HIGH | None |
| Codex | **APPROVE** | HIGH | None |
| Claude | **APPROVE** | HIGH | None |

Unanimous, no requested changes, so nothing to rebut or fix.

What each reviewer independently called out:

- **Gemini** — atomic write semantics (temp + rename), 0600/0700 permissions, `'wx'` lock with stale reclaim, and fail-open polling "so active review runs are never improperly blocked".
- **Codex** — confirmed the cross-process framing (in-memory memoization cannot stop parallel bursts) and the fail-open bias. Note: it could not execute Vitest — its sandbox is read-only and Vitest failed creating `.vite-temp` — so it reviewed by inspection and relied on the reported green build/tests. Those were independently verified here by porch's own `fix`-phase `build` + `tests` checks.
- **Claude** — verified the burst regression test plus the disabled-cache control case, and the sign-in recovery path.

waleedkadous:

Excellent work — this is a genuinely well-engineered fix. The single-prober lock election with cache-polling waiters, the fail-open bias, and the spawn-count-observed (not mocked) burst regression test are all exactly right. My two integration lanes agree with your CMAP-3: **codex APPROVE** (called the module boundary "the right system boundary for a cross-process concern") and **claude APPROVE, high confidence** (independently verified the two-tree mirroring and CLAUDE/AGENTS byte-identity).

On your two flagged judgment calls:

1. **Probe deviation** — agreed, your version is better than the issue's proposal. Reusing the real run as the probe avoids an extra spawn on the authenticated path and keeps marker detection in one place. No concern.
2. **Scope** — agreed, no split needed. ~370 of the lines are tests and ~58 docs; one module + two call sites is BUGFIX-shaped regardless of line count.

One change I'd like before merge, because it protects the headline guarantee:

**Let positive OAuth-marker evidence write the cache unconditionally.** Today the marker path reports `unauth` only through `preflight.publish`, which is once-only. If the grace timer has already fired at 3s (agy slower than 3s to emit its banner — a node CLI doing a network token check makes that plausible), the timer publishes a false `auth` that lives for 30 minutes, and the real `unauth` evidence a moment later becomes a no-op. During that window every burst spawns tabs again — the fix silently reverts to the status quo on exactly the machines that need it. Suggested shape: in the marker handler, call `recordAgyAuthState('unauth', bin)` directly (in addition to the existing `publishAuth('unauth')` for lock release). That also means *non-prober* runs that hit the banner publish the verdict, so a misfired grace timer gets corrected by the very next spawn instead of waiting out the TTL. `recordAgyAuthState` already carries the disabled-guard, so the call is safe everywhere.

Perhaps add one test alongside: grace timer fires `auth`, marker arrives late, assert the cache ends up `unauth`.

Everything else is merge-ready.

waleedkadous:
## Requested change pushed — 42cf18e8

Good catch, and you were right that this protected the headline guarantee. Marker detection now calls `recordAgyAuthState('unauth', bin)` directly, so positive evidence wins regardless of arrival order.

Worth noting it was **two** bugs, not one — the second I had missed as well:

1. **The grace-timer race you described.** Banner later than the 3s grace → timer publishes `auth` → real `unauth` becomes a no-op → false verdict for the full 30-min TTL.
2. **Non-prober evidence was discarded entirely.** A caller that failed open (another process holding the lock) still spawns and can be the one to see the banner, but `publish` is inert for non-probers, so it wrote nothing at all. Even with the ordering fixed, a misfired guess would have sat uncorrected until the TTL lapsed rather than being fixed by the next spawn. Writing unconditionally from the marker handler resolves both at once, exactly as you suggested.

This was a real inversion of the fail-safe bias I claimed in the PR body: I wrote that "only positive marker evidence records `unauth`", but the implementation let a *speculative* verdict outrank *positive* evidence purely on timing.

### Tests

Two added, and I verified both **fail against the previous code** before restoring the fix — not just pass with it:

| Test | Without fix | With fix |
|---|---|---|
| late OAuth banner (3.4s, past the grace period) overwrites the speculative verdict | FAIL (cached `auth`) | cached `unauth` |
| the run that sees the banner is not the prober | FAIL (nothing cached) | cached `unauth` |

Full suite **3722 passed**, 0 failed (up 2). Typecheck clean. Cross-process burst smoke re-run after the change: 5 parallel unauthenticated consults → still **1** spawn.

No changes to the docs or the public surface. Gate is still yours.

PR #1266, opened 2026-07-27
[Bugfix #1264] A clean harness exit reruns fresh instead of killing the session

### LINKED ISSUE #1264: Regression (3.2.4): double Ctrl-C on an agent no longer restarts it fresh — classified as deliberate quit
## Problem

Since PR #1244 (shipped in v3.2.4), killing an agent with **Ctrl-C Ctrl-C no longer restarts it**. The intended UX for double Ctrl-C is *restart fresh* — it's how users recover a wedged agent. Instead the session now ends cleanly and stays dead, requiring a manual respawn.

## Root cause

PR #1244 (bugfix #1241) added exit classification so a deliberate quit (typing `exit` / `/quit`) stops auto-restarting the agent. The classifier is:

```ts
// shellper-protocol.ts:128
export function isDeliberateExit(exit) {
  if (exit.code !== 0) return false;
  return !exit.signal; // (null/undefined/''/'0')
}
```

The Claude CLI traps SIGINT and exits **code 0** on a double Ctrl-C — indistinguishable, at the exit-status level, from a typed quit. So the double-Ctrl-C path is now classified deliberate → no restart. The #1241 test matrix covered typed quits and crash exits but not the double-Ctrl-C recovery gesture.

## Expected behavior

- Typing `exit` / `/quit` in the agent → session ends cleanly, no restart (the #1241 fix — keep this).
- **Double Ctrl-C → agent restarts fresh** (pre-3.2.4 behavior — restore this).
- Crashes / signals / nonzero exits → auto-restart (unchanged).

## Implementation hint (not prescriptive)

Exit status alone can't discriminate the two gestures. One candidate: the shellper/PTY layer sees the raw input stream — recent `^C` (0x03) bytes in the input immediately preceding the exit could mark the exit as Ctrl-C-driven and therefore restartable. Another: check whether the Claude CLI distinguishes these cases in any observable way (exit code, stderr, tty state) before inventing input-sniffing. Verify against the real CLI, not assumptions.

## Notes

Regression is live in published v3.2.4 — fix likely warrants a fast-follow patch release.

waleedkadous:
> **Reworked 2026-07-27 after a spec correction from the architect** ([comment on #1264](https://github.com/cluesmith/codev/issues/1264#issuecomment-)). The first implementation discriminated `^C^C` from a typed quit by sniffing `0x03` in the shellper's input stream; it was reviewed (CMAP below) and one real defect fixed. The corrected spec makes that whole approach unnecessary. The branch was force-pushed to carry only the new design — the superseded commits (`ed10e2fc`…`3d525114`) remain reachable via the force-push record.

## Summary

Since v3.2.4, killing an agent with double Ctrl-C left it dead and needed a manual respawn. The fix is **not** to tell `^C^C` apart from a typed quit — the corrected framing is that **a clean harness exit should never kill the shellper at all**.

Fixes #1264

## Root Cause

#1241 made *any* clean exit end the session, reading a quit as a decision that must not be overridden. Probing the real CLI shows why discrimination was a dead end anyway — Claude CLI 2.1.220 under node-pty:

| Gesture | node-pty exit |
|---|---|
| `^C^C` | `{exitCode: 0, signal: 0}` |
| typed `/quit` | `{exitCode: 0, signal: 0}` |

Byte-identical. The corrected spec removes the need to distinguish them.

Scope note: `restartOnExit` is set **only** for architect sessions (`tower-instances.ts` ×2, `tower-terminals.ts` ×1), so this is the surface that actually died. Builders run the harness inside a bash loop and are **out of scope by architect decision** — tracked in #1267.

## Fix

- **Any clean exit** → shellper and session survive; harness reruns in the same PTY **without recovery** (no `--resume`; fresh conversation).
- **Unnatural exits** (crash, signal, nonzero) → unchanged, restart **with** recovery.
- **A session ends only via explicit kill** (afx / UI / `DELETE`).

| File | Change |
|---|---|
| `session-manager.ts` | Clean exit reruns via a caller-supplied `FreshLaunch` factory instead of ending the session; not counted as a restart. |
| `pty-session.ts` | Clean exit takes the restart-wait path (session survives), sharing an extracted `startRestartWait` with the crash path — differing only in the notice. |
| `tower-utils.ts` | `buildArchitectFreshLaunch` — mints a new conversation id, re-injects the role, **persists** the id to the architect row. |
| `tower-instances.ts` ×2, `tower-terminals.ts` ×1 | Wire the factory. |

Two things that are load-bearing rather than incidental:

**`FreshLaunch` is a factory, not a precomputed arg list.** Every clean exit is a genuinely new conversation and needs its own minted id. Reusing one set of args would make the second `^C^C` collide with the session id the first one just created.

**The new id is persisted.** Otherwise an unnatural exit *after* a clean rerun would resume the conversation the user just deliberately walked away from. All three wiring sites build the factory from `baseArgs`/`cmdParts`, never from the resolved args — those may already carry `--resume`, which would defeat the whole point.

### Design note: the fast-clean-exit valve (architect-approved, first-class)

Making clean-exit reruns unlimited is correct — quitting is a gesture, not a failure — but it removes a bound that **both** prior versions had (#1241 ended the session; pre-3.2.4 had `maxRestarts`). A harness misconfigured to exit 0 on startup would then respawn forever.

So: a clean exit within **2s of launch** is evidence of a broken command, not a human gesture — **5 consecutive** such exits give up. A real quit never trips it, and one fast exit followed by a healthy session resets the counter. Genuine gestures stay unlimited.

The give-up is **loud by design**: a new `session-gave-up` event carries a plain-language reason that Tower writes *into the terminal* (`PtySession.notice`) as well as the log, so a user with a misconfigured command sees why their pane went quiet instead of meeting a mystery dead session. This deliberately does **not** reuse `session-error`, which also fires for transient client errors that do not end the session — surfacing that as a teardown notice would lie.

**Known gap, not addressed here:** `maxRestarts` exhaustion has the identical silent-death UX and does not get this treatment. Out of scope for this fix; flagging it as a follow-up candidate rather than quietly widening the diff.

## Test Plan

**End-to-end** — real `SessionManager` → real shellper process → real `claude`, asserting on the argv actually relaunched:

| Gesture | Events | Relaunch argv | |
|---|---|---|---|
| `^C^C` | `fresh-restart`, session live, agent back up | new `--session-id`, **no `--resume`**, factory called 1× | PASS |
| SIGKILL | `restart#1`, session live, agent back up | **argv unchanged** (recovery preserved), factory called 0× | PASS |

**Unit** (replacing the #1241 matrix, which asserted the behavior this reverses):
- session survives a clean exit; no `session-clean-exit`
- relaunch carries no `--resume` and none of the old conversation id
- a **new** conversation per rerun, never reusing the first
- `restartCount` and crash-loop history untouched across repeated clean exits
- no-factory sessions (plain shells) still rerun, args unchanged
- crash and signal death restart with recovery, factory never consulted
- PtySession suppresses `exit` and awaits the rerun; bounded wait still tears down if none arrives
- the real-shellper `exit 0` case now asserts the fast-clean-exit valve, and still never applies `crashLoopFallback` (#1149)

**Full suite: 3759 passed, 0 failed, 48 skipped** (194 files); build green.

## CMAP on the superseded design (kept for the record)

gemini APPROVE · claude APPROVE · **codex REQUEST_CHANGES** — codex correctly found that Ctrl-D after a Ctrl-C would have restarted an EOF quit; fixed before the rework. Two approvals did not outvote one specific, reproducible finding. Re-CMAP on this diff is running.
- the fast-clean-exit valve: gives up after 5 immediate clean exits with a reason naming the behavior and saying respawning stopped; one healthy session resets the counter; real (slow) quits never trip it at 2× the threshold
- **freshLaunch survives `reconnectSession`** — a post-Tower-restart clean exit still relaunches fresh, driven through the real reconnect path over a real shellper socket (CMAP finding, see below)

## CMAP

**Round 1 (superseded `0x03` design)** — gemini APPROVE · claude APPROVE · **codex REQUEST_CHANGES**: Ctrl-D after a Ctrl-C would have restarted an EOF quit. Fixed before the rework.

**Round 2 (this design)** — gemini APPROVE · claude APPROVE · **codex REQUEST_CHANGES**:

> `reconnectSession()` accepts `restartOptions.freshLaunch` in its type, but never copies it into `session.options`.

Correct, and the most consequential finding in either round. A reconnected architect's args already carry `--resume`, so **after any Tower restart** (i.e. every `local-install`) a clean exit would have relaunched into the conversation the user just quit — the headline guarantee silently broken on the most common surface, with every fresh-path test still passing. Fixed in `4b7db2d6`, with the regression test codex noted was missing.

My first attempt at that test asserted on a hand-built object literal and never called `reconnectSession` — it would not have caught the bug it was written for. Rewritten against a real `ShellperProcess` over a real socket.

Two rounds, two APPROVEs each sitting alongside one specific reproducible codex finding, and both findings real.

**Incidental discovery while fixing it:** several pre-existing reconnect tests guard with `if (client)` and pass `Date.now()` as the process start time, which never matches the real one — so those test bodies silently never execute. Not touched here; worth its own issue.

### DIFF
--- a/codev/projects/bugfix-1264-regression-3-2-4-double-ctrl-c/status.yaml
+++ b/codev/projects/bugfix-1264-regression-3-2-4-double-ctrl-c/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1264
 title: regression-3-2-4-double-ctrl-c
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-27T00:49:40.123Z'
-updated_at: '2026-07-27T00:49:40.124Z'
+updated_at: '2026-07-27T00:56:30.983Z'
--- a/codev/projects/bugfix-1264-regression-3-2-4-double-ctrl-c/status.yaml
+++ b/codev/projects/bugfix-1264-regression-3-2-4-double-ctrl-c/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1264
+title: regression-3-2-4-double-ctrl-c
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-07-27T00:49:40.123Z'
+updated_at: '2026-07-27T00:49:40.124Z'
--- a/codev/projects/bugfix-1264-regression-3-2-4-double-ctrl-c/status.yaml
+++ b/codev/projects/bugfix-1264-regression-3-2-4-double-ctrl-c/status.yaml
@@ -6,11 +6,12 @@ plan_phases: []
 current_plan_phase: null
 gates:
   pr:
-    status: pending
+    status: approved
     requested_at: '2026-07-27T01:14:29.680Z'
+    approved_at: '2026-07-27T07:20:57.861Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-27T00:49:40.123Z'
-updated_at: '2026-07-27T01:14:29.680Z'
-pr_ready_for_human: true
+updated_at: '2026-07-27T07:20:57.862Z'
+pr_ready_for_human: false
--- a/codev/projects/bugfix-1264-regression-3-2-4-double-ctrl-c/status.yaml
+++ b/codev/projects/bugfix-1264-regression-3-2-4-double-ctrl-c/status.yaml
@@ -1,14 +1,16 @@
 id: bugfix-1264
 title: regression-3-2-4-double-ctrl-c
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-07-27T01:14:29.680Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-07-27T00:49:40.123Z'
-updated_at: '2026-07-27T00:56:30.983Z'
+updated_at: '2026-07-27T01:14:29.680Z'
+pr_ready_for_human: true
--- a/codev/state/bugfix-1264_thread.md
+++ b/codev/state/bugfix-1264_thread.md
@@ -1,0 +1,98 @@
+# bugfix-1264 — double Ctrl-C no longer restarts the agent (3.2.4 regression)
+
+> **Superseded design.** The first implementation discriminated `^C^C` from a typed quit by
+> sniffing `0x03` in the shellper's input stream. The architect corrected the spec on 2026-07-27:
+> the problem was never the discrimination, it's that a clean harness exit kills the shellper at
+> all. The branch was force-pushed to carry only the corrected design, so the superseded commits
+> (`ed10e2fc`…`3d525114`) are no longer on the branch — they remain reachable by SHA via the PR's
+> force-push record. Their reasoning is kept below, because the empirical findings are what justify
+> the corrected design.
+
+## Investigate
+
+**Empirical probes against the real Claude CLI (v2.1.220), node-pty, trusted cwd.**
+
+| Probe | Result |
+|---|---|
+| Double Ctrl-C in the REPL | `{exitCode: 0, signal: 0}` |
+| Typed `/quit` | `{exitCode: 0, signal: 0}` |
+| Generated builder launch loop under ^C^C | bash **survives**, prints "Press Enter to relaunch" |
+
+Exit status cannot discriminate the two gestures. (A first probe in a fresh temp cwd returned code 1
+— it hit the trust-folder prompt and never reached the REPL. Discarded.)
+
+**Why bash survives**: Claude puts the tty in raw mode (ISIG off), so `^C` is never turned into
+SIGINT — it arrives as a literal `0x03` byte. That is also why the shellper *could* see it.
+
+### Blast radius
+
+`restartOnExit: true` is passed **only** for architect sessions (`tower-instances.ts:597`, `:1092`).
+Builders and shells use `defaultSessionOptions()` and rely on the in-PTY bash loop.
+
+- **Architect terminal** — ^C^C → session dropped, row cleared, dead, needs `afx workspace start`.
+  The reported symptom.
+- **Builder terminal** — bash loop Enter-gates the relaunch. Not dead.
+
+## Superseded attempt (kept for the record)
+
+Tracked `0x03` in `ShellperProcess.handleData`, flagged the EXIT frame, `isDeliberateExit` treated a
+flagged exit as restartable. CMAP: gemini APPROVE, claude APPROVE, **codex REQUEST_CHANGES** — and
+codex was right: my clear-the-stamp rule fired only on printable input, so **Ctrl-D after Ctrl-C**
+would have restarted an EOF quit. Reproduced, then fixed by collapsing the rule to "the stamp
+survives iff the most recent input byte is 0x03" (simpler and strictly more correct, net −11 lines).
+
+Lesson worth keeping: two APPROVEs did not outvote one specific, reproducible finding.
+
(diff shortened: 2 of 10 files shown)


waleedkadous:
## CMAP review

| Model | Verdict |
|---|---|
| gemini | APPROVE — "no issues" |
| claude | APPROVE — "no issues" |
| codex | **REQUEST_CHANGES** — addressed in 3d525114 |

### Codex's finding — real, and worth the round trip

> In raw mode, a deliberate EOF exit (`Ctrl-D`, `0x04`) after using `Ctrl-C` to interrupt work within the 10s window will still be marked `ctrlC: true`.

Correct. My clear-the-stamp rule was "clears on printable input, CR, or LF", and `0x04` is none of those. Reproduced the rule in isolation before changing anything:

```
after ^C then ^D, ctrlC stamp set = true  -> exit would be flagged ctrlC and RESTART
```

The mistake was reasoning about *printability* when the actual question is *recency* — a quit gesture doesn't have to be printable.

### Fix

The rule collapses to its correct form: **the stamp survives iff the most recent input byte is `0x03`**.

```ts
private noteInput(payload: Buffer): void {
  if (payload.length === 0) return;
  this.lastCtrlCAt = payload[payload.length - 1] === CTRL_C_BYTE ? Date.now() : null;
}
```

Simpler *and* strictly more correct — the `isRealInput` helper and the `lastIndexOf`/`subarray` scan both disappear (net −11 lines). Every previously passing case still holds, because the post-chunk state only ever depended on the last byte anyway.

Added regression coverage for `^C`→`^D`, and for `^C` superseded by Ctrl-Z / Tab / bare Escape.

**Re-verified end-to-end after the change** (real SessionManager → real shellper → real `claude`): `^C^C` still restarts, `/quit` still ends cleanly. Full suite **3769 passed, 0 failed**.

### On the split verdict

Two APPROVEs with "no issues" didn't outvote one specific, reproducible finding — codex was the only reviewer that traced the byte-level rule against a gesture the tests didn't cover, which is the same class of gap (an uncovered recovery gesture) that let #1264 ship in the first place.

waleedkadous:
## CMAP round 2 (reworked design)

| Model | Verdict |
|---|---|
| gemini | APPROVE — "no issues" |
| claude | APPROVE — "no issues" |
| codex | **REQUEST_CHANGES** — addressed in `4b7db2d6` |

### Codex's finding — the most consequential of either round

> `reconnectSession()` accepts `restartOptions.freshLaunch` in its type, but never copies it into `session.options`.

Verified against the source: the reconnect path assembled `options` field-by-field and simply omitted the new one.

The impact is worse than the phrasing suggests. A reconnected architect's args **already carry `--resume`** (baked by the #832 restart resolution), so after *any* Tower restart — which is every `local-install` — a clean exit would have relaunched straight back into the conversation the user had just quit. That is the headline guarantee of this PR, silently broken on the most common surface in the system, while every fresh-path test stayed green.

Fixed, plus the regression test codex correctly noted was missing.

**My first attempt at that test was bad and I want it on the record:** it asserted on a hand-built object literal and never called `reconnectSession` at all — it would have passed against the buggy code. Rewritten to drive the real `reconnectSession` over a real `ShellperProcess` socket, then confirmed it fails without the one-line fix.

### Also in this commit — the architect's ruling on the fast-exit valve

Keep it, with the give-up made **loud**: a new `session-gave-up` event carries a plain-language reason that Tower writes into the terminal (`PtySession.notice`) as well as the log. Deliberately not `session-error`, which also fires for transient client errors that don't end the session. Valve test coverage added per the same ruling. Both are now first-class in the PR body's design notes rather than a footnote.

### Verification after the fixes

- E2E on the real stack: `^C^C` → new `--session-id`, no `--resume`, session live, agent back. SIGKILL → argv unchanged, recovery preserved.
- Full suite **3759 passed, 0 failed**, 48 skipped; build green.

### Incidental discovery

Several pre-existing reconnect tests guard with `if (client)` and pass `Date.now()` as the process start time, which never matches the real one — so those test bodies silently never execute. Untouched here; worth its own issue.

### Note on the split verdicts

Two rounds, and in both a pair of APPROVEs sat alongside a single specific, reproducible codex finding — both of which were real defects. Worth remembering when weighing a 2-1 CMAP.

PR #1302, opened 2026-07-30
Fix send-integration e2e flake: teardown budget asymmetry (10s afterAll vs 120s beforeAll)

waleedkadous:
## Problem

`send-integration.e2e.test.ts` fails intermittently in the Tower Integration Tests CI job with `Error: Hook timed out in 10000ms` — **after all 5 tests pass**. It has hit three PRs in two days (#1283, #1290, #1301) plus multiple earlier branches, each time requiring a manual rerun.

## Root cause

(Credit: the air-1288 builder's diagnosis on PR #1301.) The suite's `afterAll` declares an explicit `10_000`ms budget, overriding the config's 300s `hookTimeout`, while `beforeAll` in the same suite gets `120_000` — an unintentional asymmetry. The teardown performs two unbounded deactivate fetches, `stopServer`'s 2s SIGTERM wait, and three `rmSync` calls; on a loaded ubuntu runner that exceeds 10s.

## Fix

- Raise the `afterAll` budget to `120_000`, matching `beforeAll`.
- Bound the two deactivate fetches with `AbortSignal.timeout(10_000)` so a hung Tower can't consume the enlarged budget either — the teardown still completes and the existing `.catch(() => {})` handles the abort.

## Testing

Suite passes locally under the e2e config: 5/5 in 1.8s. No production code touched — one test file, +6/−3.

### DIFF
--- a/packages/codev/src/agent-farm/__tests__/send-integration.e2e.test.ts
+++ b/packages/codev/src/agent-farm/__tests__/send-integration.e2e.test.ts
@@ -254,12 +254,12 @@ describe('send integration (POST /api/send → /ws/messages)', () => {
     // Guard against setup failure — variables may be undefined
     if (workspaceA) {
       const encA = encodeWorkspacePath(workspaceA);
-      await fetch(`http://localhost:${TEST_TOWER_PORT}/api/workspaces/${encA}/deactivate`, { method: 'POST' }).catch(() => {});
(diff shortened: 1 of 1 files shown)


waleedkadous:
Additional evidence from PR #1301's branch (air-1288 builder's tally): the send-integration teardown failed **3 of 4 first-attempt CI runs** on that branch alone, passing on both re-runs. That hit rate is well past 'occasional flake' on loaded runners and supports the full fix (budget parity with beforeAll + bounded fetches) over a smaller bump.

Issue #1119, opened 2026-06-29
Replace hot-context managed block in CLAUDE.md/AGENTS.md with @import lines (single source of truth, kills drift footgun)


waleedkadous:
## Summary

Replace the materialized hot-context **managed block** in `CLAUDE.md` / `AGENTS.md` with Claude Code **`@import` lines** pointing at the hot files. Today the block inlines a *verbatim copy* of `codev/resources/arch-critical.md` + `lessons-critical.md` into both root docs — three copies of the same content (source + CLAUDE.md + AGENTS.md), a DRY violation, with a real drift footgun: edit a hot file, forget `codev update`, and the committed block silently goes stale (no test/`doctor` guard catches it).

Claude Code expands `@path` imports into context at session launch (transcluded, always-on — *not* lazy), so two import lines give the same always-on guarantee with a single source of truth and no copy to drift.

**Decision is locked (maintainer):** use `@import`. Do NOT add a drift-check, and do NOT worry about whether non-Claude AGENTS.md tools expand `@import` — out of scope.

## Prescribed implementation

1. **`lib/managed-block.ts` → `renderHotContextBlock`:** emit the two import lines instead of inlining file contents:
   ```
   @codev/resources/arch-critical.md
   @codev/resources/lessons-critical.md
   ```
   - **Keep the existing `<!-- BEGIN/END CODEV HOT CONTEXT -->` markers** around them. This is deliberate: it preserves the idempotent `upsertHotContextBlock`, so an existing adopter's *old inlined block* is auto-migrated to the `@import` form on their next `codev init`/`codev update`.
   - Keep the "generated — do not edit; edit the hot files" comment + the short intro lines.
   - Keep returning `''` when neither hot file resolves (callers leave the doc untouched).
   - Only import a given file if it resolves (don't emit an `@import` for a hot file that isn't present).
2. **Keep `copyHotTierDefaults`** materializing the hot files into `codev/resources/` on init/adopt/update, and keep them user-data-protected — the `@import` only resolves if the file is on disk at that path. Paths are relative to the importing doc's dir (repo root), so `@codev/resources/...` is correct for both our repo and adopters.
3. **Do NOT change porch injection.** `commands/porch/prompts.ts` (`readHotTierFiles` → inline) must keep inlining the **live** hot content: porch prompts go straight to the model (consult/agent SDK), which does NOT expand `@import`. That path is already DRY-safe (reads source live) — leave it.
[...]

PR #1319, opened 2026-08-01
[Spec 1280] PR-1: correct the prompt-surface measurement instrument (no prompt surface touched)

waleedkadous:
## PR-1 of Spec 1280 — the corrected measurement instrument, landing early

**No prompt-surface file is touched by this PR.** That is the point: the instrument must be
correct and public *before* anything is scored by it. Spec 1280 criterion **M0b**, precedent
#1290.

### Why the old instrument couldn't score this project

`scripts/measure-prompt-surface.sh` (Spec 1252) was committed, deterministic, and wrong:

| # | Defect | Effect |
|---|---|---|
| 1 | Derived the phase-task term from `codev-skeleton/porch/prompts/` — a dead Ralph-SPIR-era tree with **no runtime consumer** — while `loadPromptFile` (`commands/porch/prompts.ts:78`) loads `protocols/<p>/prompts/` | Real SPIR phase prompts average 1,396 words; the dead tree averaged 400. **The metric was structurally blind to the largest always-on surface in the system** — cutting those prompts would have moved it by zero |
| 2 | Omitted `roles/builder.md` (1,837w), which `spawn-worktree.ts:854` injects into **every** builder spawn | 1,837 always-on words uncounted |
| 3 | Its own comment asserted `CLAUDE.md` "already inlines" the hot tier. Since #1119 (`managed-block.ts`) it carries `@import` lines, which are **transcluded** at session launch | `wc -w CLAUDE.md` excluded 736 always-loaded words |

**Reported baseline: 21,702 → 34,235.**

### A fourth inaccuracy, found while writing the tests — and it was mine

Include expansion was *additive*: it counted a `{{> path}}` directive's own tokens **plus** the
content substituted for them (~2 words per include). `expand_text` now performs real
substitution mirroring `lib/skeleton.ts resolveCodevIncludes` — regex replace in place,
recursive, depth-guarded, unresolved → empty.

This is why the figure is **34,235** and not the **34,255** quoted in the approved spec: a
20-word delta across 10 iterations. Size is reporting-only under the charter amendment
(2026-08-01), so no acceptance criterion moves — but the spec's number is superseded, and
`1280-word-baseline.md` says so in its header rather than quietly disagreeing.

### What's new in the instrument

- **Per-file four-tier resolution** (`.codev` → `codev` → skeleton), replacing two-tier
  *directory-level* selection that mis-measured single-file overrides
- **Exclusive buckets** (which sum) reported separately from **derived audience loads** (which
  overlap by design) — never presented as a sum
- **`TOTAL_AUTHORED`** on a deliberately different basis (physical files, no dedup, no
  expansion) across **all four skill trees**, so relocation can never be reported as deletion
- Protocols enumerated **from disk**, both trees unioned — including `release`, which is
  `codev/`-only and was missed by every earlier inventory

### Frozen pre-rewrite capability inventory (M5)

`codev/resources/1280-capability-inventory.json` — **57 capabilities, 47 present in served
prompts, 10 absent**. The 10 are pre-existing: porch delivers those gates and checks via task
JSON rather than authored prompt text. That asymmetry is captured as baseline state, not
something the rewrite must invent — and it confirms the soft-mode concern the plan raised for P6
is already real.

The inventory is extracted **over served prompt text, not `protocol.json`**. Extracting from
unchanged config would report every capability present even if every instruction vanished from
the prompts. Representation accepts a resolvable `{{> }}` include, so a conformant P6 rewrite
cannot fail M5.

**A `pipefail` bug in the extractor initially reported all 57 as absent**: `grep -q` exits on
first match, `printf` takes SIGPIPE, and the pipeline reports failure *because the match
succeeded*. Fixed with a here-string.

### Tests — the script had none, which is how three defects survived in a "committed and reproducible" artifact

22 passing: **T1** (sources the served directory, asserted against the real loader), **T1b**
(per-file four-tier, fixture with one overridden prompt), **T2** (phantom-savings: moving text
into an include is exactly neutral; non-markdown includes expand), **T3** (report completeness),
**T11** (buckets vs audience loads), **T12** (determinism), **T15** (relocation visible, never
reported as deletion), **T16** (manifest completeness — written now, before the first manifest
exists, because the guard must predate what it guards).

Full suite: **203 files, 4,063 tests, 0 failures.**

### Record correction

`1252-word-baseline.md` and `1252-word-after-phase7.md` are annotated **in place** — originals
preserved unaltered, marked superseded, with the reason and a pointer. They are shared knowledge
other work reads; the record gets corrected while this project builds, not after.
`1252-behavior-baseline.md` is **unaffected** (different instrument) and remains the valid
"before" for M8.

### Review note

Per **M11**, this batch is: 2 scripts · 2 test files · 3 artifacts · the manifest format. The
spec and plan artifacts also ride along (both gate-approved) so `main` carries the approved
documents early.


amrmelsayed:
## CI failure on this merge — diagnosed: runner-speed flake, not a regression

The `Tests` workflow failed on this PR's merge commit: `spec-1280-measurement-instrument.test.ts > honours PHASE_ITERS as a comparison constant` — **timed out at vitest's 5s default**.

Verified locally against the same main HEAD: all 20 tests pass, file total 23.5s. That test invokes the measurement script twice (`PHASE_ITERS: '1'` then `'2'`); on the shared CI runner the file ran ~26% slower than locally (29.6s), and those two sequential script invocations under parallel-suite load crossed 5s.

Suggested one-line fix for a follow-up (or PR-2 of the spec): give the script-invoking tests an explicit timeout budget, e.g. `it('honours PHASE_ITERS…', { timeout: 30_000 }, …)` — or a file-level `describe`-scoped timeout, since every test in this file shells out to the instrument and any of them could flake the same way on a slow runner.

Leaving the fix to the spec lane rather than hot-patching main. (Diagnosed from the main-architect session after the CI cron alert.)

amrmelsayed:
Rerun of the failed `Tests` job passed with zero changes — flake diagnosis confirmed, main is green again. The explicit-timeout suggestion above stands as the durable fix so the cron doesn't page on the next slow runner.

PR #1325, opened 2026-08-01
[Bugfix #1323] Sandbox agy + metrics DB for every test suite

### LINKED ISSUE #1323: Test-suite consult runs escape isolation: real agy spawns (login-window burst) + user-global metrics DB pollution
## Observed (2026-08-01, ~05:56-05:59 local)

A builder (aspir-1286) ran the unit + CLI-integration suites plus manual real-agy verification. Within the same two-minute window the user got a burst of **dozens of agy login browser windows** — the #1077 forkbomb, back through a path the #1250 fix explicitly does not cover.

Evidence from the user-global consult metrics DB (`consult stats --json`):
- 8 gemini + 2 codex invocations timestamped within ~1.5s (12:58:57-58Z), all 0.0s duration
- `workspace_path` values are throwaway test dirs: `/var/folders/.../codex-test-mmxOf7`, `.../codev-consult-test-1785589137328`
- one gemini row carries `error_message: "agy timed out producing the review"` — a real agy spawn attempt from a test process
- An identical fingerprint exists from 2026-07-25 06:11 (8× gemini 0.0s + 2× codex 0.0s), so this happens on **every** suite run that exercises these paths; it only becomes visible when agy happens to want login (agy updated to 1.1.9 on 07-31, which triggered re-auth)

This also matches the earlier report on #1077 of a tab burst while running `pnpm --filter @cluesmith/codev test:e2e:cli`.

## Why the #1250 fix doesn't help here

Two deliberate design points of the auth cache compose badly in the test path:

1. `agyAuthCacheDisabled()` returns true under `VITEST` unless `CODEV_AGY_AUTH_CACHE_DIR` is set — correct for keeping unit tests off the developer's real `~/.cache/codev`, but it also means **burst protection is off** for any test that reaches the real lane. Spawned `codev consult` child processes inherit `VITEST` from the runner's env, so CLI-integration/e2e children are inert too.
2. Tests that don't pin a fake agy resolve the **real** binary from PATH. Pre-flight disabled + real binary = unconditional spawn per test = one login window per spawn when agy is unauthenticated.

Separately, those same child processes record into the **real user-global metrics DB** (rows id ~35478-35487 and counting) — every suite run appends ~10 junk rows with test-temp workspace paths, silently skewing `consult stats`.

## Fix directions (for the builder to evaluate)

- The test harness (unit setup + the helpers that spawn `codev consult` subprocesses in CLI-integration/e2e) should pin `CODEV_AGY_BIN` (or equivalent) to a fake agy **unconditionally**, so no test can resolve the real binary by omission.
- Consider a belt-and-braces guard in `runAgyConsultation`: refuse to spawn a real agy when running under a test environment (`VITEST`/`CODEV_TEST`-style detection) without an explicit opt-in — misconfigured future tests should fail loudly, not open browser windows.
- Metrics isolation: same treatment as the auth cache — tests pin a metrics DB path into their temp dir (or metrics writes are inert under test env unless a path is pinned). Consider whether the existing junk rows warrant a one-off scrub.
- Builders doing *deliberate* real-agy manual verification (as aspir-1286 legitimately did) are out of scope — the issue is tests spawning real agy **by omission**.

## Acceptance criteria

- [ ] No test in unit, CLI-integration, or e2e suites can spawn the real agy binary (verify with a canary: put a logging fake at the real agy's PATH position, run full suites, assert zero invocations)
- [ ] Suite runs add zero rows to the user-global consult metrics DB
- [ ] A test that reaches the agy lane without pinning a fake bin fails loudly rather than spawning
- [ ] Existing #1250 burst/regression tests still pass unchanged

## Related
- #1077 / PR #1250 — the production-path fix (pre-flight auth cache); this issue is the test-path escape it explicitly scoped out
- #1077 comment by mohidmakhdoomi — earlier sighting via `test:e2e:cli`
- agy 1.1.9 (2026-07-31 update) triggered the re-auth state that made this visible; the isolation gap exists regardless of agy version

waleedkadous:
## Summary

Test suites could reach two pieces of user-global state by simple omission: the real `agy` binary (one OAuth browser window per spawn once agy's login lapsed) and `~/.codev/metrics.db`. Both are now sandboxed by default, with loud guards for anything that escapes the sandbox.

Fixes #1323.

## Root Cause

Two independent escapes, both failures of *omission* — safety depended on each test remembering to opt out.

**1. Real agy spawn.** None of the three vitest configs declared `setupFiles` — there was no global test harness at all. `resolveAgyBin()` falls back to `~/.local/bin/agy` and then PATH when `CODEV_AGY_BIN` is unset, so any test reaching the gemini lane without pinning resolved the developer's real binary. `agy-integration.e2e.test.ts` did exactly this by design (it deliberately does not mock `node:child_process`). The #1250 burst protection could not help: `agyAuthCacheDisabled()` turns itself off under `VITEST` unless a cache dir is named, and spawned `consult` children inherit `VITEST` via `setupCliEnv`'s `{ ...process.env }`.

**2. Metrics pollution.** `MetricsDB`'s path was hardcoded to `join(homedir(), '.codev', 'metrics.db')` with no override, so in-process tests running under the real `HOME` appended junk rows with temp-dir workspace paths.

Two corrections to the issue's reading, both verified against the code:

- The 0.0s gemini rows in `consult stats` — including the one carrying `agy timed out producing the review` — come from `consult.test.ts`, which mocks `node:child_process` at module scope. That message is emitted by the test's own fake process, **not** a real agy spawn. So that fingerprint is evidence of escape #2; escape #1 is real but lives in the CLI-integration lane and leaves no metrics rows (its subprocesses run under a sandboxed `HOME`). Both bugs are real, they just aren't the same event.
- `codev doctor`'s `verifyAgy()` is a **second** real-spawn path into agy that the issue didn't mention. It's guarded too.

## Fix

- **`packages/codev/vitest-setup.ts`** (new), wired into all three vitest configs. Pins a fake `agy`, a sandbox auth cache, and a sandbox metrics DB for every suite. CLI children inherit the pins through the environment. Pinning the auth cache also restores #1077 burst protection under test, which was previously inert.
- **`packages/codev/src/lib/test-env.ts`** (new) — belt-and-braces guards applied at both real-spawn sites (`runAgyConsultation` and `doctor.ts:verifyAgy`). Under a test runner with no pinned `CODEV_AGY_BIN` and no opt-in, they **throw** rather than fall through to the lane's usual non-blocking skip: a misconfigured test must fail the suite, not quietly degrade to a COMMENT verdict (which would hide the misconfiguration on a machine where agy simply isn't installed).
- **`CODEV_METRICS_DB`** redirects the metrics database; under a test runner with no redirect, `MetricsDB` refuses rather than falling back. `MetricsDB.defaultPath` honours it too, so `consult stats`' existence check and its subsequent open agree on one file.
- **`CODEV_ALLOW_REAL_AGY=1`** is the single named way to opt back into the real binary.
- Fixed a `delete`-instead-of-restore antipattern in `doctor.test.ts` and `consult.test.ts` that dropped the harness pin for the rest of each file.

## Test Plan

New `test-isolation.test.ts` asserts the harness is really wired, that the lane guard throws when a test forgets to pin, and that `MetricsDB` refuses the user-global path. `spawn()` is stubbed to throw, so a regression fails an assertion instead of invoking the real binary.

**Canary** (acceptance criterion 1) — temporarily instrumented `resolveAgyBin()` to log every resolution and the `MetricsDB` constructor to log every path, then ran both suites. Instrumentation removed before commit.

- **Every** `resolveAgyBin()` call across both suites had `override=` set to a temp-dir fake. Zero calls with no override — and no override is the only path to `~/.local/bin/agy` or PATH.
- **Zero** `MetricsDB` opens against `~/.codev/metrics.db` (19 opens, all sandbox or explicit test paths).
- Unit: 205 files / 4085 tests pass. CLI-integration: 8 files / 91 tests pass.
- #1250 burst/regression tests pass unchanged.

⚠️ **Row-count delta on `~/.codev/metrics.db` is not a valid instrument on a shared machine** and I initially mis-read it as my fix failing intermittently. The DB is user-global; ~34 sibling builder worktrees and 100+ concurrent vitest processes were writing to it. The count rose by 64 between two of my runs with no test run of mine in between. Instrumenting the code path is what actually answers the question.

## Note for the reviewer: this re-enables PR #1324

The merge conflict in `agy-integration.e2e.test.ts` was resolved by **re-enabling** the suite (`describe`, not `describe.skip`), per the architect's instruction to make restoring that coverage part of this fix. #1324's stated re-enable condition — "once the suite pins a fake agy (or an equivalent no-browser guard)" — is what landed here. Flagging it explicitly since this undoes a hotfix that is only minutes old.

## Scope

618 insertions / 18 deletions, but production code is ~116 LOC across four files (`test-env.ts` +67, `metrics.ts` +34, `consult/index.ts` +8, `doctor.ts` +7). The remainder is the regression test (182), the harness (80), the builder thread (108), and docs. Single-concern and contained, so I did not escalate — flagging the raw number since it exceeds the 300-LOC guideline.

`spec-1280-measurement-instrument.test.ts` failed intermittently during this work. Not related: it shells out to `measure-prompt-surface.sh` (~2.0s per call, 2-3 calls per test, 5s default timeout) under heavy machine load. A back-to-back A/B against a clean clone of `main` had both trees pass with identical duration (23.79s vs 23.71s). Left untouched.

## Behavior change worth knowing about

**Porch benchmark (e2e).** The benchmark drives a real agent that shells out to `consult`. Its gemini lane now runs against the harness's fake agy by default, so that lane returns an instant COMMENT instead of a real review. This is a step change in benchmark trends — if you're comparing runs across this commit, that's why. Restore the old behavior with `CODEV_ALLOW_REAL_AGY=1 pnpm --filter @cluesmith/codev test:e2e`.

**Adopters.** `VITEST` is exported by *any* vitest run, including a downstream project's. A suite that deliberately shells out to `consult -m gemini` now throws until it sets `CODEV_ALLOW_REAL_AGY=1`. Intentional — a suite reaching the real agy is what this issue was about — but it's a behavior change, and the error names the variable to set. Documented in `consult.md` in both trees.

## Review response

CMAP: gemini APPROVE / codex APPROVE / claude APPROVE (all HIGH). Architect integration review requested two changes, both landed:

- The test-isolation guard moved **inside `resolveAgyBin()`**'s no-override branch — the single chokepoint covering `runAgyConsultation`, `verifyAgy`, and `checkAgy`. Guarding only the spawn sites was insufficient because the unpinned lookup runs `agyRespondsToVersion()`, which *executes* the candidate binary.
- `describe.skipIf` on `agy-integration.e2e.test.ts`, so the three opt-in cases report as **skipped** rather than as passing tests that ran nothing (CLI suite: 91 passed → 88 passed / 3 skipped).

Declined one optional item — `setupFiles` for `packages/core` and `packages/artifact-canvas` — because neither references consult, agy, or `MetricsDB`, and wiring codev's harness into them would invert the workspace dependency. Reasoning in the PR comment.

Out of scope, filed separately as #1326: `~/.agent-farm/global.db` has no env override, so the harness cannot pin it.

## Carried fix: pre-existing CI flake in `spec-1280-measurement-instrument.test.ts`

Not caused by this PR, but it blocks CI on this branch, so the fix is carried here via a merge of `origin/hotfix/1280-test-timeouts` (commit `216b7932`) rather than a duplicate change.

That file shells out to `scripts/measure-prompt-surface.sh`, which I measured at ~2.0-2.5s per invocation. Cases that call it two or three times run 4-9s against vitest's 5s default timeout — deterministically over budget on a loaded runner. The hotfix gives the 12 script-shelling cases explicit 60s budgets, including `honours PHASE_ITERS as a comparison constant` (line 303), which invokes the script twice.

Verified on this branch after the merge: 20/20 pass in 26.65s with 106 concurrent vitest processes on the machine.

### DIFF
--- a/codev-skeleton/resources/commands/consult.md
+++ b/codev-skeleton/resources/commands/consult.md
@@ -208,6 +208,25 @@ through to a normal spawn.
 | `CODEV_AGY_AUTH_CACHE_DIR` | `~/.cache/codev` | Cache location (mainly for tests). |
 | `CODEV_AGY_AUTH_CACHE_DISABLE` | unset | `1` restores the always-spawn behaviour. |
 
+#### Test isolation
+
+The cache above protects *production* runs. It did not protect test runs: the
+cache disables itself under `VITEST` unless a cache directory is named, and any
+test that reached the gemini lane without pinning `CODEV_AGY_BIN` resolved the
+real binary — so a suite run could still open a login window per spawn (#1323).
+Consult now treats a test runner as a hard boundary. Under `VITEST` (or
+`CODEV_TEST`) it refuses to resolve an unpinned `agy` and refuses to open the
+user-global metrics database, failing loudly instead of reaching either. Codev's
+own suites pin a fake `agy`, a sandbox auth cache, and a sandbox metrics DB for
+every test; spawned `codev` / `consult` children inherit the pins through the
+environment.
+
+| Variable | Default | Purpose |
+|---|---|---|
+| `CODEV_AGY_BIN` | auto-resolved | Pin the agy binary. The harness sets this to a fake for every test. |
+| `CODEV_ALLOW_REAL_AGY` | unset | `1` opts a test run into the REAL agy binary — the guarded integration smoke and real-AI e2e runs. Expect a browser window if agy's login has lapsed. |
+| `CODEV_METRICS_DB` | `~/.codev/metrics.db` | Redirect the consult metrics database. Required under a test runner; the harness points it at a temp dir so suite runs stop skewing `consult stats`. |
+
 ### Claude auth: subscription vs. metered API
 
 `consult -m claude` runs on the Claude Agent SDK. When `CLAUDE_CODE_OAUTH_TOKEN`
--- a/codev-skeleton/resources/commands/consult.md
+++ b/codev-skeleton/resources/commands/consult.md
@@ -215,7 +215,7 @@ cache disables itself under `VITEST` unless a cache directory is named, and any
 test that reached the gemini lane without pinning `CODEV_AGY_BIN` resolved the
 real binary — so a suite run could still open a login window per spawn (#1323).
 Consult now treats a test runner as a hard boundary. Under `VITEST` (or
-`CODEV_TEST`) it refuses to resolve an unpinned `agy` and refuses to open the
+`CODEV_TEST_ISOLATION`) it refuses to resolve an unpinned `agy` and refuses to open the
 user-global metrics database, failing loudly instead of reaching either. Codev's
 own suites pin a fake `agy`, a sandbox auth cache, and a sandbox metrics DB for
 every test; spawned `codev` / `consult` children inherit the pins through the
--- a/codev-skeleton/resources/commands/consult.md
+++ b/codev-skeleton/resources/commands/consult.md
@@ -226,6 +226,14 @@ environment.
 | `CODEV_AGY_BIN` | auto-resolved | Pin the agy binary. The harness sets this to a fake for every test. |
 | `CODEV_ALLOW_REAL_AGY` | unset | `1` opts a test run into the REAL agy binary — the guarded integration smoke and real-AI e2e runs. Expect a browser window if agy's login has lapsed. |
 | `CODEV_METRICS_DB` | `~/.codev/metrics.db` | Redirect the consult metrics database. Required under a test runner; the harness points it at a temp dir so suite runs stop skewing `consult stats`. |
+| `CODEV_TEST_ISOLATION` | unset | `1` applies the same guards to a harness that is not vitest. |
+
+**Adopters:** `VITEST` is exported by *any* vitest run, including yours. If your
+project's own suite deliberately shells out to `consult -m gemini`, that call now
+throws until you set `CODEV_ALLOW_REAL_AGY=1` (and `CODEV_METRICS_DB` if you want
+the metrics recorded somewhere). This is intentional — a suite reaching the real
+agy is what #1323 was about — but it is a behaviour change, not a silent one: the
+error names the variable to set.
 
 ### Claude auth: subscription vs. metered API
 
--- a/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
+++ b/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1323
 title: test-suite-consult-runs-escape
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-01T13:07:26.835Z'
-updated_at: '2026-08-01T13:07:26.836Z'
+updated_at: '2026-08-01T13:10:51.247Z'
--- a/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
+++ b/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1323
 title: test-suite-consult-runs-escape
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-01T13:07:26.835Z'
-updated_at: '2026-08-01T13:10:51.247Z'
+updated_at: '2026-08-01T13:44:59.961Z'
--- a/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
+++ b/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
@@ -6,11 +6,12 @@ plan_phases: []
 current_plan_phase: null
 gates:
   pr:
-    status: pending
+    status: approved
     requested_at: '2026-08-01T13:45:05.688Z'
+    approved_at: '2026-08-01T17:11:00.829Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-01T13:07:26.835Z'
-updated_at: '2026-08-01T13:45:05.689Z'
-pr_ready_for_human: true
+updated_at: '2026-08-01T17:11:00.830Z'
+pr_ready_for_human: false
--- a/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
+++ b/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1323
+title: test-suite-consult-runs-escape
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-08-01T13:07:26.835Z'
+updated_at: '2026-08-01T13:07:26.836Z'
--- a/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
+++ b/codev/projects/bugfix-1323-test-suite-consult-runs-escape/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-08-01T13:45:05.688Z'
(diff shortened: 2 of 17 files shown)


waleedkadous:

Really strong work — the harness inversion (sandbox by default, opt-in by name) is the right architecture, your canary methodology was more rigorous than the issue asked for, and both your corrections to my issue text were accurate. My two integration lanes (codex, claude) both flag the same two items, and I verified each against the PR head myself, so I'd like to see these before merge:

1. **Guard ordering in `doctor.ts`** — `verifyAgy()` calls `resolveAgyBin()` before `assertAgyLaneAllowedUnderTest()`, and `resolveAgyBin` can `execSync` the real binary (`agyRespondsToVersion`). So an unpinned test still executes real agy — no browser window from `--version`, but it violates your own acceptance criterion. Move the assert above the resolve (the consult lane already has this ordering right). While there, check `checkAgy()` — claude's lane reports it reaches resolution unguarded too.

2. **Make the opt-in skips visible** — the three re-enabled cases early-return without the opt-in, which reports as three *passing* tests that executed nothing. `describe.skipIf(!realAgyOptIn())` (or per-case `it.skipIf`) makes the state honest in every test report, and turns this into what #1324 asked for: skipped, but for a named, machine-checked reason.

Three smaller ones, cheap while you're in there — take or leave with a note:

3. `packages/core` and `packages/artifact-canvas` vitest runs are `VITEST`-tainted but have no harness `setupFiles` — the first test there that shells into consult gets a confusing hard throw. Wiring the same setup (or a minimal pin) pre-empts it.
4. `metrics.test.ts:512` now passes only because the sandbox deliberately mirrors the `.codev/metrics.db` layout — assert against `process.env.CODEV_METRICS_DB` instead so the coupling is explicit and the test name stops lying.
5. A short note in `test-env.ts` on *why* the guard is per-call-site rather than inside `resolveAgyBin` (the resolution tests need the unpinned path) — you found a second spawn site the issue missed; leave the next person a map.

FYI, handled outside your PR: I'm filing the `~/.agent-farm/global.db` gap (no env override on `AGENT_FARM_DIR`, so the harness can't pin it) as a follow-up issue rather than growing your scope. The porch-benchmark step change (gemini lane now instant-COMMENT under fake agy) is worth a line in your PR body so future readers of benchmark trends aren't confused.

waleedkadous:
## Review response — both required items landed

**1. Guard ordering in `doctor.ts` (incl. `checkAgy`)** — done, but taken further than moving the assert above the resolve. The guard now lives **inside `resolveAgyBin()`**, in the no-override branch, which is the single chokepoint every route passes through. That covers `runAgyConsultation`, `verifyAgy`, **and** `checkAgy` in one place, and it closes the hole exactly as you and the claude lane described it: resolution isn't passive — the unpinned lookup runs `agyRespondsToVersion()`, which *executes* the candidate with `--version` — so guarding only the spawn sites still let a suite run the real binary. The now-redundant per-site asserts were removed.

**2. `skipIf`** — done. `describe.skipIf(!REAL_AGY)`, with `REAL_AGY` resolved once at module scope and short-circuited on the opt-in so `resolveAgyBin()` is never called against the harness pin. Verified in the reporting, which was the whole point:

| CLI-integration suite | before | after |
|---|---|---|
| files | 8 passed | 7 passed, **1 skipped** |
| tests | 91 passed | 88 passed, **3 skipped** |

**4. `metrics.test.ts:512`** — done. Asserts against `process.env.CODEV_METRICS_DB` directly, plus a second case pinning the refusal-under-test behaviour. The old name claimed something it had stopped verifying.

**5. Map note in `test-env.ts`** — added, with one correction. Your note suggested documenting why the guard is per-call-site "because the resolution tests need the unpinned path." That premise doesn't hold: every resolution test pins `CODEV_AGY_BIN` before calling (they're testing override handling), and `agy-integration` short-circuits on the opt-in. Since nothing needed the unpinned path, the guard moved to the resolver and the note documents the three routes into agy instead — including the `checkAgy` one that doesn't spawn but still executes.

**3. `packages/core` / `artifact-canvas` setupFiles — declined, with reason.** Neither package references consult, agy, or `MetricsDB`; the only grep hit is the word "consult" inside a code comment. Wiring codev's harness into them would invert the workspace dependency (core is a dependency *of* codev, not the reverse), and duplicating the pin logic adds maintenance surface for a hypothetical. If one ever does shell into consult, the guard's error names the variable to set. Happy to add it if you'd rather have it pre-emptively.

**Porch-benchmark note** added to the PR body as you asked.

### Verification after these changes
- Unit: 205 files / **4086 pass**, 0 fail
- CLI-integration: 7 pass + 1 skipped / 88 pass + 3 skipped, 0 fail
- Not executed: the e2e suite (\~$4/40min real-AI porch benchmark). Same `setupFiles` is wired into its config, so the pin applies by construction, but that path is verified by inspection rather than execution.

Ready for the `pr` gate whenever the human is.

PR #1327, opened 2026-08-01
fix(test): stop pinning live-measured totals; assert instrument correctness three ways

waleedkadous:
## Stop pinning live-measured totals in the instrument tests

Routed from the architect after **Spec 1307** hit this on day one.

### The problem with the current form

The reproduction tests assert `ALWAYS_ON_WORDS === 34231` against the **real repo**, so *every
always-on edit by every project* fires them. 1307 handled it exactly right — causal check, then
a justified bump with the derivation preserved.

But the incentive it creates for the *next* project is **bump-without-checking**, which is the
regression the test exists to catch. The number moves for two different reasons — the instrument
broke, or the surface changed — and the test cannot distinguish them, so it delegates that
judgement to whoever is least placed to spend time on it.

### Three layers replace it

None can be disturbed by a legitimate surface edit:

1. **Invariants over the live repo** — the composition arithmetic, true at any surface size:
   `ALWAYS_ON = SHARED + BUILDER_SPAWN[spir] + I × (HOT + PHASE mean)`, `architect = SHARED +
   ARCHITECT`, and `PHASE_ITERS` linearity.
2. **Absolute values on a synthetic fixture** whose arithmetic a human can check by hand —
   `SHARED 7`, `SPAWN 10`, `ALWAYS_ON 107`. This pins the **instrument's** correctness without
   pinning the **repo's** content.
3. **Byte-assertion on the frozen baseline artifact** — a historical record; editing it should
   be a deliberate act.

### Why layer 2 isn't redundant — verified by mutation, not argued

Invariants have a blind spot: a component that is silently wrong still satisfies every internal
identity, because the wrong value propagates consistently.

Mutating the script to drop `HOT` from `SHARED` — a real instrument bug:

| Layer | Result under mutation |
|---|---|
| 1 — invariants | **all three PASS** |
| 2 — fixture absolutes | **all three FAIL** |

Invariants alone would have shipped it. (The pre-existing T1 structural test also fires here,
but only because it pattern-matches the `SHARED=` source line — it would not catch a component
bug that resolves the wrong file.)

### Verified in the other direction too

Adding 44 words to `CLAUDE.md` — the exact edit shape that fired the old assertion — keeps all
**24 tests green**. No unrelated builder is asked to bump a number they don't own.

### Where live numbers go instead

Into the phase manifests and generated artifacts, where a human reads them as **findings**
rather than maintaining them as **expectations**.

### Note

Branched off `d42a061a`. 1307's interim bump (34293/8661) is **not** on `origin/main` as of this
branch point — `main` still reads `toBe(34231)` — so there is no conflict to resolve here. If
that bump lands separately it will conflict with this file; happy to rebase.

### DIFF
--- a/packages/codev/src/__tests__/spec-1280-measurement-instrument.test.ts
+++ b/packages/codev/src/__tests__/spec-1280-measurement-instrument.test.ts
@@ -130,7 +130,7 @@ describe('T1b — per-file four-tier resolution, not directory-level selection',
     const out = run(dir);
     // Directory-level selection would have missed the override entirely.
     expect(out).toMatch(/\| spir \| \d+ \| 8 \| \d+ \|/);
-  });
+  }, 60_000);
 
   it('prefers .codev/ over codev/ over codev-skeleton/', () => {
     const src = fs.readFileSync(script, 'utf-8');
@@ -154,7 +154,7 @@ describe('T2 — phantom-savings proof: includes are expanded', () => {
     fs.writeFileSync(path.join(dir, 'codev-skeleton/protocols/spir/templates/frag.md'), 'cc dd ee\n');
 
     expect(num(run(dir), 'ALWAYS_ON_WORDS')).toBe(before);
-  });
+  }, 60_000);
 
   it('expands non-markdown includes too — protocol.json delivery depends on it (P6)', () => {
     const dir = makeFixture();
@@ -169,7 +169,7 @@ describe('T2 — phantom-savings proof: includes are expanded', () => {
     );
     // The JSON's words must appear in the served count, not vanish.
     expect(num(run(dir), 'ALWAYS_ON_WORDS')).toBeGreaterThan(before);
-  });
+  }, 60_000);
 });
 
 describe('T3 — per-surface reporting completeness (not a ceiling)', () => {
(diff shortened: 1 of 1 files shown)


waleedkadous:
## Merged #1321's timeout fix into this branch — it was carrying the same latent red

Self-review catch before this sat in the merge queue: this PR branched from `d42a061a`, which
**predates #1321**, so it inherited **10 unbudgeted script-shelling tests** — the exact latent
5s-default failure that took `main` red in the first place. My three new layer-1/layer-2 tests
carried their own budgets, but the ten I did not touch did not.

Merged `origin/hotfix/1280-test-timeouts` (rather than duplicating the change) so that when
#1321 lands, shared ancestry keeps the merge clean instead of conflicting on a duplicated edit.

**One conflict, resolved in favour of this PR:** the hotfix carried a budgeted copy of
`honours PHASE_ITERS as a comparison constant` — a test this PR *replaces* with
`PHASE_ITERS is a linear comparison constant` under layer 1, which carries its own budget and
no longer pins `736 + 1396`. Keeping the hotfix's copy would have reinstated a live-measured
literal, which is the thing this PR exists to remove.

Verified after resolution: **zero unbudgeted shelling tests**, zero conflict markers, 24 tests
green.

Merge order is unchanged — this still belongs behind #1321.

mohidmakhdoomi (approved):
Approving to unblock main. Test-only change to spec-1280-measurement-instrument.test.ts: gives the script-shelling measurement tests explicit 60s budgets and reworks assertions to stop pinning live-measured totals. Verified green on CI (Tests + CLI Integration). Resolves the 5000ms timeout on 'honours PHASE_ITERS as a comparison constant' that has main red; supersedes #1321.

Issue #1205, opened 2026-07-19
Shellper replay buffer stores the stream, not the screen: unbounded growth for full-screen TUIs (17MB+ replays)


amrmelsayed:
## Problem

The shellper's replay buffer is bounded by **line count** (10,000 lines) plus an unbounded whole-kept partial. A full-screen TUI (Claude sessions) redraws in place via cursor addressing and emits almost no newlines, so an entire multi-day session accumulates as *one line that never ends*: the line cap never triggers and the buffer grows monotonically for the life of the session.

Measured impact (from the #1198 incident, 2026-07-19):

- Two six-day-old architect sessions reached **17.6MB / 17.7MB** replay buffers — past the 16MB `MAX_FRAME_SIZE` wire cap. The resulting oversized REPLAY frame deterministically errored Tower's parser on every reconnect. On pre-#1198 code that swallowed error minted the original silent-zombie terminals; during the #1198 rollout it escalated to two killed sessions before the tolerant-parser fix landed.
- A third live session is at 14MB and climbing (`Terminal partial monitor: max partial 14081 KB`).
- Of those megabytes, only roughly the final screenful (a few KB) is information a reconnecting viewer needs. The rest is every superseded repaint frame the TUI ever painted: spinner ticks, token-stream updates, status-bar refreshes.

Byte-trimming the buffer from the front was deliberately rejected previously (#1047): a trim can cut mid-escape-sequence or drop the alt-screen-enter, corrupting replay rendering. So the buffer is whole-by-design, and its growth is unbounded-by-design — but the wire (16MB frame cap) and the UI (xterm parse cost per attach) are not.

## Containment already in place (PR #1204)

These make the current design *safe*, not *right*:

- Tower's frame parser discards oversized frames (`'frame-skipped'`) instead of treating them as fatal stream errors.
- New shellpers truncate the replay they send to an 8MB tail (`REPLAY_PAYLOAD_MAX`).
- Adoption seeds at most a 1MB tail into the ring buffer (`RING_SEED_MAX_BYTES`); viewers rely on the post-connect resize nudge to repaint TUIs.

Every cap is a lossy tail-cut that can transiently render rough until the nudge repaints. The buffer itself still grows without bound in shellper memory.

## Proposed direction

Make replay **O(screen), not O(session lifetime)**: store something screen-shaped rather than stream-shaped.

[...]

PR #1330, opened 2026-08-01
[Spec 1313] afx send: mailbox-first delivery (never force-inject)

### LINKED ISSUE #1265: afx send can corrupt in-progress user input: idle-time is a bad proxy for line-occupancy
## Problem

An `afx send` message can land in the middle of a user's half-typed input in the receiving terminal, corrupting both. Reproduced by typing a message in the architect terminal while a builder ran `afx send` — the builder's message was appended to the in-progress input line and submitted as one blob.

## How `afx send` reaches a terminal today

There's already a "typing awareness" system (Spec 403), but it's built on the wrong signal, which is why the corruption still happens.

Pipeline (`afx send` → `POST /api/send` → `handleSend`):

1. **`packages/codev/src/agent-farm/servers/tower-routes.ts:1516`** — the deliver-vs-defer decision:
   ```ts
   const shouldDefer = !interrupt && !session.isUserIdle(sendBuffer.idleThresholdMs);
   ```
2. **`packages/codev/src/terminal/pty-session.ts:502`** — `isUserIdle` is *purely a timer*: `Date.now() - lastInputAt >= 3000`. True whenever the user hasn't pressed a key in the last **3 seconds**.
3. If deferred, `SendBuffer` (`packages/codev/src/agent-farm/servers/send-buffer.ts`) holds the message and re-checks every 500ms, flushing when the user goes idle **or** after a **60s** max age.
4. On delivery, **`packages/codev/src/agent-farm/servers/message-write.ts:47`** writes the message text, then ~50ms later writes `\r` (Enter) — which **submits whatever is on the line**, i.e. the user's half-typed input + the injected message as one command.

## Root cause

**Idle-time is a bad proxy for "is the input line empty?"** A user who types a few words then pauses >3s (reading, thinking) becomes "idle" and looks identical to someone at an empty prompt — so the message delivers immediately onto a non-empty line and Enter submits the mess.

A real line-occupancy signal *does* exist — `session.composing` (`pty-session.ts:527`), set on non-Enter keystrokes and cleared on Enter — but it was **deliberately removed** from the defer decision in bugfix #492 (comment at `tower-routes.ts:1514`) because it "got stuck true" after Ctrl+C / arrows / Tab. So today nothing checks whether the line actually has text.

## Target-app context

The default architect/builder command is `claude` (`packages/codev/src/agent-farm/servers/tower-instances.ts:491`), and Codex is the other common agent. Both are TUIs with their own line editors (Ink-based), not bash/readline — which matters for any keystroke-injection delivery tactic.

`Ctrl+U` (kill line), `Ctrl+E` (cursor to end), and `Ctrl+Y` (yank) are **confirmed working in both Claude Code and Codex**, and also work in a readline/zsh shell target. So a kill-ring–based restore is viable on all the agents in practice.

## The kill/yank delivery tactic and its remaining risks

One delivery idea: when a message arrives and the line is non-empty, run a kill/yank maneuver — `Ctrl+E` (cursor to end) → `Ctrl+U` (kill whole line into the kill-ring) → write the message → `Enter` (submit the message cleanly) → `Ctrl+Y` (yank the user's draft back onto a fresh prompt). This preserves and restores the user's input, and since Ctrl+U/E/Y work on both default agents the restore is dependable.

Two open risks remain:

1. **Multi-line drafts (confirmed limitation).** Claude Code / Codex input can be multi-line, and `Ctrl+U` clears only the *current* line, not the whole composition. So the kill/yank tactic partially destroys a multi-line draft and cannot restore it — a byte-replay approach (Option H) is required whenever the draft may span more than one line.
2. **Concurrency race.** Between `Ctrl+U` and `Ctrl+Y` (message write + Enter, ~150ms) the user may keep typing; new keystrokes interleave with the injection and the yank restores stale text on top. Mitigated by Option I (atomic single write).

## Options

Two independent axes — **detection** (whether/when to deliver) and **delivery** (how to inject non-destructively) — plus a **channel** axis (change the model so injection isn't destructive at all). Fixing detection is the root cause; the kill/yank trick is a delivery tactic.

### Detection — decide when it's safe to deliver

- **A — Fix + re-enable `composing` deferral (recommended core).** Feed a *corrected* line-occupancy signal back into `shouldDefer`: buffer while the line is non-empty, flush when it's empty/submitted. Fix the "stuck true" bug properly — clear `composing` on Ctrl+C (0x03), Ctrl+U, Escape, Ctrl+G, not just Enter. Keep the 60s max-age valve so a message can't be starved. Low risk, app-agnostic (it only *defers*). Completes what #492 tore out, done right.
- **E — Event-driven flush on submit.** The buffer currently flushes on a 500ms idle *timer*. Instead, piggyback on the natural "line just became empty" moment: the websocket handler already calls `session.stopComposing()` when it sees `\r`/`\n` (`tower-websocket.ts:96`). Turn that into an event that immediately flushes buffered messages onto the now-known-empty prompt. Near-zero corruption because delivery happens at the instant the line is provably empty. Highest-value detection change; pairs with A.
- **F — Active terminal probe (DSR).** Send `ESC[6n` to query the cursor column before delivering; a column past the prompt start means the line is non-empty. More direct than timing, but fragile — the response interleaves with app output and the prompt-end column is unknown. Lower confidence; mention-only.
- **G — PTY output-stream line modeling.** Tower already keeps a ring buffer of rendered output (`pty-session.ts` `ringBuffer`). Parse it to model the current input line. Most *accurate* occupancy signal, but the heaviest to build and keep correct across apps. Overkill unless A/E prove insufficient.

### Delivery — inject non-destructively

- **B — Kill/yank trick.** Deliver onto a non-empty line but preserve/restore via the kill-ring (see the tactic above). Restores **single-line** drafts on both default agents; it **cannot restore a multi-line draft** — `Ctrl+U` clears only the current line (confirmed) — and is still subject to the race.
- **C — Hybrid (A + B as the max-age tactic).** Buffer while composing (A). On idle/empty-line flush → plain write (safe). On the *60s max-age* flush (line still busy) → kill/yank restore (single-line only), else deliver with `noEnter` (already plumbed) so it doesn't submit garbage. Because the restore breaks on multi-line drafts, the max-age tactic is better served by H below.
- **H — Tower-side draft capture + verbatim replay.** Tower sees every user keystroke in the websocket handler. Accumulate the un-submitted bytes into a per-session draft buffer (cleared on Enter/Ctrl+C/Ctrl+U). To deliver onto a busy line: clear the line → write message + Enter → **replay the captured byte stream verbatim** to reconstruct the draft. Advantages over the kill/yank: **no dependency on the app's kill-ring**, and replaying the exact bytes (including the user's own backspaces/arrows) reproduces the exact draft — so it **handles multi-line correctly**. The robust version of B.
- **I — Atomic single-write kill/yank (race fix for B/C).** Send the entire `Ctrl+E Ctrl+U <msg> Enter Ctrl+Y` as one `write()` string so the user can't interleave keystrokes mid-sequence, shrinking the race window to ~zero. Tension: `message-write.ts` deliberately *paces* multi-line writes (bugfix #584) to dodge paste-detection swallowing the Enter, which conflicts with a single atomic write for multi-line messages. Clean for short messages; multi-line needs H or J.
- **J — Bracketed-paste framing.** Wrap the injected message in `ESC[200~ … ESC[201~`. Claude Code and Codex both honor bracketed paste: control chars in the message won't be interpreted as commands, and it sidesteps the paste-detection/Enter-swallow problem (#584) without line-by-line pacing. Doesn't solve occupancy alone — a *complement* to H or I that makes the message body itself safe.

### Channel — change the model so injection isn't destructive

- **D — Side channel (no keystroke injection).** Stop modeling inter-agent messages as synthetic typing. Large architectural change; the umbrella for K and L.
- **K — Escalate to a side channel after max-age instead of forcing.** Today at 60s max-age the buffer force-injects even onto a busy line (the destructive path). Instead, if the user is still holding a draft, post the message to the dashboard/notification (`broadcastMessage` already exists) and log it, and **don't inject keystrokes**. The draft is never corrupted; the agent gets the message on the user's next submit. A pure safety-valve.
- **L — Notification-only + agent mailbox.** Deliver to a per-agent inbox the agent reads when it next acts, rather than injecting into the input line. Cleanest, but changes the core interaction model (the point of `afx send` today is to inject into the agent's input so it acts immediately).

## Recommendation

- **Primary:** Option **A** (fixed `composing` occupancy) **+ E** (event-driven flush on submit) — this alone eliminates the common case, delivering only when the line is provably empty.
- **For the max-age case** (user holds a draft too long), use **H (byte replay)**. `Ctrl+U` in Claude Code / Codex clears only the current line, not a multi-line draft (confirmed), so the **B/C kill/yank cannot safely restore multi-line input** — reserve it for the single-line case only, if at all. Frame the message body with **J** and, if a kill/yank path is kept for single-line, harden it with **I**; or use **K** to never risk the busy line at all.

The building blocks already exist — the `composing` flag, the `SendBuffer`, the 60s max-age valve, `noEnter`, the output ring buffer, and `broadcastMessage`.

## Relevant files

- `packages/codev/src/agent-farm/servers/tower-routes.ts` (defer decision, `handleSend`)
- `packages/codev/src/agent-farm/servers/send-buffer.ts` (`SendBuffer`, idle/max-age flush)
- `packages/codev/src/agent-farm/servers/message-write.ts` (paced write + Enter)
- `packages/codev/src/terminal/pty-session.ts` (`isUserIdle`, `composing`, `ringBuffer`)
- `packages/codev/src/agent-farm/servers/tower-websocket.ts` (records input, toggles `composing`)

Related: Spec 403 (typing awareness), bugfix #450, bugfix #492 (removed the `composing` check), bugfix #584 (paced writes).

### LINKED ISSUE #1313: afx send: mailbox-first delivery — persist, gate on rendered-empty, never force-inject
## Decision (supersedes the delivery-maneuver direction of #1265 / spike 1265)

**No message is ever forced onto a busy line.** A busy line means a human is present at that terminal — the right response is to tell the human through UI, not to fight them for the keyboard with clear/restore maneuvers. This drops the 60s max-age force-inject requirement, and with it ~70% of the apparatus the spike explored (H byte-replay, kill/yank, stash, per-app delivery forms, cursor-aware draft modeling, pre-Enter equality gate, canonical-stream verify).

## Problem (from #1265)

`afx send` can land mid-draft and submit the user's half-typed input fused with the message. `shouldDefer` (`tower-routes.ts:1570`) is a 3s idle timer — a bad proxy for line occupancy. Additionally, today's `SendBuffer` is in-memory (held messages die with a Tower crash) and `stop()` force-flushes onto the line at shutdown; two concurrent sends blob into one submit (spike `w1a`); menus/trust dialogs/wrapper screens silently eat or strand deliveries.

### 1. Persistent mailbox at enqueue (spike option K, core)

Every send is persisted as a mailbox row **before** the send response returns. The in-memory `SendBuffer` queue collapses into it (fixes crash loss and the shutdown force-flush). Response vocabulary: `delivered` | `held`+id. Rows address agents, not PTYs, so a respawned terminal drains its predecessor's mail. Held-too-long escalates **visibility** (broadcast/dashboard/inbox), never injection.

### 2. Rendered-empty gate before every delivery (spike G-lite — validated)

Replay the existing output ring buffer through headless `@xterm/headless` and classify: composer marker present AND zero normal-intensity text in the composer region → safe. Measured 2–22ms per check. One mechanism answers "is a menu open?", "is the line occupied?", and "is there an agent behind this PTY at all?" — it correctly trips on slash menus, the codex model picker, agy's trust dialog, half-typed drafts (including `afx attach`-typed ones input tracking can't see), and the builder launch-loop's relaunch/boot screens. Sessions are born dirty; they converge only via this rendered proof.

Delivery moments: after a user submit, on output quiescence, and a poll backstop. Clean → deliver; not clean → keep holding. **There is no force path.**

### 3. Minimal per-session write serialization

Concurrent sends to one session currently interleave/blob (spike `w1a`, real `message-write.ts`). A small per-session FIFO with completion-based chaining fixes it. `--interrupt` stays an explicit human-invoked bypass.

### Explicitly cut (archived as spike evidence, path not taken)

- DraftTracker cursor/line modeling; per-keystroke occupancy state machines
- All busy-line delivery maneuvers: H byte-replay, I/B kill/yank, `^S` stash
- Pre-Enter equality gate and canonical-stream differential verification
- Per-app *delivery form* matrix (bracketed/unbracketed atomic forms) — only per-app *marker profiles* for the gate are needed, because delivery only ever targets an empty verified prompt

## Why this eliminates corruption by construction

Message bodies are only ever written to an empty, rendered-verified prompt. A message cannot fuse with a draft because it is never delivered while one exists; a draft cannot be destroyed because nothing ever clears it.

## Estimate

~400–700 LOC (vs. the spike's ~2,400–2,800 for the full-maneuver design). Salvage from spike 1265: the G-lite classifier approach + measured per-app marker/dim-placeholder facts, the keystroke/submit-semantics table (informs the submit-detection delivery moment), the born-dirty constraint, the wrapper-loss and attach-bypass findings (motivating the output-side gate), and the POC harness as a version-bump smoke test.

## Relevant files

- `packages/codev/src/agent-farm/servers/tower-routes.ts` (`handleSend`, `shouldDefer`)
- `packages/codev/src/agent-farm/servers/send-buffer.ts` (collapses into mailbox)
- `packages/codev/src/agent-farm/servers/message-write.ts` (write serialization)
- `packages/codev/src/terminal/pty-session.ts` (ring buffer source for the gate)
- Spike findings: `codev/spikes/1265-afx-send-line-occupancy.md` (branch `spike-1265`)

Refs #1265.

mohidmakhdoomi:
## Summary

Replaces `afx send`'s timer-based, in-memory, force-flushing delivery (`SendBuffer`, Spec 403) with a **mailbox-first** pipeline. Every message is persisted to `global.db` before the send returns, and its body is **only ever written to a prompt a headless-terminal render-gate proves is empty**. Corruption is eliminated *by construction* — a message can never fuse with a draft, menu, dialog, or wrapper screen — and **there is no force path**: no timeout, valve, or fallback ever writes onto a non-clean screen. A busy line means a human is present; the response is to escalate visibility (indicator + `afx inbox`), never to fight for the keyboard.

Closes #1313

## Changes (9 phases, single PR)

1. **Mailbox store** — durable `mailbox` table (migration v15, agent-addressed rows); kills silent loss.
2. **Render-empty gate** — `render-gate.ts` classifies a headless-terminal screen (via `@xterm/headless`); claude/codex profiles. Idle-gate cost well under the ≤ ~50ms budget.
3. **agy profile (blocking)** — net-new empirical measurement; agy's normal-intensity gray hint needed a color-keyed `placeholderFgPalette` rule (dim/bold couldn't separate idle from draft). Trust dialog classifies not-clean (no blind Enter confirms filesystem trust).
4. **Delivery orchestration** — `handleSend` rewrite: persist → serialize → gate → deliver/hold. Retires `SendBuffer` and every force path. Per-PTY write serialization (`write-queue.ts`). Response vocab `delivered | held`+reason (`busy`/`no-profile`/`no-live-pty`), surfaced client-side (single + `--all`).
5. **Fast delivery triggers** — submit + output-quiescence triggers via a single `handleUserInput` chokepoint on `PtySession`; held mail delivers near-immediately once the line clears.
6. **Cron rerouting** — the most-unguarded writer joins the one gated path; per-task supersede; honest run-log outcomes.
7. **`afx inbox` + broadcasts + escalation** — CLI (list/show/dismiss, workspace-scoped), `overview-changed` held-count SSE, `mailbox-escalation` visibility event (never delivery), liveness telemetry.
8. **Dashboard + VSCode indicators** — count-only held indicators + attention state on escalation (read-only; dismissal stays CLI-only).
9. **Docs + skeleton mirror** — send vocab, `afx inbox`, mailbox config; skeleton mirrored; the mailbox-first invariant added to the `arch-critical.md` hot tier.

## Post-gate hardening (folded into this PR)

After the first pr-gate approval, live testing on installed code surfaced real defects; the architect authorized a **verify→implement rollback** to fold the fixes into this PR across several additional CMAP rounds:

- **Silent-loss on a dropped PTY write** — a shellper socket that dies *during* the paced text→…→Enter sequence (#1198) was invisible to the t=0 `writable` precheck, so the row was marked delivered off the paced-write timer. The dropped-write result is now threaded end-to-end as a boolean; a dropped write **holds** the row (`no-live-pty`) — never a false delivery.
- **Architect-identity resolution** — `afx send architect` was always `held(no-profile)` (architect sessions had no persisted `command`). Fixed with a restart-safe identity SSOT on the session row (migration **v16**) + a legacy self-heal that honors `TOWER_ARCHITECT_CMD` precedence.
- **Render-gate correctness** — the classifier had only ever been validated against a *synthesized* `claude-idle` fixture, so real-output false-`busy` bugs shipped latent. Resolved by capturing **real** gzipped terminal rings as fixtures and moving classification onto a **persistent bounded `SessionScreen` mirror** per session, fed incrementally at `PtySession`'s output chokepoint: each classify is O(viewport) rather than O(ring), the `[PIR #1205]` ring cap can no longer tear the classifier's input, and the whole-render `#1047` OOM residual is closed. A per-session `ringToken` verdict memo keeps repeated checks cheap.
- **Ghost-cursor exemption** — claude paints a dim suggested-command "ghost" into an idle composer whose first cell doubles as the inverse block cursor; the classifier now exempts exactly that cell (inverse + non-dim, at the headless cursor, with a dim non-empty tail). Requires positive ghost evidence, so a real one-char draft can never be false-cleaned.

## Testing

- **Automated e2e** for the #1265 repro: draft → held(`busy`) → submit → clean delivery (subprocess harness).
- Unit: mailbox lifecycle, gate classification vs captured real fixtures (claude/codex/agy: idle/draft/menu/picker/trust/wrapper/ghost), write serialization, dropped-write hold (sync first write + delayed Enter/multiline), inbox routes (incl. 405 method guard), cron supersede, escalation-is-visibility-only, restart-safe identity, and production-path render tests that push real >2 MiB captures through a real `RingBuffer` **and** a real `SessionScreen` (ring path torn/BUSY, mirror path CLEAN).
- Playwright: dashboard held-count indicator (4/4 real chromium, incl. live update).
- Last verified green: `packages/codev` **4551 passing / 48 skipped / 0 failing**; VSCode 667; dashboard 328; `send-integration.e2e` 7/7.
- **Live verification completed** (architect ran the built + installed code): idle→delivered; draft→held(`busy`)→delivers on clear; bg-task/monitor→delivered (no false-busy); real >1 MiB rings classify clean; idle-with-ghost→delivered; agy trust-dialog held. No held-message regressions.
- See the review's **Flaky Tests** section for an environmental temp-dir/`chdir` race under concurrent-builder load (not a code defect; green on direct run + retry) and the CI-aware render-gate perf bound.

## Notes for reviewers

- **Phase 7 ended on a porch force-advance** at the 3-iteration ceiling: the iter-3 Codex fix (a real `POST`-only method guard on `/api/inbox/:id/dismiss` + regression test) landed and Claude approved, but there was no iter-4 re-consult. Flagged honestly in the review; the pr gate is the backstop.
- **`CLAUDE.md` / `AGENTS.md` are untouched** — those two prompt surfaces are owned by the open Spec 1280, so 1313's send-outcomes docs live only in the `afx` reference (`agent-farm.md`) + skeleton mirror + `arch-critical.md`. Both files are byte-identical to `origin/main` and to each other; 1313 makes no net change to them.
- **Adopt/reconnect torn-seed (`#1361`, fast-follow)** — a long-lived alt-screen session whose replay exceeds the 1 MiB reconnect cap can seed its gate mirror born-torn on a Tower restart and hold mail until the next repaint/viewer. This is **pre-existing** (the reconnect cap fed the old delivery path the same capped bytes) and **fail-safe** (the row stays persisted, visible in `afx inbox`, and escalates — delayed delivery, never loss or misdelivery), and self-heals on any repaint. Tracked separately so this PR stays scoped.

## Spec / Plan / Review

- Spec: `codev/specs/1313-afx-send-mailbox-first-delivery.md`
- Plan: `codev/plans/1313-afx-send-mailbox-first-delivery.md`
- Review: `codev/reviews/1313-afx-send-mailbox-first-delivery.md`


mohidmakhdoomi:
@waleedkadous — heads up on a cross-project collision. This PR edits `CLAUDE.md` and `AGENTS.md`, which trips Spec 1280's T16 manifest-completeness guard (`spec-1280-phase-manifest.test.ts`): it diffs every prompt-bearing change on the branch against 1280's manifests, so any concurrent PR touching those files fails it.

Since 1280 is still in progress (phase_0), I've reverted this PR's change to that test back to its original rather than modify your guard unilaterally. How would you like to resolve — scope the check to 1280's own changes, or something else?

mohidmakhdoomi:
## Architect Integration Review — 3-way CMAP

Ran a 3-way integration CMAP (Gemini · Codex · Claude) on the rebased branch (`46232e19`).

**Consensus: 2× APPROVE (Gemini, Codex) + 1× COMMENT (Claude), all HIGH confidence. No blocking issues.**

### Independently verified
- **Silent write-loss fix (the original blocker) is real** — all three confirmed in code: `WritableSession.write()` returns `boolean` → threaded through `writeMessagePaced` → `deliverAgentMail` holds `no-live-pty` on a dropped write instead of reporting `delivered`.
- **Clean rebase onto current `main`** — 0 behind, merge-base == main tip; MERGEABLE.
- **Spec 1280 boundary respected** — `CLAUDE.md` / `AGENTS.md` byte-identical to `main` (both `916f75de`); the 1280 T16/manifest guard was correctly reverted and escalated to the owner.
- **Spec 1307 `--delay` preserved** — re-homed onto the mailbox (in-memory timer fires → mailbox enqueue at due-time; pre-due drops on restart per 1307), with a new regression test covering the load-bearing ordering guarantee (a delayed send never overtakes already-queued mail).
- **Full suite green via CI** — all 6 checks pass on `46232e19` (Unit, CLI ×2, CLI Integration, Tower Integration, Package Install).

### Findings triage
- **Claude — skeleton docs "missing `afx inbox`/mailbox vocab": NOT REPRODUCED.** Verified the skeleton twin contains the `afx inbox` section, the `delivered`/`held` vocabulary, and the "Mailbox retention and escalation" config. The line-count delta vs the `codev/` twin is pre-existing structural divergence (self-hosted extras), unrelated to this PR. Codex independently confirms the mirrors were updated. No action.
- **Claude — `writeEscapeToSession` doesn't thread the `write()` boolean:** minor consistency nit on the ESC/`--interrupt` path (not a mailbox-delivery path, so the no-silent-loss invariant doesn't strictly apply). Non-blocking; optional follow-up.

### Design strengths noted across reviews
Dependency-injected `DeliveryPorts` (testable without a live Tower); per-agent `KeyedSerializer` prevents blob-fusion by construction; `ringToken` TOCTOU mitigation; generation-guarded drainer lifecycle; additive migrations (`mailbox` table, `terminal_sessions.command`) and additive API/SSE; read-only/count-only UI surfaces.

**Recommendation:** ready for the `pr` gate — pending human approval. No code changes required to merge.

---
Architect integration review

mohidmakhdoomi:
## Architect live-test finding — render-gate false-`busy` on claude's suggested-command ghost (blocks delivery to idle agents)

**Context**: live-tested this branch's build (local-install, installed dist md5-matches this worktree's dist). Five `afx send`s to `main` delivered clean (56–64ms each), then messages started holding: row `a21b6c64…` (builder-task-shxz → main, 04:25:28Z) is stuck `held`/`busy`/escalated while main's composer is visibly empty.

**Root cause (verified byte-level against the live ring, not inferred)**: claude v2.1.220 paints a **suggested-command ghost** into the idle composer when the agent's own reply mentions a runnable command. The final composer frame in main's ring is:

```
❯ ESC[7m a ESC[27m ESC[2m fx cleanup -p task-VdfD ESC[22m    (cursor parked on the 'a')
```

The ghost's first char doubles as the software block cursor: **SGR-7 inverse at normal intensity**; the rest of the ghost is SGR-2 dim. `classifyScreen`'s dim-placeholder rule skips the ghost body, but nothing exempts the inverse cursor cell → exactly **1 counted user cell** → `busy`/`user-text`, permanently, on an idle terminal. There are zero keystroke echoes in the ring — the composer is genuinely empty (the ghost text came from the agent's reply, not typing).

**Verification**: ran the *installed* `classifyScreen` + `CLAUDE_PROFILE` on the *live* ring → `{"clean":false,"reason":"busy","detail":"user-text"}`; the one counted cell is row 58 col 2 `"a"`, `isInverse=true`, `dim=false`, with `buffer.active.cursorX/Y` parked exactly on it. Counterfactuals: skipping the cursor-position cell → CLEAN(0 cells); skipping inverse cells → CLEAN(0). Positive control: an idle no-ghost claude terminal classifies CLEAN with the same code.

**Exonerated (all verified working)**: whole-ring render is live (ring is 0.09MB/1 partial — not the cap-tear class), backstop drainer + verdict memo/ring-token semantics, escalation-once, `afx inbox` list/dismiss, no-force-path. Single root cause: the verdict itself.

**Fix guidance** — full analysis, tradeoffs, and timeline in the main checkout's handoff dir: `/home/user/code/codev_root/codev/codev/spir-1313-captures/findings-ghost-cursor-false-busy.md`

- Preferred: **skip the cell at the headless buffer's cursor position** in the user-cell scan — or the stricter ghost-signature variant (inverse, non-dim cell at the cursor followed by a dim-or-empty run). Do **not** blanket-skip all `isInverse()` cells: an inverse-rendered *selection* over a real draft would false-clean.
- Known residual of plain cursor-skip (document or close via the ghost-signature form): a 1-char draft with the cursor Home'd onto its only char would false-clean; multi-char drafts still hold via their other cells.
- Add the captured fixture `claude-ghost-suggestion-empty.replay.bin` (139×63, same dir) as a regression test: current gate → `BUSY user-text(1)`; post-fix → `CLEAN`. Keep the existing fixtures green, and add/keep a real-draft fixture asserting BUSY post-fix ("no new corruption vector").
- Optional hardening: a sustained `user-text` hold whose counted-cell set is exactly {the cursor cell} is chrome, not a human — today this drift class is invisible to the classifier-stuck liveness net.
- Check codex/agy for analogous ghost/suggestion renderings before assuming claude-only.

**End-to-end check available**: held row `a21b6c64…` is still in the mailbox — after your fix is local-installed, it should deliver on the first clean gate pass.

Repro tools in the same captures dir: `gate-diag.mjs` (capture + classify any live terminal + dump counted cells), `counterfactual.mjs`.

---
Architect live-test finding

mohidmakhdoomi:
## Architect Integration Review — ghost-cursor iteration (3-way CMAP + live E2E)

**Live E2E: PASS.** Fixed build local-installed at 06:10Z. Held row `a21b6c64…` (stuck `busy` 105 min against a genuinely-empty ghost composer) survived the Tower restart and **delivered to `main` 1.6s after the backstop drainer started**. A queued row to `builder-task-shxz` drained right behind it, and fresh sends gate-and-deliver in ~100ms. Restart-persistence and clean-gate delivery both proven on the live system.

**CMAP verdicts**
|---|---|---|
| Gemini | APPROVE | HIGH |
| Claude | APPROVE | HIGH |
| Codex | REQUEST_CHANGES | HIGH |

### Required change (Codex blocker — architect-verified in source, agreed)

`isGhostCursorCell` (`render-gate.ts`) grants the exemption on a **dim-or-EMPTY tail**: the tail loop `continue`s past whitespace/empty cells and returns `true` if it never sees a non-dim cell. So the documented residual — a 1-char draft with the cursor relocated onto its only char — classifies **false-CLEAN** and mail would inject into a non-empty composer. That violates the spec's "no new corruption vector" / fail-toward-hold invariants; pre-fix, no such path existed.

**Fix**: require **≥1 dim, non-whitespace, non-chrome tail cell** as positive ghost evidence before exempting; an empty tail stays BUSY. Every real suggested-command ghost is multi-char (the live capture's tail is 23 dim cells), so the live fix is unaffected; a hypothetical 1-char ghost false-holding is the safe, spec-compliant direction. Add the regression test: inverse non-dim cursor cell with empty tail → `busy`/`user-text` (the 1-char-draft case), alongside the existing dim-tail → clean case.

### Follow-ups (not blockers — to be tracked as issues, echoed by two reviewers)

1. `gemini`/`opencode` are spawnable harnesses with no gate profile → all mailbox sends hold `no-profile` indefinitely with no spawn-time warning; consider profiles-as-extensible-data (Claude).
2. #1047 residual: one whole-ring render of a pathological unbounded `partial` can exhaust the Tower heap — bounded/off-thread or persistent-headless design (Codex + Claude, already deferred in-code to #1047; keep it visible there).
3. Nit: `writeEscapeToSession` doesn't thread the `write()` boolean (non-mailbox path; consistency only).

**Gate status**: `pr` gate stays unapproved until the tightening + regression test land green. Everything else — architecture fit, migrations, additive API, silent-loss fix, 1280 boundary — verified clean by all three reviewers.

---
Architect integration review

mohidmakhdoomi:
## Architect Integration Review — round 2 (Codex re-consult verified)

**Ghost-cursor blocker: CLOSED.** Verified in source: `isGhostCursorCell` now requires positive ghost evidence (`sawDimTail` — ≥1 dim non-ws/non-chrome tail cell; empty/whitespace-only tail stays `busy`). Empty-tail regression test present; I independently ran the render-gate suite in the worktree: **40/40**.

**New blocker (Codex re-consult): CONFIRMED by architect repro — the whole-ring gate never receives the whole ring in production.**

The branch carries `[PIR #1205] Cap the ring buffer's incomplete-line partial` (`MAX_PARTIAL_CHARS = 2 MiB`, `trimPartial()` halves to ~1 MiB, ESC-aligned) — merged from main. The render-gate header still claims the partial is "kept whole/unbounded", and the large-capture tests feed `classifyScreen` directly, bypassing the trim. Pushed both real captures through a **default production `RingBuffer`** (`pushData` in 64 KiB chunks → `getAll().join('\n')`), worktree dist code:

| Capture | Direct | Via production RingBuffer |
|---|---|---|
| `claude-bgtask-empty` | 2,794,991 chars → **CLEAN** | 1,680,872 chars → **busy/no-region-end** |
| `claude-bigring-empty` | 2,991,283 chars → **CLEAN** | 1,877,164 chars → **busy/no-region-end** |

This is the D2 alt-screen tear resurrected one layer down: any session whose no-newline partial crosses 2 MiB gets a torn reconstruction and holds **permanently** (`no-region-end`/`no-composer-marker`) — the exact outage the over-ceiling removal was meant to eliminate, now for the busiest long-lived agents. Liveness telemetry makes it loud, but it is still a delivery outage. **Merge-blocking.**

### Required to close

1. **Reconcile the gate with the capped ring without reverting #1205's memory bound.** Directions, builder's choice with CMAP: (a) persistent bounded headless-screen per session, fed incrementally — kills the trim sensitivity, the per-classify render cost, AND the #1047 OOM residual in one move (Codex's suggestion; the principled fix); (b) another faithful bounded snapshot. **Not acceptable**: hold-on-trim (recreates the outage as steady state). Prior art worth reading: #1205's own rationale says the display path relies on "the client's post-connect repaint nudge" — the gate has no such nudge today.
2. **Production-path regression tests**: both real large captures pushed through an actual default `RingBuffer` before classification (the current direct-feed tests masked this).
3. **Fix `ringToken` content-identity**: `partialBytes` *decreases* on trim, so `(currentSeq, partialBytes)` is no longer monotone/unique — the verdict memo can alias a stale verdict. Use a genuinely monotone signal (e.g. a cumulative bytes-appended counter on `RingBuffer`).
4. **Update stale docs/comments**: render-gate header's "unbounded partial"/"whole cumulative stream at any size", arch.md §7 mailbox section.

Everything else stands from round 1 (2× APPROVE, live E2E pass, ghost fix verified). Gate stays unapproved until this closes.

---
Architect integration review

mohidmakhdoomi:
## Architect Integration Review — PR #1330 (Spec 1313, mailbox-first delivery)

**Risk tier: HIGH** (~13.3k additions / ~2.2k deletions, 100 files, core Tower/agent-farm/DB-migration/terminal surfaces) → full 3-way CMAP + independent architect verification against source.

### 3-way CMAP verdicts
|-------|---------|-----------|
| Gemini | **APPROVE** | HIGH — no blocking issues |
| Claude | **APPROVE** | HIGH — no blocking issues |
| Codex | **REQUEST_CHANGES** | HIGH — one blocker: the deferred #1361 adopt/reconnect liveness gap |

The split is entirely on **one point**: the adopt/reconnect "born-torn" gate-mirror gap (#1361). Everything else is unanimous praise for a correctness-by-construction refactor.

### The contested point (#1361), adjudicated against source
Codex: an adopted/reconnected shellper session seeds `SessionScreen` from `capRingSeed()`'s 1 MiB tail (`tower-terminals.ts:777/1036`), so a long-lived alt-screen frame >1 MiB can be *born torn* → classified `busy` → mail held indefinitely until a repaint/viewer nudge; "would not merge with #1361 outstanding."

Architect verification (`pty-session.ts:227–230`): `attachShellper` feeds the gate mirror the **identical capped bytes** the ring already receives — `pushData(replay)` (pre-existing) and `feedGateScreen(replay)` (round-2) use the *same* `replay` string. Therefore this gap is:
- **Pre-existing, not a regression** — the reconnect cap predates Spec 1313; the old path fed the gate off the same capped ring. Round-2 fixes the *live-ring* tear (common case) and worsens nothing (strict Pareto improvement).
- **Fail-safe** — the row stays durably persisted, visible in `afx inbox`, and escalates. It is *delayed delivery, not loss* — the antithesis of the silent-loss/corruption this spec exists to kill.
- **Self-healing** — any repaint / viewer attach / output byte clears it (a Tower restart is normally followed by dashboard/VSCode viewers reconnecting).
- **Narrow** — requires >1 MiB single newline-free alt-screen frame + Tower reconnect + sustained idle + no viewer + no output, simultaneously.
- **Tracked** — #1361, with two concrete fixes (uncapped ≤8 MiB mirror seed, or repaint-nudge on adopt).

**Architect adjudication: non-blocking for this PR.** Claude concurs ("fail-safe, pre-existing, track don't block"); Codex's concern is legitimate and #1361 should be a required fast-follow, but a pre-existing, fail-safe, self-healing, tracked edge should not block a large corruption-eliminating refactor.

### Independent architect verification (all confirmed against source)
- ✅ **Silent-loss fix** (the prior integration round's blocker): correct — `writeMessagePaced(): Promise<boolean>` records any dropped write across the paced text→…→Enter sequence; `deliverAgentMail` defaults `written=false`, holds `no-live-pty` on drop *before* `markDelivered`, and invalidates the verdict memo in a rejection-safe `finally`.
- ✅ **Migrations v15/v16**: additive + idempotent — `CREATE TABLE/INDEX IF NOT EXISTS`; v16 `ADD COLUMN command` is `PRAGMA table_info`-gated. No destructive changes.
- ✅ **Branch state**: fully contains `origin/main` (0 behind, MERGEABLE) — the review's "un-rebased / T16-red / 285-behind" caveat is **stale/moot**.
- ✅ **spec-1280 collision neutralized**: `CLAUDE.md`/`AGENTS.md` are byte-identical to `origin/main` (0 net change) and to each other; 1313 touches neither prompt surface.
- ✅ **Hot-tier displacement clean**: the Spec 987 governance-tier fact and the forge-concept fact displaced from `arch-critical.md` both survive in cold `arch.md` (§Governance Docs, §Forge Concept Commands). No doc regression.
- ✅ **API changes additive**; SSE workspace-scoped; escalation is visibility-only (never a force path).

### Decision
Held at the `pr` gate pending the human's merge call (2-1 split on a HIGH-confidence blocker + the committed→integrated gate is human-only). Architect recommendation: **APPROVE for merge with #1361 as a required fast-follow.**

---
Architect integration review

waleedkadous:

This is an impressive piece of systems work — the render-empty gate with empirically-captured per-harness profiles (including the agy color-keyed placeholder rule and the ghost-suggestion fixture), the transactional cron supersede, and the no-force-path principle are all the right architecture. Ran a 2-way CMAP (codex REQUEST_CHANGES / claude APPROVE, both HIGH) with the maintainer's concerns injected; where the lanes contradicted each other I verified against the branch myself. I'd like to see three changes before merge:

1. **`--delay`'s lower bound is lost on the hold path** (verified: `mailbox-store.ts`/`mailbox-wiring.ts` carry no `notBefore` concept, and the NOT_FOUND→registry hold at `tower-routes.ts:1653` excludes only `escape && interrupt`, not delayed sends). A `--delay` message to a target without a live/writable session is persisted delay-less and can drain immediately — breaking the `/arch-save` clear→re-init sequencing that `--delay` exists for. Suggested fix: persist `not_before` on the row and have the drainer respect it — which as a bonus makes `--delay` durable across Tower restarts, *stronger* than the #1335 in-memory timer it replaces.
2. **Close the shutdown window inside the submission lock.** The timer path's delivery-time generation re-check is good (claude's lane verified it), but codex identified the residual: the due callback checks liveness *before* entering `submitToSession` without a re-check inside the lock, and marks the row delivered ahead of the queued write — a shutdown during lock-wait can write after the drop decision, and a failed write stays marked delivered. Re-check the generation inside the lock, mark delivered only after the write settles, and cover this seam with a route-level test (the existing test calls the predicate directly).
3. **Give the residue-starvation case a reachable alarm.** Both lanes converge here, and it's the maintainer's top concern: one accidental visible character on an *autonomous* builder's line holds all mail (including cron) indefinitely — `busy` streaks are excluded from liveness telemetry by design, and escalation reaches only VSCode/dashboard, which headless flows don't watch. Minimum viable: surface `heldCount`/`mailboxEscalated` in `afx status`, and route a non-injecting escalation notice to the *spawning architect's* mailbox after a held-age threshold. (The remedy `afx interrupt` exists; the missing piece is the alarm that tells anyone to use it.)

**Take-or-file follow-ups:** route `mailbox-wiring.ts:211`'s `writeMessage` through `submitToSession` so mailbox delivery can't interleave an interrupt (converges the two serializers the docs currently over-claim about); have `afx cleanup` dismiss a removed agent's held rows (today they pin `heldCount`/escalation forever); fix the stale "typing-aware send buffer" reference at `agent-farm.md:528`; and reconsider displacing the `git add -A` hot-tier line instead of the Spec 987 tier-routing meta-rule.

The cron supersede analysis came back fully clean — bounded to one held row per (workspace, task), verified transactional. And thank you for the honest T16-era course-corrections on this branch; the review record on it is exemplary.

mohidmakhdoomi:
@waleedkadous — ready for re-review. All three required changes are in, plus the take-now follow-ups, and I've re-verified the diff against source.

1. **`--delay` durable via `not_before`** — new column + PRAGMA-gated migration v17; every delayed send persists at request time on all target states, so the delay is never dropped on a hold path. Drain/escalation/count surfaces all honor eligibility (pre-due never delivers early, blocks later mail, or escalates). Now durable across restart — the documented reversal of 1307's drop-on-restart.
2. **Delayed-interrupt shutdown window closed** — ^C-only timer, body delivers via the gate; `isStillLive` + writability re-checked *inside* the submission lock; nothing marked delivered on this path. Route-level shutdown-during-lock test added.
3. **Reachable starvation alarm** — `afx status` shows held/escalated + remedy hint; coalesced non-injecting notice to the spawning architect (fallback main → first; no notice-about-notice; cleared on drain).

Take-now: `afx cleanup` dismisses held rows; both `agent-farm.md` trees re-trued; `arch-critical` Spec 987 rule restored / `git add -A` displaced; both doc over-claims scoped. Serializer convergence filed as #1365.

Re-verification found one real defect — a non-hermetic test (ANSI-strip missing the ESC byte, green only with color off) — now fixed and confirmed green under `FORCE_COLOR=1` and `NO_COLOR=1`. Optional owner-notice tightening folded in; one coverage item deferred as fast-follow.

Green under `FORCE_COLOR=1`: `packages/codev` 4587 pass / 0 fail, e2e 7/7. CLAUDE/AGENTS byte-identical to `origin/main`. Commits: `a8d8477c`, `cb7d10dd`, `9f7eeb4f`, `7d699773`. Held at the `pr` gate — no merge.

*— Architect integration review*

Issue #1280, opened 2026-07-28
Prompt-surface judgment rewrite: acceptance = blog-principles conformance per file (size is reporting-only)


waleedkadous:
## Problem

The always-on prompt surface a builder consumes is ~21,900 served words (measured: `1252-word-baseline.md`), dominated by process recipes: porch phase tasks (11,430w across a 10-iteration project), `protocol.md` inlined into every spawn (3,703w), CLAUDE.md how-to prose, and consult-type preambles. Spec 1252's measurement proved deduplication alone yields only −7%: the surface is not duplicated, it is **over-instructed**.

[The new rules of context engineering for Claude-5-generation models](https://claude.com/blog/the-new-rules-of-context-engineering-for-claude-5-generation-models) reports >80% of Claude Code's system prompt was deleted with no measurable performance loss, by replacing rules with judgment, deleting worst-case padding, designing interfaces instead of examples, and progressive disclosure. That operation — content deletion on judgment-trust grounds — was explicitly a Non-goal of Spec 1252 and has never been attempted here.

## Goal (SUPERSEDED — see Amendment below)

Rewrite the always-on instruction surface for frontier-model consumers: **>50% reduction in served always-on words**, measured with 1252's committed measurement script (counts served/expanded words; phantom-savings-proof).

Attack order by size: phase-task prompts → `protocol.md` spawn inlining (keep the state machine, gates, artifact contracts; delete narrative process) → CLAUDE.md residual how-tos (→ progressive-disclosure skills) → consult-type prompts (rubric + verdict contract; delete process prose).

## Baked Decisions

- **All prompt consumers are frontier models** (Claude 5, GPT 5.6, Gemini 3.6 class). No weak-model tier, no fallback scaffolding variant, no tiering mechanism. One form.
- **Scar rules are exempt and verbatim** — the eight compressed canonicals developed in Spec 1252 Phase 5 (six repo rules + shellper verified-orphan + Tower-restart permission) ship with the rewrite; the registry/enforcement concept from 1252 is rebuilt fit-for-purpose around the post-shrink surface, not before it.
- **Validation is A/B, not observational**: same issues executed by builders on old vs new prompts, compared on outcomes (gate friction, review rounds, correctness). Spec 1252's M12 established that observational baselines (n=17) can only detect large regressions — insufficient at deletion scale. The A/B design is a first-class spec section.
- Spec must define a rollback story (prompt surfaces are files; reverting is cheap — say so concretely).

[...]

PR #1331, opened 2026-08-01
Fix #759: pr-search includes all PR states so merged PRs are found

### LINKED ISSUE #759: forge: github pr-search.sh defaults to open PRs only; post-merge consult fails to find merged PRs
## Symptom

After PR #757 merged, the post-merge \`consult --type pr\` step failed with \`No PR found for branch: builder/spir-755\`. The same lookup worked pre-merge.

## Root cause

\`packages/codev/skeleton/forge/scripts/github/pr-search.sh\` (or wherever \`pr-search\` lives in the forge layer) uses \`gh pr list --search "...":\` which defaults to filtering by \`state:open\`. Once a PR merges, this lookup returns nothing.

## Fix shape

Add \`--state all\` (or equivalent flag) to the underlying \`gh pr list\` call so the search includes merged + closed PRs.

## Suggested protocol

BUGFIX — single small flag addition in one script.

waleedkadous:
## Summary

The `pr-search` forge concept only matched **open** PRs, so once a PR merged, the post-merge `consult --type pr` lookup failed with `No PR found for branch: ...`. Add an all-states flag so the search includes merged/closed PRs.

Fixes #759

## Root Cause

`packages/codev/scripts/forge/github/pr-search.sh` ran `gh pr list --search "$CODEV_SEARCH_QUERY"` with no `--state`. `gh pr list` defaults to `--state open` (confirmed via `gh pr list --help`), so merged PRs were excluded and the lookup returned nothing after merge. The GitLab counterpart (`glab mr list --search`) has the identical latent defect (defaults to opened).

## Fix

- `github/pr-search.sh`: add `--state all` to the `gh pr list` call.
- `gitlab/pr-search.sh`: add `--all` to the `glab mr list` call (the established GitLab all-states flag, per `gitlab/pr-exists.sh`).

This mirrors the pr-exists precedent (bugfix #568), which added `--state all` for the same reason. No gitea `pr-search.sh` exists; these scripts are single-source under `packages/codev/scripts/forge/` (no skeleton mirror).

## Test Plan

- [x] Regression test added (`bugfix-759-pr-search-state-all.test.ts`) — asserts each script fetches all states while preserving the search query; fails without the fix, passes with it
- [x] Build passes
- [x] All tests pass (4123 passed, 0 failed)

### DIFF
--- a/.claude/skills/codev/SKILL.md
+++ b/.claude/skills/codev/SKILL.md
@@ -57,3 +57,34 @@ codev doctor
 - `codev init` creates a new directory — use `codev adopt` for existing projects
 - Always run `codev adopt` and `codev update` from the project root
 - `codev update` only updates framework files — it never touches specs/plans/reviews
+
+## Local build and install (this repository)
+
+Test changes locally before publishing. Run from the repository root:
+
+```bash
+pnpm build                  # builds core first, then codev (including dashboard)
+pnpm -w run local-install   # packs both packages, installs globally, restarts Tower
+```
+
+`local-install` (`scripts/local-install.sh`) packs `@cluesmith/codev-core` and
+`@cluesmith/codev`, installs both in a single `npm install -g` (separate installs fail —
+`codev-core` is not on the public registry), restores the executable bit on
+`scripts/forge/**/*.sh` that `pnpm pack` strips, and restarts Tower last. Install runs while
+Tower is up; only the final restart causes downtime. **Do not stop Tower first**, and do not use
+`npm link` / `pnpm link` — it breaks global installs.
+
+`pnpm build` also runs `copy-skeleton`, which copies `codev-skeleton/` into
+`packages/codev/skeleton`. **Tests read that copy**, so after editing anything under
+`codev-skeleton/` you must rebuild before the suite reflects your change.
+
+### Where to run things
+
+- `pnpm install` — repository root (installs all workspace packages)
+- `pnpm build` / `pnpm test` — `packages/codev/`, or `pnpm --filter @cluesmith/codev build`
+- Unit tests `packages/codev/tests/unit/` · E2E `packages/codev/tests/e2e/`
+- Never run npm commands from the repository root unless told to
+
+### Measuring code size
+
+`tokei -e "tests/lib" -e "node_modules" -e ".git" -e ".builders" -e "dist" .`
--- a/.claude/skills/runnable-worktrees/SKILL.md
+++ b/.claude/skills/runnable-worktrees/SKILL.md
@@ -1,0 +1,119 @@
+---
+name: runnable-worktrees
+description: Make builder worktrees runnable — the `.codev/config.json` `worktree` block (symlinks, postSpawn, devCommand), the `afx dev` CLI, VSCode dev controls, and per-stack config recipes. Use when configuring a repo so reviewers can run a builder's branch, when `afx dev` fails to bind or start, when a dev process is orphaned holding a port, or when asked why worktree dev uses the same ports as main.
+---
+
+# Runnable worktrees
+
+When configured, each builder worktree (`.builders/<id>/`) becomes runnable: reviewers can run
+whatever your dev command starts — a dev server, `cargo run`, `expo start`, a test watcher, a
+build script — against the builder's branch without `cd`'ing, installing, or hunting for the
+command. Opt-in via `.codev/config.json`; unconfigured repos see zero behavior change.
+
+## Config: the `worktree` block
+
+```jsonc
+{
+  "worktree": {
+    "symlinks":   ["..."],   // globs symlinked from the workspace root into each new worktree
+    "postSpawn":  ["..."],   // shell commands run inside each new worktree after createWorktree
+    "devCommand": "..."      // consumed by `afx dev <builder-id|main>`
+  }
+}
+```
+
+- **`symlinks`** — globs resolve from the workspace root and link into the worktree at the same
+  relative path. Root `.env` and `.codev/config.json` are *always* symlinked regardless.
+  **Symlinks, not copies**, so edits to main's env files reflect instantly in a running dev
+  session. A directory match is silently skipped (a glob cannot mask the worktree's own source)
+  **unless** the entry ends in a slash: `".local-user-data/"` is treated as a literal path and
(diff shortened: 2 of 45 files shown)


amrmelsayed:

Three-lane consultation (Gemini, GPT-5.6 Sol via Codex, Claude Opus) plus an independent architect pass. Because GitHub is serving a stale diff for this PR (see below), the lanes were fed the true effective diff and all consumer call sites directly; every finding below was verified against the live repo, several by direct `gh` reproduction.

**Verdict: REQUEST CHANGES — one blocker, everything else small.** All three lanes returned REQUEST_CHANGES, converging independently on the same regression, which was also reproduced live against this repository before the lanes ran. The fix itself is correct and needed; it just has to bring one caller with it.

### First, the diff you see is not the diff you get

The PR page shows 44 files, +2812/−5759. That is a stale artifact: this branch was cut from the Spec 1280 working branch, whose commits have since merged to main via PR #1362 and are now shared history. The **true effective diff, computed locally against the real merge-base, is 5 files, +101/−2**: the two script one-liners, one regression test, and the thread/status artifacts. Recommend rebasing onto main before merge so the record reflects reality; the merge itself is safe either way (branch is MERGEABLE, and a merge-tree run confirms it).

### Premise verified empirically

Issue #759 is real and still unfixed on main, reproduced against this repo with merged PR #1468 (branch `builder/pir-1465`):

```
gh pr list --search "head:builder/pir-1465"              → []        (today's script: post-merge lookup fails)
gh pr list --state all --search "head:builder/pir-1465"  → [#1468]   (fixed script: found)
```

The `--state all` / `--all` choices mirror the #568 precedent in `pr-exists.sh` exactly, and `glab mr list --all --search` is a valid combination.

### Blocking: the fix breaks `afx spawn`'s open-PR collision check

`spawn-worktree.ts:592` queries pr-search with `in:body #<issue>` and **no state qualifier**, then treats every hit as an open PR and fatals unless `--force`. Today the open-only default makes that correct. With `--state all`, merged and closed PRs flood in — and the `PrSearchItem` contract (`{number, headRefName, baseRefName}`) carries **no `state` field, so the caller cannot filter results**; the query itself must be state-explicit. Reproduced live:

```
gh pr list --search "in:body #1198"              → []                          (today: spawn proceeds)
gh pr list --state all --search "in:body #1198"  → [#1204, #1218, #1330, #1353, all MERGED]
```

After this PR, spawning a builder for any issue that ever had a merged or closed PR referencing it — which is every re-spawn, every follow-up to a `Part of #N` partial fix, every retry after a closed PR — aborts with a factually wrong "Found N open PR(s)" message. Two lanes independently reproduced the same result against different issues.

**Required fix, in this PR:** change the query to `in:body #${issueNumber} is:open`. An explicit `is:` qualifier overrides the state flag in both directions — verified: `head:builder/pir-1465 is:open` returns `[]` under `--state all`, and cleanup.ts's `is:merged` queries already work this way. Add a test asserting the spawn-side query carries `is:open` (the existing spawn-worktree test mocks the forge call, so it cannot catch this).

### Recommended: make the regression test able to fail

The new test asserts `content.toContain('--state all')` — but the explanatory comments this PR adds to both scripts also contain those exact strings, so **the assertions pass even if the flag is removed from the `exec` line**. Anchor them to the command line (e.g. match `^exec gh pr list --state all` / `^exec glab mr list --all`) so the pin is real. Two micro-nits while in there: `toContain('--all')` is a loose substring (it also matches `--state all`), and the file sits under `commands/porch/__tests__/` while testing forge scripts.

### Corrections to claims that came up in review (so they don't propagate)

- **cleanup.ts is NOT broken today and is NOT changed by this PR.** Two lanes claimed its `head:X is:merged` query returns nothing under the open-only default and that this PR repairs it. Tested directly: it returns the merged PR today — gh does not inject its default state qualifier when the search string carries its own. No repair, no regression; behavior there is identical before and after.
- **Result ordering is relevance-ranked, not newest-first.** `in:body #1280 --state all` returns 12 results led by #1409 and including the unrelated #272. `prs[0]` in `findPRForCurrentBranch`/`findPRForIssue` was safe when only one open PR could match; across all states it can pick a stale PR for reused branches or heavily referenced issues. Non-blocking — the bug being fixed is strictly worse — but worth a follow-up issue: sort explicitly, disambiguate, or add `state` to `PrSearchItem` so callers can.

### No conflict with PR #1146, in fact as well as in paths

Verified directly: `git merge-tree` on the two PR tips merges cleanly, the effective diffs share zero files, gitea has no `pr-search.sh`, and the gitea preset explicitly disables the concept (`forge.ts:129`). The two PRs are independent in both orders. The exit-0-on-error house style ruled on #1458 does not apply here: these scripts are `exec gh`/`exec glab` one-liners, and both CLIs report failure through their exit codes honestly.

### Follow-ups (file after merge, none blocking)

- Relevance-ordering risk on `prs[0]` (above) — sort/disambiguate, or add `state` to the pr-search output contract.
- glab's `--search` filters title/description only, so GitHub-syntax queries (`head:`, `is:`) are semantically inert on GitLab — a pre-existing limitation of the gitlab pr-search concept, unchanged by this PR, worth its own issue.

### Process notes

- The consult lanes were run against the true diff rather than GitHub's stale view; lane findings contradicted by direct reproduction were discarded with the evidence stated above.
- Merge authority rests with the maintainer — this review is the reviewing architect's recommendation, not a gate approval.

amrmelsayed:
One clarification on the blocker above, because the framing matters: **the latent defect is ours, not yours.** Your change is correct in isolation and fixes #759 exactly as specified. What it exposes is an unstated open-only assumption in `spawn-worktree.ts` that predates your PR — a caller that relied on pr-search's default state filter instead of saying what it meant, with an output contract that gives callers no `state` field to defend themselves with. Asking for the `is:open` line in this PR is only because your change is what makes the latent path reachable; you would be patching our defect as a courtesy, not fixing a mistake of yours.

The underlying contract gap (`PrSearchItem` carrying no `state`) is being filed as a separate follow-up issue on our side — it is not yours to solve here.

waleedkadous:
Superseded by #1619 — the same two commits (fix + regression test) cherry-picked onto current main. This branch was accidentally cut from the Spec 1280 builder branch, which is why the diff carried ~8.5k unrelated lines. Fix is now on main; #759 closed.

PR #1334, opened 2026-08-02
[Bugfix #1333] afx send: surface Tower's descriptive error message instead of bare NOT_FOUND

### LINKED ISSUE #1333: afx send: builder→non-spawning-architect send surfaces bare [error] NOT_FOUND, dropping the descriptive reason
## Problem

When a builder addresses an architect it isn't allowed to (`afx send architect:<name>` where `<name>` isn't the builder's spawning architect), the CLI prints only the opaque code:

```
Sending Instruction
───────────────────
[error] NOT_FOUND
```

The agent/human can't tell *why* it failed — "no such architect" vs. "not authorized to address that architect" look identical. During Spec 1313 manual testing a builder literally had to guess: *"no architect named codex-architect is registered/active (or, since I'm a builder, architect:<name> only works for my own spawning architect)."*

Tower **does** generate a descriptive message for exactly this case — it just never reaches the CLI.

## Root cause (verified in code)

Two layers collaborate to drop the detail:

1. `resolveArchitectByName` (`packages/codev/src/agent-farm/servers/tower-messages.ts`) returns a helpful message on the spoofing rejection:
   ```js
   return { code: 'NOT_FOUND', message: addressSpoofingErrorMessage(sender) };
   // "builder <id> may only address its own spawning architect"
   ```
   (and a distinct `"Architect '<name>' not found in workspace '<ws>'."` for a genuinely-missing architect.)

2. `handleSend` (`packages/codev/src/agent-farm/servers/tower-routes.ts`) serializes both fields:
   ```js
   sendJson(res, statusCode, { error: errorCode, message: result.message });
   ```

3. But the client's generic error extraction (`packages/core/src/tower-client.ts`, `request()`) prefers the **code** over the **detail**:
   ```js
   error = json.error || json.message || text;   // → "NOT_FOUND", message discarded
   ```
   `send.ts` then throws `new Error(result.error)` → `fatal("NOT_FOUND")`.

So the descriptive `message` is produced server-side and thrown away client-side.

## Impact

Diagnosability only — no misrouting, no corruption. But it makes a common, legitimate rejection (the builder spoofing guard) hard to understand, especially in multi-architect workspaces.

## Suggested fix (one to decide during planning)

- **Preferred:** when the response carries both, surface the human-readable `message` (optionally as `"<message> (<code>)"`) instead of the bare code — either in `tower-client.request()`'s extraction (`json.message || json.error`, or combine) or in `send.ts`'s error rendering. Keep the machine code available for programmatic callers.
- Confirm the two error strings remain distinguishable: spoofing rejection vs. genuinely-absent architect.

## Notes

- **Pre-existing**, not introduced by Spec 1313 — the generic `json.error || json.message` extraction predates it. Surfaced while testing #1313 (see `SPIR-1313-testing.md`, steps 7/9/14/15).
- Scope is the `afx send` / Tower error-surfacing path; addressing/spoofing **semantics** are correct and should not change (a builder may only address its own spawning architect).
- Small, isolated (~error-rendering seam); good BUGFIX candidate.

mohidmakhdoomi:
## Summary

`afx send architect:<name>` from a builder to an architect it isn't allowed to address printed only the opaque code:

```
[error] NOT_FOUND
```

The agent/human couldn't tell *why* it failed — "no such architect" vs. "not authorized to address that architect" looked identical. Tower **does** generate a descriptive message for exactly these cases; it was being discarded client-side. This PR surfaces that message (with the code kept as a suffix) so the two cases are distinguishable and the reason is visible.

Now:

```
[error] builder bugfix-1333 may only address its own spawning architect (NOT_FOUND)
```

Fixes #1333.

## Root Cause

The descriptive message is produced server-side and thrown away client-side:

1. `resolveArchitectByName` (`tower-messages.ts`) returns a helpful `message` for the spoofing rejection (`builder <id> may only address its own spawning architect`) and a distinct one for a genuinely-missing architect (`Architect '<name>' not found in workspace '<ws>'.`).
2. `handleSend` (`tower-routes.ts`) serializes **both** fields: `{ error: code, message }`. ✓ The message reaches the wire.
3. **`TowerClient.request()` (`packages/core/src/tower-client.ts`) extracted `json.error || json.message`** — preferring the machine code and discarding the human message. `sendMessage` then returned `{ error: 'NOT_FOUND' }` and `send.ts` rendered `fatal('NOT_FOUND')`.

The same extraction block existed **verbatim in two methods** (`request()` and `pasteImage()`), so both dropped the message identically.

## Fix

`packages/core/src/tower-client.ts` — extracted a single module-private helper `extractTowerError(text)` used by both sites (consolidating the duplicated block rather than editing it twice):

- When both a machine `error` code and a human `message` are present and **distinct** → `"<message> (<code>)"`.
- Otherwise → `message || code || rawText` (unchanged behavior for code-only responses like `STARTING_UP`, message-only responses, and non-JSON bodies).
- `typeof === 'string'` guards harden against non-string fields.

This keeps the machine code available for programmatic callers (it's still in the string) and is strictly more informative for **every** Tower CLI error, not just `send`.

**Scope:** addressing/spoofing *semantics* are unchanged — a builder may still only address its own spawning architect. This is purely the error-surfacing seam.

**Blast radius (verified):** no client-side code compares the extracted `.error` against code strings (the only `=== 'NOT_FOUND'/'AMBIGUOUS'/'NO_CONTEXT'` checks are server-side on `result.code`); server tests assert on `result.code` and the HTTP response body (unaffected); `send.test.ts` mocks `sendMessage` directly (unaffected).

## Test Plan

New regression test `packages/codev/src/agent-farm/__tests__/bugfix-1333-error-surfacing.test.ts` (6 cases), exercising `TowerClient` via the codev re-export → built core dist (the artifact the CLI consumes) against a stubbed `fetch`:

1. **The #1333 scenario** — `sendMessage` to a non-spawning architect surfaces `builder … may only address its own spawning architect (NOT_FOUND)`, not a bare code.
2. **Distinguishability** — the genuinely-missing-architect message differs from the spoofing message.
3–6. **Backward-compat guards** — code-only (`STARTING_UP`), message-only, equal code==message (no `"foo (foo)"`), and non-JSON body.

Proven **fails-without / passes-with**: reverting only the core source and rebuilding makes cases 1 & 2 fail with `Received: "NOT_FOUND"` (the exact bug); the 4 guards still pass. With the fix, all 6 pass.

- `npm run build` — passes
- `npm test` (full codev suite) — passes

### DIFF
--- a/codev/projects/bugfix-1333-afx-send-builder-non-spawning-/status.yaml
+++ b/codev/projects/bugfix-1333-afx-send-builder-non-spawning-/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-08-02T04:35:30.554Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-02T04:02:14.892Z'
-updated_at: '2026-08-02T04:16:37.303Z'
+updated_at: '2026-08-02T04:35:30.554Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-1333-afx-send-builder-non-spawning-/status.yaml
+++ b/codev/projects/bugfix-1333-afx-send-builder-non-spawning-/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1333
 title: afx-send-builder-non-spawning-
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-02T04:02:14.892Z'
-updated_at: '2026-08-02T04:06:33.826Z'
+updated_at: '2026-08-02T04:16:37.303Z'
--- a/codev/projects/bugfix-1333-afx-send-builder-non-spawning-/status.yaml
+++ b/codev/projects/bugfix-1333-afx-send-builder-non-spawning-/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1333
 title: afx-send-builder-non-spawning-
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-02T04:02:14.892Z'
-updated_at: '2026-08-02T04:02:14.893Z'
+updated_at: '2026-08-02T04:06:33.826Z'
--- a/codev/projects/bugfix-1333-afx-send-builder-non-spawning-/status.yaml
+++ b/codev/projects/bugfix-1333-afx-send-builder-non-spawning-/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1333
+title: afx-send-builder-non-spawning-
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-08-02T04:02:14.892Z'
+updated_at: '2026-08-02T04:02:14.893Z'
--- a/codev/state/bugfix-1333_thread.md
+++ b/codev/state/bugfix-1333_thread.md
@@ -1,0 +1,69 @@
+# bugfix-1333 — afx send drops descriptive NOT_FOUND reason
+
+Issue #1333 (area/tower). BUGFIX protocol, strict mode.
+
+## Investigate (complete)
+
+**Bug:** `afx send architect:<name>` from a builder to a non-spawning architect prints
+only `[error] NOT_FOUND`. The human/agent can't tell "no such architect" from "not
+authorized to address that architect" — both look identical.
+
+**Root cause (verified in code):** Tower *produces* a descriptive message; the client
+*drops* it in favor of the machine code.
+
+1. `tower-messages.ts:229-240` `resolveArchitectByName` returns a helpful `message`:
+   - spoofing: `builder <id> may only address its own spawning architect`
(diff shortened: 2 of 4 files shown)


mohidmakhdoomi:
## Architect Integration Review — approved, ready for **maintainer** merge

Human (architect) explicitly approved this change. Posting the review for the record — **not self-merging**; leaving the merge to a cluesmith/codev maintainer per repo policy.

**Root cause** — `TowerClient.request()` did `error = json.error || json.message`, preferring the machine code and discarding Tower's descriptive `message`, so the CLI surfaced a bare `NOT_FOUND`.

**Fix** — `extractTowerError()` returns `` `${detail} (${code})` `` when both are present (e.g. *"builder … may only address its own spawning architect (NOT_FOUND)"*), falling back to code, then raw body text; applied at both `request()` sites.

**Blast radius (independently verified)** — no consumer compares `result.error` against a code string programmatically, so broadening `code` → `"message (code)"` is safe; every caller just displays or throws it.

**Testing** — 93-line regression test, 6 cases (the #1333 spoofing scenario via `sendMessage`, spoofing-vs-genuinely-missing distinguishability, and backward-compat guards: code-only / message-only / code==code / non-JSON), verified fails-without/passes-with. Full build + full codev suite green (**4112 passed, 0 failed**). Prior CMAP: claude **APPROVE**; gemini/codex REQUEST_CHANGES were *solely* on an unrelated spec file, since removed by rebasing onto current `main`.

**Diff** — 4 files, `MERGEABLE`, base `main`: `packages/core/src/tower-client.ts` (27+/14-), the regression test, thread, porch status.yaml.

**Risk: Low.** Ready for a maintainer to merge — please use a regular merge commit (no squash), per repo convention.

— Architect integration review

waleedkadous:
Merged — thank you! Clean, well-scoped fix, and the before/after error text in the summary made this an easy review. The descriptive-message-with-code-suffix pattern is exactly right.

PR #1341, opened 2026-08-03
[Spec 1286] consult: configurable per-lane models and per-review-type lane selection

### LINKED ISSUE #1286: consult: configurable per-lane models and per-review-type lane selection
## Problem (owner-directed request from a production workspace, 2026-07-29)

Consult hardcodes the model per lane (`commands/consult/index.ts` — claude lane pinned to `claude-opus-4-6`, codex lane to `gpt-5.4` with medium reasoning). Workspaces cannot run newer models (e.g. `claude-opus-5` for spec/plan reviews, `gpt-5.6` for the codex lane) without patching dist or shadow-forking protocol files.

## Ask 1 — per-lane model config

Read per-lane model ids from `.codev/config.json`, e.g.:

```jsonc
{ "consult": { "models": { "claude": "claude-opus-5", "codex": "gpt-5.6" } } }
```

Fail fast on an unknown/invalid id (no silent fallback to the hardcoded default — per the fail-fast convention).

## Ask 2 — per-review-type lane selection

Config-level `porch.consultation.modelsByType` (e.g. `{spec:[...], plan:[...], impl:[...], pr:[...]}`) so lane composition can differ by review type. The requesting workspace implemented this today by tier-2-shadowing spir/aspir/pir `protocol.json` copies — which works but recreates exactly the stale-shadow-copy rot class that bit this repo (17 drifted files, fixed in PR #1281). A sanctioned config mechanism removes the incentive to fork protocol files.

## Evidence offer

The requesting workspace has lane-value data from production use (codex load-bearing at every review type; claude unique-finder at spec/plan; gemini ~zero unique findings and the only flaky lane) and is happy to share as input for defaults. Matches this repo's own experience (Codex caught the decisive defect in six consecutive review rounds on recent projects; the agy/gemini lane is #1032/#1033-degraded).

## Note

Config > protocol precedence is already porch's documented model-list rule; this extends the same principle to consult lane models. PIR's CMAP-2 cost invariant (config can silently inflate lighter protocols) should be considered in the design — per-protocol scoping may need to ride along.

waleedkadous:
Closes #1286.

Lets a workspace choose **which model** each consult lane runs, and **which lanes** run at all — per
protocol and per review type — from `.codev/config.json`.

This removes the incentive behind the issue: the requesting workspace had been tier-2-shadowing
`spir`/`aspir`/`pir` `protocol.json` copies to change lane composition, recreating exactly the
stale-shadow-copy rot class PR #1281 had just cleaned up (17 drifted files).

**Absent config, nothing changes** — ids included.

```json
{
  "consult": { "models": { "claude": "claude-opus-5", "codex": "gpt-5.6-sol" } },
  "porch": {
    "consultation": {
      "models": ["gemini", "codex", "claude"],
      "byProtocol": { "pir": { "models": ["gemini", "codex"] } }
    }
  }
}
```

## ⚠️ Behavior change (one, deliberate) — blast radius

**`porch done` no longer swallows config errors.** It previously wrapped config loading in a bare
`catch` that turned any error into a silent fall-back to protocol defaults.

**Who is affected**: only a workspace whose `porch.consultation` config is *already malformed*. That
workspace currently limps along on protocol defaults. After this change, `porch done` fails with:

```
Invalid consultation model "codexx" in porch.consultation.byProtocol.pir.models.
Valid models: "gemini", "codex", "claude", "hermes". Special modes: "none", "parent".
```

The error names the offending key, the value, and the valid alternatives, so the fix is the next
thing the user does. **A workspace with valid or absent config sees no change.**

Worth noting the previous state was worse than either alternative: `porch next` already failed this
way, so a typo made `next` refuse to run while `done` quietly demanded a *different* lane set — with
neither command printing the set it derived. Fail-fast, no fallbacks.

## 📋 Review history — what was and was not reviewer-approved

Stated in the PR, not only in the review file, so the gate reader can see it without opening
anything.

| Phase | Outcome |
|---|---|
| 1–5 | Unanimous codex + claude `APPROVE` |
| 6 (docs) | **Force-advanced at the iteration cap — NOT approved** |
| Spec | **Force-advanced at the iteration cap — NOT approved** |

**Phase 6** ended iteration 3 with codex `REQUEST_CHANGES` / claude `APPROVE`, hit
`max_iterations: 3`, and porch force-advanced (`force_advanced` record in `status.yaml`). The fixes
for codex's iter3 finding were committed but never re-reviewed. They are docs-only — JSON examples
made parseable, real pricing rates, PIR example consistency (`2cb1e2f7`).

**The spec** was likewise force-advanced at iteration 3; codex requested changes on all three passes,
each time on a genuinely different and valid defect, and its iter3 fix was committed post-review. An
architect-required 4th codex pass was run before planning (2 of its 4 findings were factually false
and are rebutted on file).

**Force-advance is not approval, and my own verification does not close out a standing
`REQUEST_CHANGES`.** So at the architect's direction a scoped confirming codex pass was run over
exactly the un-re-reviewed surface — phase 6's iter3 docs fixes (`2cb1e2f7`) plus the three PR-gate
CMAP fixes (`251c867f`, including the reverse exhaustiveness assertion).

> **Confirming verdict: codex `APPROVE` (HIGH)** — *"Both commits correctly resolve the outstanding
> findings without introducing defects."* KEY_ISSUES: None.

Codex independently confirmed the strict-JSON examples parse, the two doc trees are byte-identical,
documented values match the implementation, the SDK exhaustiveness guard is bidirectional, and the
replacement tests are non-circular. The verdict is reproduced verbatim at
`codev/projects/1286-consult-configurable-per-lane-/1286-confirming-codex-scoped.md` (run with an
explicit `--output` outside the porch project dir so it could not auto-persist and be miscounted as
a phase review; raw consult `.txt` outputs are gitignored repo-wide).

**Net: every change on this branch is now reviewer-approved**, either within its phase or by this
confirming pass.

My first draft of the review said "6 phases, all approved unanimously, 93 commits". That was wrong
on both counts and codex caught it at the gate; corrected in the review, the thread, and here.

## 🐛 Two real concurrency bugs found and fixed in `metrics.ts`

Both produced the **same invisible symptom** — a silently missing metrics row — because
`recordMetrics` swallows errors by design (a metrics failure must never take down a consultation).
Both trigger on the normal path: a CMAP opens one `MetricsDB` connection per lane, in parallel.

**1. The migration was check-then-act.** `PRAGMA table_info` then `ALTER TABLE`, unserialized. On
the first consultation after upgrading, all three lanes can see the column as absent; one adds it,
the others fail `duplicate column name` and lose their rows. Fixed with a fast path, `BEGIN
IMMEDIATE` + re-check inside the lock, and duplicate tolerance.

**2. `journal_mode = WAL` threw SQLITE_BUSY.** `busy_timeout` was set *after* the WAL pragma — and
does not protect a journal-mode switch anyway, since that needs an exclusive lock no busy-handler
waits for. So concurrent opens of a non-WAL database threw straight out of the `MetricsDB`
constructor. `stats.ts` and `analytics.ts` construct it unguarded, where the throw propagates.
Fixed in `enableWal()`: timeout first, skip the switch when already WAL, treat `SQLITE_BUSY` as
success-by-someone-else.

Neither was introduced by this spec; the second surfaced only once reviewers forced the concurrency
test to stop running against a stale `dist/`. Both fixes are mutation-verified (migration 5/6 runs,
WAL 5/5 deterministic).

## What's in it

| Phase | Delivered |
|---|---|
| 1 | Config schema, validators, resolvers (`lib/consult-lanes.ts`, `lib/skeleton.ts` helpers) |
| 2 | claude + codex lane model wiring; `--model-id` flag |
| 3 | agy `--model` passthrough; skip-vs-hard-failure split |
| 4 | `model_id` metrics column + guarded migration; codex costs `null` rather than wrong |
| 5 | One lane-selection resolver shared by `porch next` and `porch done` |
| 6 | Config reference, precedence ladder, fail-fast contract; skeleton parity |

**Precedence** (highest first; first level present wins, levels do not merge):
`byProtocol[P].modelsByType[T]` → `byProtocol[P].models` → `modelsByType[T]` → `models` →
the protocol's `verify.models`.

**Design notes**
- **No allowlist of model ids anywhere** — a hard spec constraint. Syntax is validated; existence is
  the provider's call, so a new model works the day it ships. Reasoning effort is the deliberate
  opposite: a closed enum bound to the Codex SDK type via `satisfies`, so SDK drift breaks the build.
- `consultation_metrics.model` still holds the **lane name** (`consult stats` groups on it); the
  resolved id went into a new `model_id` column.
- `consult.models` rejects `hermes` (no model selector → inert), while `porch.consultation` lane
  lists still accept it.
- Unknown `byProtocol` / `modelsByType` keys are **errors, not warnings** — a warning would leave
  you silently on the defaults you were trying to override.
- PIR's CMAP-2 cost invariant is documented with a worked example, since config outranking protocol
  is exactly how a SPIR-tuned 3-lane default silently inflates PIR.

## Testing

642 tests across the consult, porch, and lane suites; full unit suite, build, and `tsc` all green.
All 453 pre-existing porch tests pass **unmodified** — the plan named them as the regression net for
the resolver substitution.

Review with lessons learned: `codev/reviews/1286-consult-configurable-per-lane-.md`.

## Note for the reviewer

Three assertions in this branch were caught by CMAP as structurally unable to fail (a
`.rejects.toThrow()` against a `process.exit`, a tautology, a file-existence check standing in for
"the row landed"), and the docs phase needed four iterations, mostly for examples that didn't run.
Both are written up in the review. Every end-to-end assertion here is now mutation-verified.


waleedkadous:
**Post-review update** — CMAP-2 at the PR gate ran codex `REQUEST_CHANGES` / claude `APPROVE`; all three codex findings are fixed and pushed (`251c867f`).

**One code fix**, and it's a real gap rather than polish: the reasoning-effort list was bound to the Codex SDK union with `satisfies`, which catches values the SDK *removes or renames* but not ones it *adds*. An added value would compile clean while Codev hard-rejected it as invalid — failing open, in the direction the spec specifically required to break the build. The comment above it claimed all three cases were covered. Now paired with an `Exclude<...> extends never` assertion for the other direction, mutation-verified (dropping a covered member produces `TS2322`, since a type-level guard that never fires is indistinguishable from no guard). Its test was also circular — it iterated the list it was validating — so values are now pinned as literals and drift is carried at compile time.

**Two corrections to my own review**, which codex caught and which matter for judging this at the gate:

- **Phase 6 was force-advanced at the iteration cap, not unanimously approved.** It ended iteration 3 with codex `REQUEST_CHANGES` / claude `APPROVE`; my fixes for those findings were committed but never re-reviewed. Those unreviewed changes are docs-only (JSON examples made parseable, real pricing rates, PIR example consistency) and I verified them myself — all six JSON blocks through `json.loads`, skeleton `diff` empty — but no reviewer did. Phases 1–5 did end unanimous.
- Commit count is 96, not 93.

My review header had claimed "6 phases, all approved unanimously". I'd summarized my own project from memory instead of reading `status.yaml` — the same habit that caused every docs defect in phase 6. Corrected in the review and thread, and the review now states exactly which changes went unreviewed.

Claude's three non-blocking findings are recorded under **Known limitations and follow-up candidates** in the review rather than fixed here, since each would mean overriding a spec requirement or widening scope at the gate: `byProtocol` name validation being workspace-scoped while config can be global; `model_id` being write-only until `consult stats` surfaces it; and two cosmetic recomputations off the hot path.

No behavior change beyond what the PR body already describes.

waleedkadous:
**Confirming pass: codex `APPROVE` (HIGH), KEY_ISSUES: None.**

> *"Both commits correctly resolve the outstanding findings without introducing defects."*

Run at the architect's direction because force-advance is not approval and my own verification does not close out a reviewer's standing `REQUEST_CHANGES`. Scoped to exactly the surface that was never re-reviewed:

| Commit | What | Why it needed confirming |
|---|---|---|
| `2cb1e2f7` | phase_6 iter3 docs fixes | phase_6 force-advanced at `max_iterations: 3` with codex still at `REQUEST_CHANGES`; these landed after that verdict |
| `251c867f` | the three PR-gate CMAP fixes, incl. the reverse exhaustiveness assertion | made in response to codex's gate `REQUEST_CHANGES`, never re-reviewed |

Codex independently confirmed: strict-JSON examples parse, the two doc trees are byte-identical, documented values match the implementation, the SDK exhaustiveness guard is bidirectional, the replacement tests are non-circular, and the corrected review metadata matches `status.yaml`.

Run with an explicit `--output` outside the porch project directory so it could not auto-persist as `1286-review-iter1-codex.txt` and be miscounted as a phase review. Verdict preserved verbatim at `codev/projects/1286-consult-configurable-per-lane-/1286-confirming-codex-scoped.md` (raw consult `.txt` outputs are gitignored repo-wide, so no phase review is committed as a raw file either).

The PR body now carries the full review history up front — a table of what was and wasn't reviewer-approved, which specific changes went unreviewed, and this verdict closing them out.

**Net: every change on this branch is now reviewer-approved**, either within its phase or by this pass. Nothing further from me — the `pr` gate is Waleed's.

Issue #1350, opened 2026-08-05
codev-sdk: message-bus WebSocket client absent from v1 (tracking; mobile chat + gate-event consumers need it)


amrmelsayed:
## Finding (from the #1189 RN/Metro async review of PR #1346)

Scope note rather than defect: #1189's design listed a "WS/message-bus client" among the sdk's contents, and v1 ships without one. The sdk covers REST (`tower-client`) and SSE decoding (`sse`), but subscribing to `/ws/messages` (MessageFrame stream) still requires each consumer to hand-roll a WebSocket client, reconnect policy wiring, and frame parsing.

Consumers that will need it: the mobile chat + feed (first RN spike consumer), the web dashboard's eventual migration onto the sdk, and every consumer of #1194's typed `gate-event` frames (which ride the message bus as `BusEventMeta`).

## Ask

Confirm this was a deliberate v1 cut (likely: no consumer exists yet, matching the extract-on-evidence principle), and keep this issue as the tracking point for adding it when the first consumer lands. Shape suggestion when it happens, mirroring the existing seams: injected WebSocket constructor (RN provides a global `WebSocket`; Node consumers inject `ws`), `reconnect-policy` for backoff, typed `MessageFrame` from `codev-types`, and lifecycle hooks compatible with RN `AppState` foreground reconnects (verified necessary on-device in the LAN spike, `codev/research/mobile/spikes/expo-lan-reachability-2026-07.md`).

amrmelsayed:
Confirming from the merged plan (`codev/plans/1189-introduce-packages-codev-sdk-c.md`): **the cut was deliberate in effect but under-documented — SSE was scoped in as the v1 event channel; the message-bus WS client was never in the v1 build list at all.** The plan's only explicit event-channel decision (Design Decisions, ~line 306) scopes `subscribeEvents` (SSE) for the controller/streamdeck/mobile lanes and defers even in-repo SSE consolidation; a `/ws/messages` bus client has no consumer among the in-repo trio the PR migrated (dashboard/vscode consume terminals + SSE via their own paths), so under the plan's no-facades/single-source discipline it had nothing to migrate FROM and was implicitly out.

[...]

PR #1351, opened 2026-08-05
research: interaction-model §8.1/§8.3 amendment per #1206 rulings

amrmelsayed:
The follow-up main's #1206 rulings required: records the decided detection mechanism in §8.1 (harness-side hooks, builders-only v1, harness-neutral Tower contract, #1194-first transport split) and corrects §8.3's wrong claim that Tower "injects a tool_result" — hooks observe, they cannot answer; v0 is detect-and-notify with answering at the desk, and remote answering (synthesized PTY input) is its own gated spike per ruling Q2. The consumer contract (question_resolved dismissing cards cross-surface) is unchanged; only the injection mechanism was mischaracterized.

Related: #1206 rulings (comment 5013881893), decisions/askuserquestion-detection.md, #1194, #1350.

### DIFF
--- a/codev/research/mobile/interaction-model.md
+++ b/codev/research/mobile/interaction-model.md
@@ -256,7 +256,7 @@ Not mobile-specific — benefits web dashboard equally. **All of §8.1–8.3 is 
 
 ### 8.1 `AskUserQuestion` detection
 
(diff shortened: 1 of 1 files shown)


amrmelsayed:
Architect review: **approved**. Verified line-by-line against the #1206 rulings this records (comment 5013881893) — the §8.1 decision block, the builders-only v1 scope, the Q3 transport split, and the §8.3 retraction with the injection-fragility rationale are all faithful. The correction is exactly what ruling Q2 required. Docs-only in mobile's research tree — good to merge.

Issue #1347, opened 2026-08-05
Migrate the Stream Deck plugin into the monorepo as apps/streamdeck, consuming @cluesmith/codev-sdk


amrmelsayed:
## Goal

Execute the migration decided on #1189 (comments 5185933478 + 5185972772): the Stream Deck plugin moves from the `codev-integrations` repo into this monorepo as `apps/streamdeck`, consuming `@cluesmith/codev-sdk` — becoming the SDK's first consumer outside the original in-repo trio.

**Owner: the streamdeck architect** (this issue is teed up as its first spawnable work on launch). Prerequisite: PR #1346 (the SDK package) merged — satisfied.

## Scope

1. **Move the plugin source** from `codev-integrations/packages/streamdeck` → `apps/streamdeck` (fits #855's `apps/` = end-user-surfaces taxonomy). Preserve git history if practical (`git subtree`/filter-repo), else document the import commit as the history boundary with a pointer to the old repo.
2. **Migrate imports** from `@cluesmith/codev-client` → `@cluesmith/codev-sdk` subpaths. The mapping (per #1189's absorption plan):
   - `TowerClient` REST surface → `codev-sdk/tower-client` (auth arrives as an injected getToken; use the local-key profile)
   - `subscribeEvents`/SSE + `parseSseText` → `codev-sdk/sse`
   - `sendCommand` + `COMMAND_ROUTE` + the outside-in verb surface → `codev-sdk/controller`
   - local-key reading → `codev-sdk/node` (the plugin is a Node 20 process — the one consumer where this subpath is correct)
   - Gap check: anything the plugin needs that the SDK lacks is a FINDING — file it against #1189's follow-ups rather than hand-rolling locally.
3. **Build/CI wiring**: esbuild bundle + `streamdeck validate`/pack steps as a workspace member; CI job in `.github/workflows` (the `apps/vscode` marketplace-artifact pattern is the precedent). Ensure `pnpm install` cost for other builders stays reasonable (`@elgato/streamdeck` lands in the workspace).
4. **The published-SDK canary** (from #1189's plan): a scheduled CI job building `apps/streamdeck` against the **latest published** `@cluesmith/codev-sdk` instead of `workspace:*` — the external-fidelity check that replaces the dogfooding lost by moving in-repo. (May need to wait for the SDK's first npm publish — next release; land the workflow disabled/allowed-to-fail until then.)
5. **Deprecations and retirement** (trailing, gated on the SDK's first npm publish):
[...]

PR #1358, opened 2026-08-05
[#1357] codev-sdk: controller subpath re-exports overview wire types; codev-types becomes a regular (type-only) dependency

### LINKED ISSUE #1357: codev-sdk: controller subpath does not re-export the overview wire types its first consumer needs
## Finding (streamdeck architect's consumer-stakeholder review of the controller surface, per #1189 comment 5185972772)

The Stream Deck plugin (migrating in as apps/streamdeck under #1347) consumes four overview wire types as type-only imports: `OverviewData`, `OverviewBuilder`, `OverviewPR`, `OverviewBacklogItem` (in store.ts and actions.ts). The old `@cluesmith/codev-client` re-exported them from `@cluesmith/codev-types` so an integration got "the client + the contract from a single import".

No sdk subpath re-exports them today. `codev-sdk/controller` exports `TowerClient`, `COMMAND_ROUTE`, `TowerClientOptions`, `TowerWorkspace`, `parseSseText`, `SseEnvelope`, and `DEFAULT_TOWER_PORT`, but not the overview types, even though `TowerClient.getOverview` returns `OverviewData`.

#1347's import-boundary acceptance criterion (plugin imports only sdk subpaths + `@elgato/streamdeck`) is therefore unsatisfiable as the sdk stands: the migration would need a direct `codev-types` dependency, weakening the curated-surface posture the controller subpath exists for.

## Proposed fix

`controller.ts` adds type-only re-exports:

```ts
export type { OverviewData, OverviewBuilder, OverviewPR, OverviewBacklogItem } from '@cluesmith/codev-types';
```

This matches the subpath's stated purpose (a controller READS overview state) and the old client's single-import stance. Everything else the plugin needs already maps cleanly (method surface, `TowerWorkspace` as a compatible superset of the old `WorkspaceSummary`, verbatim-lifted `subscribeEvents`/`parseSseText`, the `getAuthKey`+`readLocalKey` local-key profile).

## Related wrinkle worth deciding at the same time

`codev-sdk` lists `@cluesmith/codev-types` as a devDependency, but its emitted `.d.ts` files reference it (`import type ... from '@cluesmith/codev-types'` in tower-client.ts). Workspace consumers resolve it via pnpm; a published-npm consumer running `tsc` will not, unless codev-types is a regular or peer dependency. This bites at the sdk's first npm publish and in #1347's published-SDK canary. Type-only, so "zero runtime deps" stays true either way; it is a packaging-metadata call.

## Ownership

Sdk changes are main's lane (#1189 ownership comment); filing rather than hand-rolling locally, per #1347's gap-check instruction. #1347's plan will reference this issue; the re-export is small enough to fold into the migration PR if main prefers that over a separate change.

Refs: #1189, #1347, PR #1346.

amrmelsayed:
Closes #1357. Unblocks #1347's import-boundary acceptance criterion (plugin imports only sdk subpaths + `@elgato/streamdeck`).

## Summary

Two small changes to `@cluesmith/codev-sdk`, per the issue and the architect's scope guidance:

1. **Overview wire types ship with the client.** `./controller` now re-exports `OverviewData`, `OverviewBuilder`, `OverviewPR`, and `OverviewBacklogItem` type-only from `@cluesmith/codev-types` — a controller READS overview state, and `TowerClient.getOverview` returns `OverviewData`, so the contract belongs on the subpath (matching the old `@cluesmith/codev-client` single-import stance). `./tower-client` additionally carries `OverviewData` for consumers importing the client directly.

2. **Packaging fix for npm consumers.** `@cluesmith/codev-types` moves from `devDependencies` to `dependencies`. The sdk's published `.d.ts` files reference it (`import type ... from '@cluesmith/codev-types'` in `tower-client.d.ts`), so a fresh npm consumer running `tsc` could not resolve it while it stayed a devDependency. codev-types is published (3.2.4 on npm) and types-only, so this would have bitten at the sdk's first npm publish and in #1347's published-SDK canary.

## Key decisions

- **`export type { ... } from` (whole-statement form), pinned by test.** The mixed `export { type X } from '...'` form is NOT equivalent — it still emits a runtime re-export statement (`export {} from '...'`), which would load codev-types at runtime. The import-boundary test now has a UNIVERSAL rule forbidding non-`export type` re-exports of codev-types, mirroring the existing `import type` rule, plus a pin test asserting the controller subpath re-exports all four overview types in the erased form.
- **Zero-runtime-deps intent restated, not weakened.** The boundary test's `dependencies` assertion previously required an empty field; its intent was always *zero RUNTIME dependencies*. It now asserts `dependencies` is exactly `['@cluesmith/codev-types']`, with a doc comment explaining the one allowed entry exists only so published `.d.ts` files typecheck — the source rules guarantee it is erased at build and never loaded at runtime. Verified against the emitted output: `dist/controller.js` and `dist/tower-client.js` contain no codev-types reference; only the `.d.ts` files do.
- **Lockfile updated** for the dependency-group move (`pnpm install`, workspace resolution unchanged).

## Test plan

- Extended `packages/sdk/src/__tests__/import-boundary.test.ts` (new re-export rule, restated dependencies assertion, new controller pin test).
- `pnpm --filter @cluesmith/codev-sdk build` — clean; emit inspection confirms type re-exports are erased from JS and present in `.d.ts`.
- `pnpm --filter @cluesmith/codev-sdk test` — 6 files, 73 tests passed.
- `porch check` — build + tests green.

### DIFF
--- a/codev/projects/1357-codev-sdk-controller-subpath-d/status.yaml
+++ b/codev/projects/1357-codev-sdk-controller-subpath-d/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-08-05T22:37:24.073Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-05T22:31:40.320Z'
-updated_at: '2026-08-05T22:36:18.590Z'
+updated_at: '2026-08-05T22:37:24.074Z'
+pr_ready_for_human: true
--- a/codev/projects/1357-codev-sdk-controller-subpath-d/status.yaml
+++ b/codev/projects/1357-codev-sdk-controller-subpath-d/status.yaml
@@ -6,11 +6,12 @@ plan_phases: []
 current_plan_phase: null
 gates:
   pr:
-    status: pending
+    status: approved
     requested_at: '2026-08-05T22:37:24.073Z'
+    approved_at: '2026-08-05T22:39:37.067Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-05T22:31:40.320Z'
-updated_at: '2026-08-05T22:37:24.074Z'
-pr_ready_for_human: true
+updated_at: '2026-08-05T22:39:37.068Z'
+pr_ready_for_human: false
--- a/codev/projects/1357-codev-sdk-controller-subpath-d/status.yaml
+++ b/codev/projects/1357-codev-sdk-controller-subpath-d/status.yaml
@@ -1,0 +1,14 @@
+id: '1357'
+title: codev-sdk-controller-subpath-d
+protocol: air
+phase: implement
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-08-05T22:31:40.320Z'
+updated_at: '2026-08-05T22:31:40.321Z'
--- a/codev/projects/1357-codev-sdk-controller-subpath-d/status.yaml
+++ b/codev/projects/1357-codev-sdk-controller-subpath-d/status.yaml
@@ -1,7 +1,7 @@
 id: '1357'
 title: codev-sdk-controller-subpath-d
 protocol: air
-phase: implement
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-05T22:31:40.320Z'
-updated_at: '2026-08-05T22:31:40.321Z'
+updated_at: '2026-08-05T22:36:18.590Z'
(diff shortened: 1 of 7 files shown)


amrmelsayed:
## Architect Review

**APPROVE** (pr gate; merge awaits Amr's authorization per current workspace policy).

Verified in the diff:

- **Type-only re-exports, both subpaths**: `controller` carries the four Overview* wire types; `tower-client` carries `OverviewData` (the return type traveling with the method that returns it). Both use the whole-statement `export type { … } from` form with in-code rationale.
- **The packaging fix is the important half**: `@cluesmith/codev-types` devDeps → deps, so published `.d.ts` files resolve for fresh npm consumers — the publish blocker #1357 flagged, dead before the first publish.
- **The boundary test evolved instead of being weakened** — this is the part I'd have pushed back on if fudged, and it wasn't: intent restated as zero-RUNTIME-deps, an exact-one-dep assertion (`['@cluesmith/codev-types']` — nothing else can creep in), and a regex banning any runtime-form re-export of the types package. The guard is stronger after the change than before it.
- Independently relevant: 73 sdk tests + e2e green per the builder; the change unblocks #1347's import-boundary acceptance criterion.

**On consultation**: none was run, and per the AIR protocol as written ('Optional — builder decides') that is the builder's documented discretion, correctly exercised for a change this size. The contradiction between that clause and CLAUDE.md's characterization (which I wrongly enforced against air-1352) is now tracked as #1359 — this PR proceeds under the protocol's actual text.

---
Architect review

Issue #1380, opened 2026-08-09
artifact-canvas: horizontal multi-column reading mode for spec/plan review (toggleable, vertical stays default)


amrmelsayed:
## Problem

The artifact canvas (the `afx open` annotation viewer and markdown preview — the surface where gate reviews of specs/plans/reviews happen) renders documents as a single vertical column. On wide monitors this wastes most of the screen: one column of readable measure occupies roughly a third of the width, and reviewing a long spec means constant vertical scrolling with only one screenful of context visible at a time. Gate review is a cross-referencing activity — requirements against acceptance criteria, a plan phase against its constraints — and a single-column viewport forces the reviewer to hold one side of every comparison in memory.

## Proposal

A **horizontal multi-column reading mode** for the canvas: content flows top-to-bottom within fixed-height columns of readable measure (~400px), continuing into the next column rightward — newspaper flow — inside a horizontally-scrolling container. Three to four columns per screenful on a typical wide display, so an entire spec section is visible at once.

- **Toggleable, vertical stays default.** A toolbar toggle + persisted preference; horizontal is an opt-in reading mode, not a replacement.
- **Implemented natively in the canvas** (CSS multi-column layout: `column-width` + `column-fill: auto`, plus a wheel-to-horizontal-scroll remap, axis-aware keyboard navigation, and a scroll-progress indicator). **No new dependencies.**
- Works in both canvas hosts (browser via `afx open`, VS Code webview).

## Why SPIR

Real design decisions belong in a spec phase:

1. **Fragmentation policy** — `break-inside: avoid` for code blocks, tables, marker cards, and the inline composer (a mid-composer column split is unacceptable); policy for blocks taller than the viewport (fragment vs overflow-with-inner-scroll).
2. **Affordance geometry** — the current '+' overlay is absolutely positioned in a left gutter keyed off `offsetTop`; both concepts dissolve in column layout. **#1343 (full-row affordance) is a prerequisite**: an in-row affordance travels with its block wherever the column places it. Marker cards are already in-flow (#863) and fragment naturally.
3. **Navigation semantics** — #1237's jump keys need axis-aware `scrollIntoView` (`inline: 'center'`); wheel remap must not fight native horizontal trackpad gestures; the minimap's vertical dot rail needs a horizontal-progress equivalent (or is suppressed in this mode).
4. **Preference surface** — where the toggle lives, per-workspace vs per-user persistence, and whether column width is configurable.

## Sequencing

[...]

Issue #1237, opened 2026-07-24
artifact-canvas: keyboard-first review navigation — jump keys, focus management audit, minimap focusability


amrmelsayed:
## Context

The canvas's core review loop is already fully keyboard-drivable (verified in source, 2026-07-25): every block is `tabindex="0"` (stamped at render, `renderer.ts:82`), Enter/Space on a focused block opens the inline composer, Cmd/Ctrl+Enter submits, Esc cancels, and the "+" affordance and marker-card edit/delete actions are real `<button>`s with aria-labels. A complete spec/plan review pass works without a mouse today.

## Gap: possible-by-keyboard ≠ pleasant-by-keyboard

1. **Linear-only traversal.** Tab-walking is the only navigation — a 300-block spec is 300 tab stops. No jump keys for: next/previous heading, next/previous block carrying a marker, first/last marker, top/bottom of document.
2. **Focus management unaudited.** Where focus lands after Esc-cancel, after submit, and after deleting a card is unverified — a keyboard flow that drops focus to the document root after every action forfeits its position and undoes the win.
3. **Minimap focusability unverified.** Dots have click handlers and `cursor: pointer`; whether they're reachable and activatable by keyboard (and whether activating one moves focus to the target block) needs checking.
4. **No shortcut discoverability.** Nothing surfaces the existing keys (Enter-to-comment, Cmd+Enter, Esc) — a small keys legend or aria-described hint would make the existing support findable.

## Proposed

- Jump navigation (suggested, bikeshed at plan time): `n`/`p` next/previous annotated block, `]`/`[` next/previous heading, with focus + scroll-into-view.
- Focus-management pass: Esc returns focus to the originating block; submit returns focus to the block (or its new card); delete moves focus to the neighboring card or block.
- Minimap dots: ensure real-button semantics; activation focuses the target block.
- A discoverable keys hint (e.g. `?` overlay or footer legend).

Enhancement; separate from the affordance hover/sizing bug (filed independently) because this is a feature while that is a defect.

amrmelsayed:
On it! Working on this with the PIR protocol (plan + dev-approval gates before PR).

Issue #1236, opened 2026-07-24
artifact-canvas: '+' affordance disappears while mouse travels toward it; also undersized vs prose font


amrmelsayed:
## Bug 1: the affordance vanishes en route (hover-geometry defect)

The "+" add-comment button intermittently disappears exactly as the user mouses toward it. Two mechanisms, both confirmed in source:

1. **Re-anchoring en route.** The overlay is anchored to the hovered block's FIRST line (`overlayTop = el.offsetTop + lineHeight/2`, `ArtifactCanvas.tsx:activateFromTarget`) at the far-left gutter. Traveling diagonally from mid-block (or a neighboring block) toward it crosses other `[data-line]` blocks, and every crossing re-fires `activateFromTarget` → the "+" instantly re-anchors to the newly crossed block, jumping away from the pointer. Path-geometry-dependent, hence intermittent.
2. **Overshoot dismissal.** The overlay hugs `left: 0` of the canvas; `onMouseLeave` on the outer canvas clears `activeLine` **instantly** (no grace). Overshooting one pixel past the canvas's left edge unmounts the button under the cursor.

(Crossing the gutter itself is safe — the body's `padding-left: 1.9rem` belongs to the body, and `activateFromTarget` early-returns on no-`[data-line]` without clearing.)

### Fix direction

- **Dismiss/re-anchor grace** (~200ms timer on both `onMouseLeave`-clear and block-crossing re-anchor), canceled when the pointer enters the overlay.
- **Pin while targeted**: `onMouseEnter` on `.codev-canvas-overlay` pins `activeLine` so nothing re-anchors out from under the cursor.
- Optional deeper alternative if the gutter remains troublesome: move the "+" into the block's own hover row (the GitHub-diff pattern), which eliminates the travel gap class entirely.

## Bug 2: the button is undersized relative to the prose

The button visually reads smaller than the markdown base font. Mechanism (deliberate-looking but wrong for this element): the token comment at `default-theme.css:93-96` notes the overlay "+", cards, and minimap chrome live OUTSIDE the font-sized containers — so the button never inherits `--codev-canvas-font-size` (16px) and renders at the host default (13px in VS Code webviews), further shrunk by `line-height: 1; padding: 0 6px`.

### Fix direction

- Apply `font-size: var(--codev-canvas-font-size)` to `.codev-canvas-overlay` (or the button), and give the button a minimum hit target of ~24×24px (WCAG 2.5.8 target-size floor) via padding/min-width/min-height.
[...]

PR #1385, opened 2026-08-10
artifact-canvas: full-row '+' affordance (GitHub-diff pattern)

### LINKED ISSUE #1343: artifact-canvas: full-row '+' affordance (GitHub-diff pattern) — structurally eliminate the hover travel-gap bug class
## Origin

At pir-1237's dev-approval gate (2026-08-05), reviewer testing of the shipped #1236 fix (grace timers + pin-on-overlay-hover) surfaced residual UX gaps the damping approach cannot close:

- Hovering the **right side** of a wide block shows the '+' far away at the block's first line, left edge — a long diagonal travel even with grace.
- **Dead strips** (page padding, block margins) show nothing at all — no affordance is reachable from them.
- The gutter-overlay anchoring model (from #863) means the affordance's position is only loosely coupled to where the pointer actually is.

## Proposal (reviewer's, endorsed at gate)

Adopt the GitHub-diff pattern: **the whole row is the hover target, and the '+' renders inside the hovered block's own row** (leading edge, vertically centered on the row under the pointer). This structurally eliminates the travel-gap bug class — there is no journey between trigger and target, so grace timers, pinning, and re-anchor suppression all become unnecessary and can be deleted.

## Scope notes

- Supersedes: the #1236 grace+pin machinery (small, deliberate interim — shipped to fix the live bug immediately, per the option-(a) decision at the gate), the absolute-overlay anchoring contract from #863, and their tests.
- Must preserve: the composer flow (#1107 inline-below portal), marker-card layout (#863's in-flow cards), keyboard activation parity (#1237's Enter/Space on focused block should light the same affordance), and the sizing/token fixes from #1236 (16px font token, 24px hit target) carried into the new placement.
- Design questions for the plan: hover target = block bounding box or full canvas row width (including padding strips)? What does the affordance do on blocks that already carry marker cards? Interaction with text selection (the '+' must not intercept drag-selection over the row).
- Test scenarios to inherit from the gate finding: right-edge hover, page-padding/margin strips, tall multi-line blocks, rapid vertical mouse travel.

PIR — supersedes approved layout decisions, design-sensitive. Related: #1236 (interim fix), #863 (current layout contract), #1107 (composer), #1237 (keyboard parity).

amrmelsayed:
# PIR Review: Full-Row '+' Affordance (GitHub-Diff Pattern)

Fixes #1343

## Summary

Replaced the artifact canvas's canvas-anchored "+" overlay (absolutely positioned from
`offsetTop` in a canvas-level gutter, damped by #1236's grace timers and pin-on-overlay-hover)
with the GitHub-diff pattern: every top-level block is a full-width row carrying its own
block-local leading space (`--codev-canvas-gutter`), and the "+" renders inside the hovered
row's own DOM — pointer-line tracked in tall blocks, first-line centered for keyboard focus.
Trigger and target now coincide, which structurally eliminates the hover travel-gap bug class:
the grace/pin machinery became dead code and was deleted, and the geometry is block-scoped as
required by #1380's queued multi-column mode. Comment granularity is unchanged (markers still
attach to the block's source line; the per-line feel is placement only).

## Files Changed

- `packages/artifact-canvas/src/components/ArtifactCanvas.tsx` (+118 / −96)
- `packages/artifact-canvas/src/styles/default-theme.css` (+89 / −27)
- `packages/artifact-canvas/src/components/__tests__/full-row-affordance.test.tsx` (+160 / −0, replaces `hover-affordance.test.tsx` −125)
- `packages/artifact-canvas/src/__tests__/default-theme.test.ts` (+31 / −7)
- `codev/plans/1343-artifact-canvas-full-row-affor.md` (+215, plan)
- `codev/reviews/1343-artifact-canvas-full-row-affor.md` (this file)
- `codev/resources/arch.md`, `codev/resources/lessons-learned.md` (governance updates below)
- `codev/state/pir-1343_thread.md` (builder thread)

## Commits

- `5e0706d7b` [PIR #1343] Plan draft
- `995dad5d4` [PIR #1343] feat: full-row in-row '+' affordance replaces canvas-anchored overlay
- `5c297f635` [PIR #1343] test: full-row affordance behavior + theme contract updates
- `65f845b4e` [PIR #1343] Thread log: implement phase notes

## Test Results

- `pnpm build` (tsdown, CJS+ESM): ✓ pass
- `pnpm run check-types` (tsc --noEmit): ✓ pass
- `pnpm vitest run`: ✓ 97 tests pass (9 new full-row tests; the 6 grace-window tests were
  deleted with the machinery they pinned — the new suite deliberately uses **zero fake timers**,
  so any future timer in this path fails the suite's instant-transition assertions)
- Manual verification: approved at the dev-approval gate (2026-08-10) — running-flow review of
  the inherited regression floor (right-edge hover, dead strips, tall blocks, rapid vertical
  travel), keyboard parity, composer/marker-card flows, and the full-bleed chrome-block look.

## Architecture Updates

Routed **COLD** (`codev/resources/arch.md`, VS Code Extension → Key Design Decisions): new
bullet documenting the row-scoped geometry contract — block-local gutter, "+" rendered inside
the hovered row, no canvas-anchored positioning — and why it is load-bearing for #1380 (block
padding travels through CSS column fragmentation; canvas geometry does not). Includes the
documented v1 limitation (wide-table horizontal scroll carries the "+"). Nothing routed HOT:
the contract is package-scoped, not a cross-cutting always-on fact.

## Lessons Learned Updates

Routed **COLD** (`codev/resources/lessons-learned.md`, UI/UX):

- Closed the loop on the `[From #1237]` travel-gap lesson (marked landed as #1343).
- New `[From #1343]`: accumulating damping (grace timers, pins) around a UI interaction signals
  wrong geometry — collocate trigger and target and the damping becomes deletable, not tunable;
  and let a *queued* feature's constraints (#1380 fragmentation) pick the geometry now.
- New `[From #1343]`: the movable-portal-target technique — one wrapper node held in a ref,
  `appendChild`-moved between hosts, portalled into; React ownership survives moves and
  `innerHTML` wipes of imperatively-managed DOM.

Nothing routed HOT (both are situation-specific recipes, not behavior-changing global rules).

- **`ArtifactCanvas.tsx` — `activateFromPointer`'s three guard clauses** (affordance-origin
  events, primary-button drag, no-block targets). The first is subtle: without it, hovering the
  "+" hosted in a `ul`'s gutter re-resolves to the `ul` and retargets a nested `li`'s line to
  the list's first line. Pinned by the "never retarget" test.
- **The re-host path** (decoration effect): after a watch reload rebuilds the body, the wrapper
  node is detached; the effect re-appends it for a still-valid `activeLine` via
  `activeLineRef` (not a dep — depending on `activeLine` would churn card injection per hover).
- **CSS specificity around the gutter**: the generic row rule (`.body > [data-line]`, 0-2-0)
  intentionally beats `.codev-canvas-has-marker`'s 10px padding (0-1-0) on top-level rows, and
  is itself beaten by the chrome-row sums (`pre[data-line]` etc., 0-2-1). The marker bar for
  top-level rows is a `::before` strip; nested marked blocks keep the classic inset shadow.
- **Full-bleed chrome rows** (flagged at plan gate, accepted at dev gate): pre/blockquote/table
  boxes start at the canvas edge; text x-positions preserved by padding arithmetic.
- **Documented v1 limitation**: a horizontally-scrolled wide table carries its "+" with the
  scroll. `pre` is immune (scroll moved to inner `code`).

## How to Test Locally

- **View diff**: VSCode sidebar → right-click builder pir-1343 → **Review Diff**
- **Run dev**: VSCode sidebar → **Run Dev**, or `afx dev pir-1343`
- **What to verify** (browser `afx open` + VS Code webview, light + dark):
  - Right-edge hover on a wide block: "+" appears in-row instantly; travel to it never
    dismisses or moves it
  - Leading (gutter) strips light their row; margins keep the previous row lit (sticky)
  - Tall code fence: "+" tracks the pointer's line; wide code scrolls inside the block without
    carrying the "+"
  - Rapid vertical travel: row-by-row tracking, no flicker or stale anchors
  - Drag-selection across a lit "+": selection uninterrupted, no '+' in the copied text
  - Composer (#1107), marker cards + gold bar alignment (#863), edit/delete (#1055), full
    #1237 keyboard pass (Tab/jump keys light the "+", Enter/Space opens the composer)


amrmelsayed:
## Architect Review

**APPROVE** (pr gate; merge awaits Amr's authorization).

**Post-consult fixes independently verified per the escalation** — all three codex findings were real and their fixes are correct:

1. **Affordance-origin keyboard guard** (`fromAffordance`) extended to the focus and body-keydown paths — without it, Tab/Enter on the '+' re-resolved through the host row and retargeted a nested block's line (an `li`) to its container's (`ul`), silently changing which line a comment anchors to. The guard + its rationale comment are in place; regression-tested.
2. **Nested gutter scoping** — injected siblings (marker-card stacks, composer host) now carry the block-local gutter only as direct children of the canvas body (child-combinator selectors, pinned by tests), so a stack nested inside an already-guttered row no longer double-indents.
3. **lessons-learned.md repair** — verified the diff: the pre-existing #1237 entry is restored byte-intact (with a 'landed as #1343' pointer), and the two new #1343 lessons are excellent additions — the 'accumulating damping means the geometry is wrong; collocate trigger and target' principle and the movable-portal-host pattern, both correctly routed COLD.

Claude's advisories: hr/raw-HTML alignment fixed (unmapped blocks lost their leading indent under the row model — real), one style nit declined with stated reasoning, and the narrow-table hover wrinkle documented rather than hidden. Ran the full canvas suite in the worktree: 98/98.

**The structural claim of this PR held**: grace+pin machinery deleted (not disabled), the affordance is block-scoped per the #1380-compatibility constraint, and the full-bleed chrome decision approved at plan gate landed with text x-positions preserved. The travel-gap bug class is now unrepresentable rather than damped.

---
Architect review

PR #1405, opened 2026-08-11
[Air #1403] Filter Stream Deck Zoom Navigator dial to active workspaces

### LINKED ISSUE #1403: Stream Deck: Zoom Navigator cycles dormant workspaces; filter the dial to active ones
## Problem

Rotating the Zoom Navigator at the workspace altitude walks every workspace ever registered, dormant ones included. Confirmed by the owner on hardware and traced in code.

Mechanism: the plugin store fetches its list via the sdk's `listWorkspaces()` (`apps/streamdeck/src/store.ts:79`), which hits Tower's `GET /api/workspaces`. That handler maps over `getInstances()` (`packages/codev/src/agent-farm/servers/tower-routes.ts:509`), i.e. every registration in the user-global `~/.agent-farm/global.db`, and includes an `active: i.running` flag per entry. The plugin stores the array unfiltered and the zoom cursor counts `workspaces.length`, so the `TowerWorkspace.active` field is delivered on every response and never consulted.

Dormant entries are dead stops on the dial: zooming into one yields no meaningful overview, and any action from there fails because commands require the workspace active in Tower. There is no activate-on-zoom behavior that would justify showing them.

## Requirement

Filter the store's workspace list to `active === true` (strict filter; owner-ratified 2026-08-11). The alternative of showing dormant entries demoted after active ones was considered and rejected: they have no actionable content, and the dashboard is the right surface for browsing registrations.

Care points for the implementation:

- Cursor indices: the zoom cursor and `selectedWorkspacePath()` index into the stored array, so filtering must happen at (or before) storage, not per-render, to keep indices coherent.
- A workspace deactivating while selected: after a refresh removes it, the cursor must land on a valid index (reset to 0 is fine), not point past the end.
- Focus-follow (`workspaceByPath` lookup when the user focuses a VS Code window): a path that is registered but dormant should now behave like an unknown path (no-op), which is the existing fallback.
- Regression test: list with a mix of active and dormant entries renders and cycles only the active ones.

## Suggested protocol

AIR: small, fully described, no architectural decisions.

amrmelsayed:
## Summary

Closes #1403. Rotating the Stream Deck Zoom Navigator at the workspace altitude walked **every** registered workspace, dormant ones included. Dormant entries are dead stops on the dial: zooming into one yields no overview, and any action from there fails because Tower commands require the workspace active. This filters the plugin store's workspace list to `active === true` (strict filter, owner-ratified 2026-08-11).

## What changed

All in `apps/streamdeck/src/store.ts` — the fix lives entirely in how the store populates its list; `cursor.ts`, the action layer, and the SDK are untouched.

- **`fetchActiveWorkspaces()`** (new private): wraps `client.listWorkspaces()` and keeps only `active` entries. Both fetch sites route through it — `refresh()` and the stale-list re-fetch inside `syncToWorkspace()`. Filtering happens **at fetch time, not per-render**, because the zoom cursor and `selectedWorkspacePath()` index into the stored array; filtering per-render would desync the indices.
- **Cursor clamp in `refresh()`**: when a workspace deactivates while it is the current selection, the next refresh drops it and the cursor could point past the end of the shorter list. `refresh()` snaps `cursor.workspace` back to 0 so `selectedWorkspacePath()` stays valid.
- **`syncToWorkspace()` (focus-follow)**: a dormant registration is filtered out, so its path stays unfound and falls through to the existing no-op — a registered-but-dormant path now behaves exactly like an unknown path (e.g. a builder-worktree window). Docstring + inline comment updated.

## Tests

New `describe('CodevStore active-workspace filtering')` in `apps/streamdeck/src/__tests__/actions.test.ts`:

- `refresh()` stores only active workspaces, and the dial cycles the active ones alone (clamped, dormant skipped) — the requested regression test with a mix of active + dormant entries.
- Cursor clamps to a valid index when the selected workspace deactivates mid-session.
- `syncToWorkspace` treats a dormant registration like an unknown path (no-op, no overview fetch).

Existing tests that set `store.workspaces` directly with dormant fixtures are unaffected — the filter runs only at fetch time, not on hand-assigned arrays.

## Verification

- `pnpm check-types` — clean
- `pnpm build` (esbuild) — OK
- `pnpm test` — 66/66 pass

(check-types needs `@cluesmith/codev-types` and `@cluesmith/codev-sdk` built first in a fresh worktree so their `dist/` type declarations resolve.)

## Notes / lessons

- The `TowerWorkspace.active` field was already delivered on every `/api/workspaces` response and simply never consulted — this wires up existing data, no API change.
- The dial clamps rather than wraps (per `cursor.ts`), so the regression test asserts clamp-at-end + reverse, not wrap-around.

## Protocol

AIR — small, fully described in the issue, no architectural decisions. ~83 LOC of source change. No spec/plan/review files (AIR produces none; this PR body is the review).

### DIFF
--- a/apps/streamdeck/src/__tests__/actions.test.ts
+++ b/apps/streamdeck/src/__tests__/actions.test.ts
@@ -232,6 +232,62 @@ describe('CodevStore.syncToWorkspace (focus following)', () => {
   });
 });
 
+describe('CodevStore active-workspace filtering', () => {
+  const mixed = () => [
+    workspace('/work/a', 'a', true),
+    workspace('/work/dormant', 'dormant', false),
+    workspace('/work/b', 'b', true),
+  ];
+
+  it('refresh() stores only active workspaces, so the zoom dial cycles active ones alone', async () => {
+    const ctx = makeStore();
+    ctx.listWorkspaces.mockResolvedValue(mixed() as never);
+    await ctx.store.refresh();
+    expect(ctx.store.workspaces.map((w) => w.path)).toEqual(['/work/a', '/work/b']);
+
+    // The cursor cycles only the two active entries (clamped, never the dormant one).
+    expect(ctx.store.selectedWorkspacePath()).toBe('/work/a');
+    ctx.store.rotateCursor(1);
+    expect(ctx.store.selectedWorkspacePath()).toBe('/work/b');
+    ctx.store.rotateCursor(1); // clamped at the end of the 2-entry active list
+    expect(ctx.store.selectedWorkspacePath()).toBe('/work/b');
+    ctx.store.rotateCursor(-1);
+    expect(ctx.store.selectedWorkspacePath()).toBe('/work/a'); // back to the first active, dormant skipped
+  });
+
+  it('clamps the cursor to a valid index when the selected workspace deactivates', async () => {
+    const ctx = makeStore();
+    // Start on the second of two active workspaces.
+    ctx.listWorkspaces.mockResolvedValueOnce([
+      workspace('/work/a', 'a', true),
+      workspace('/work/b', 'b', true),
+    ] as never);
+    await ctx.store.refresh();
+    ctx.store.rotateCursor(1);
+    expect(ctx.store.cursor.workspace).toBe(1);
+
+    // /work/b goes dormant; the next refresh drops it and the cursor snaps back.
+    ctx.listWorkspaces.mockResolvedValueOnce([
+      workspace('/work/a', 'a', true),
+    ] as never);
+    await ctx.store.refresh();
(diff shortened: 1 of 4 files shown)


amrmelsayed:
Domain-owner review (streamdeck architect): approach, focus-follow fall-through, and tests match #1403. One fix requested before the pr gate: the deactivation clamp (`{ ...this.cursor, workspace: 0 }`) carries the old `builder` index into the fallback workspace; if that workspace has enough builders, `selectedBuilder()` silently points at an unrelated builder and slot/dev keys act on the wrong one. `moveCursor` already resets `builder: 0` on workspace change — the clamp should too, plus a test for it. Sent to the builder via afx.

Issue #1428, opened 2026-08-12
Stream Deck: Builder Action key face is illegible (text over icon, inconsistent casing, truncation)


amrmelsayed:
## Problem (owner hardware feedback, 2026-08-12, photo-verified)

Three symptoms with one root cause. The Builder Action key pushes raw wire strings through `setTitle`, which the SDK paints over a full-bleed bolt icon (`apps/streamdeck/src/actions.ts:162-163`):

1. **Text/icon collision**: the two title lines land on the icon's diagonal. Legibility is bad enough that `#1414` reads as `#1,414` on hardware - there is no number formatter anywhere in the plugin and `issueId` is a plain string on the wire, so the apparent comma is the bolt edge showing through between digits. The overlay corrupts what the data appears to say.
2. **Casing inconsistency**: line 2 is `blocked ?? protocolPhase` verbatim - wire identifiers, not display labels - so the key mirrors whatever casing the wire carries ("verified" vs "implement").
3. **Truncation**: long phase names clip mid-word ("implement" renders clipped).

The Gates key reads fine because its icon and label were composed as one design; the Builder Action key stacks two independent layers.

## Requirements

1. **Composite key face**: render the key image in the plugin (SDK `setImage` with generated SVG/canvas): icon confined to an upper zone, reserved text band below. Match the legibility of the Gates key.
2. **Presentation label map**: a small phase/gate id -> display label map with deliberate, consistent casing and lengths that fit the key's line width (e.g. Implement, Verified, Dev Approval). No mid-word truncation by construction.
3. **Presentation stays plugin-local**: no wire, types, or server change. The wire keeps raw ids; the plugin decides display.
4. Apply the same face treatment to the empty-slot state ("Slot N") where trivial, for a consistent row.

## Out of scope

Press/rotate behavior (untouched: #1404 press-selects, #1414 Automatic remap), profile changes, any new key.

## Verification

Hardware session at the dev gate: photo-level check that issue number, phase label, and icon are each legible at a glance across states (active phases, blocked gates, empty slot). Protocol: PIR (the result must be seen on the physical deck before a PR exists).

## Context

Feeds the parked #1381 buttons-posture decision: the smart key became load-bearing with #1404's press-selects, so its face needs to carry its weight. Sequence after #1414 merges (same file, `actions.ts`).

amrmelsayed:
On it! Working on this with the PIR protocol (plan + dev-approval gates before PR).

amrmelsayed:
[...]

Issue #1414, opened 2026-08-12
Stream Deck SD+: Automatic diff press should open the builder's first file diff (dial-ready), not the aggregate


amrmelsayed:
## Problem

#1404's merged Builder Action Automatic press fires `view-diff` for implement/review builders, which opens the multi-file aggregate diff editor. The SD+ Files/Changes dials navigate per-file, so for dial-driven review the press should land on the builder's **first file in per-file diff mode**, positioned so the dials step from there.

Origin: #1404 dev-approval review discussion (owner + builder); routed through the architect.

## Ground truth (verified against the code 2026-08-12)

- `diff-first-file` → `codev.diffFirstFile` → `navigateDiffToFirst` opens the first per-file diff and seeds dial navigation, but takes **no builder-id argument**: `resolveDiffContext` (`apps/vscode/src/commands/diff-nav.ts:118`) resolves the builder from the active editor's diff-inject entry or `lastPosition`, and fired cold it flashes 'open a builder file diff first' and no-ops.
- `view-diff` → `codev.viewDiff` is the only relay verb that opens a specific builder's diff by id, and it opens the aggregate editor.
- So no verb today means 'open builder X's first file diff'.

## Verify first (fold into a hardware session)

Confirm on hardware whether the dials already navigate after an Automatic `view-diff` press (whether the aggregate open populates the diff-inject registry / `lastPosition`). If they do, this issue is a 'land on file 1 in per-file mode' polish; if they don't, it is a broken-dial-flow fix and rises in priority.

## Proposed (vscode-side)

Add a builder-id-scoped 'open first file diff' path: seed `navigateDiffToFirst` from an explicit id (reuse `resolveDiffContext` internals / `openBuilderFileDiff`), expose as a relay verb (e.g. `open-diff-first` taking `[builderId]`), and point the deck's Automatic diff branch at it. The explicit 'View Diff' PI option keeps opening the aggregate, unchanged.

## Relation / sequencing

Builds on #1404 (changes where its Automatic diff branch points) — after it merges. Fits #1410's dial-driven review workflow.

## Suggested protocol

PIR (hardware verification of the dial seeding either way).

amrmelsayed:
On it! Working on this with the PIR protocol (plan + dev-approval gates before PR).

amrmelsayed:
[...]

PR #1442, opened 2026-08-13
Fix #1431: map verify-approval gate to view-diff in phaseArtifactVerb

### LINKED ISSUE #1431: Stream Deck: phaseArtifactVerb doesn't map the verify-approval gate (Automatic press falls back to terminal)
## Problem (surfaced by the pir-1428 builder while mirroring gateIconFor, 2026-08-12)

`phaseArtifactVerb` (`apps/streamdeck/src/actions.ts:229-239`) maps gates `spec-approval`, `plan-approval`, `dev-approval`, `pr` - but not `verify-approval`, which the VS Code sidebar's gate vocabulary knows (`gateIconFor` in `apps/vscode/src/views/builder-row.ts` maps it to the `verified` codicon). A builder blocked at `verify-approval` returns `undefined`, so the Builder Action Automatic press opens a terminal instead of the review artifact, and the phase-aware dial mode resolves through the same resolver.

## Fix

Add `verify-approval` to the gate branch, returning `view-diff` (same as `dev-approval`/`pr` - the human is reviewing completed work). One line plus a test alongside the existing gate cases.

## Protocol

BUGFIX (isolated one-line mapping gap with existing test patterns; no hardware gate needed - the resolver is fully unit-tested). Sequence after #1428 merges to avoid actions.ts churn.

amrmelsayed:
## Summary

Stream Deck's `phaseArtifactVerb` did not map the `verify-approval` gate, so a builder blocked there returned `undefined` and the Builder Action Automatic press fell back to opening a terminal instead of the review diff. The #1428 faces already render this gate (yellow + `verified` glyph + `Verify` label), so today the key looks gate-blocked but the press opens a terminal — a face/behaviour contradiction, the same class codex caught during #1428.

Fixes #1431

## Root Cause

`phaseArtifactVerb` (`apps/streamdeck/src/actions.ts`) mapped the `spec-approval`, `plan-approval`, `dev-approval`, and `pr` gates but not `verify-approval` — which the VS Code sidebar's gate vocabulary already knows (`gateIconFor` maps it to the `verified` codicon). The Stream Deck resolver never caught up, so the gate returned `undefined` and both callers that key off it (the Builder Action Automatic press and the phase-aware dial mode) took the fallback path.

## Fix

- Add `verify-approval` to the gate branch, returning `view-diff` (same as `dev-approval`/`pr` — the human is reviewing finished work). The Automatic press then resolves `view-diff` → `open-diff-first` via `BuilderAction.resolveVerb` (#1414), which is correct and intended.
- Add regression assertions beside the existing `dev-approval`/`pr` gate cases in `actions.test.ts`.
- Delete the now-stale `face.ts` doc-comment clause that noted the resolver did not handle this gate yet (BUGFIX #1431) — this change closes that gap.

## Second behavioural effect: review dials

`reviewMode()` derives from the same resolver (`verb === 'view-diff'` → `'diff'`), so this change also flips a `verify-approval`-blocked builder's review dials from `'none'` (dead) to `'diff'` (navigating the diff). This is **correct and desirable** — symmetric with `dev-approval`/`pr`; the dials should work while a human reviews finished work. Pinned with an added `reviewMode` assertion.

## Test Plan

- [x] Regression tests added — `phaseArtifactVerb(verify-approval)` → `view-diff` and `reviewMode(verify-approval)` → `diff` (both fail without the fix: `phaseArtifactVerb` returns `undefined`, `reviewMode` returns `none`)
- [x] Build passes
- [x] check-types clean
- [x] All tests pass (125/125)

### DIFF
--- a/apps/streamdeck/src/__tests__/actions.test.ts
+++ b/apps/streamdeck/src/__tests__/actions.test.ts
@@ -667,6 +667,7 @@ describe('reviewMode (dial mode from the shared resolver)', () => {
     expect(reviewMode(b({ protocolPhase: 'review' }))).toBe('diff');
     expect(reviewMode(b({ blockedGate: 'dev-approval' }))).toBe('diff');
     expect(reviewMode(b({ blockedGate: 'pr' }))).toBe('diff');
+    expect(reviewMode(b({ blockedGate: 'verify-approval' }))).toBe('diff'); // #1431: dials navigate the diff while the human reviews finished work
   });
   it('an unknown phase, no live status, or no builder → none', () => {
     expect(reviewMode(b({}))).toBe('none');
--- a/apps/streamdeck/src/__tests__/actions.test.ts
+++ b/apps/streamdeck/src/__tests__/actions.test.ts
@@ -640,6 +640,7 @@ describe('phaseArtifactVerb (shared resolver — recognised verb or undefined)',
     expect(phaseArtifactVerb(b({ protocolPhase: 'verify' }))).toBe('view-diff');
     expect(phaseArtifactVerb(b({ blockedGate: 'dev-approval' }))).toBe('view-diff');
     expect(phaseArtifactVerb(b({ blockedGate: 'pr' }))).toBe('view-diff');
+    expect(phaseArtifactVerb(b({ blockedGate: 'verify-approval' }))).toBe('view-diff'); // #1431: human reviewing finished work
   });
   it('gate beats phase (the stronger signal)', () => {
     expect(phaseArtifactVerb(b({ blockedGate: 'plan-approval', protocolPhase: 'implement' }))).toBe('open-plan');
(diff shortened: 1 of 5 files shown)


amrmelsayed:
**Integration review (streamdeck architect) — APPROVE.** Low-risk tier, read line-by-line. The change is one clause added to an existing gate branch (`verify-approval` → `view-diff`, symmetric with `dev-approval`/`pr` — the human is reviewing finished work), its test, and deletion of the now-stale `face.ts` comment that pointed at this very issue as an open gap.

**Board was degraded**: codex was environmentally unavailable (OpenAI billing/credits, raised to the owner as an account item), so this shipped on a 2-way CMAP — gemini + claude, both APPROVE HIGH — plus architect source verification, per the interim policy both architects have adopted for low-risk lanes while the outage lasts.

That verification earned its place: the two-seat board missed a real untested behaviour change. Because `reviewMode` derives from `phaseArtifactVerb`, this mapping also flips a `verify-approval`-blocked builder's review dials from `none` (dead) to `diff` (navigating the diff). Correct and desirable — the dials should work exactly when a human is reviewing finished work — but unasserted. Now pinned by a regression assertion (2dfbd81a8) and called out in the PR body.

Closes the face/behaviour contradiction introduced by #1428's faces, which already rendered this gate as blocked (yellow + `verified` glyph + "Verify") while the press fell through to a terminal.

Issue #1410, opened 2026-08-11
Stream Deck SD+: two-zone builder workflow (selectors + action palette, dial-driven feedback queue)


amrmelsayed:
## Problem

We now have one key per builder (#1404: merged Builder Action, phase-aware press that also selects the builder). But covering a builder end-to-end needs several actions (approve a gate, run dev, send review feedback), and the recently added queued code-review feedback has no hardware trigger. We want an efficient SD+ layout where the deck is a remote control and VSCode is the screen, bound by a single shared selection.

## Ground truth

- A shared zoom cursor already exists (`store.cursor`, `selectedBuilder()`); `DevServerAction`, the diff dials, `ScrollNav`, and `ZoomNav` act on it. After #1404, pressing a Builder Action key sets that cursor (`syncToBuilder`).
- Diff dials already use all three gestures each (rotate = navigate, press = `forward-*`, tap = jump-to-first). They are at gesture budget.
- We support **SD+ only** for now: 8 LCD keys (2 rows of 4) + 4 dials + touchstrip.
- Feedback delivery mode (forward immediately vs queue) is a **VSCode workspace setting** the engineer sets; the deck must follow it, not guess.

## Requirements

1. **Two stable zones (no modal re-organization).**
   - **Row 1 (4 keys): fleet selectors** = the #1404 Builder Action keys. Render status; press selects (+ opens the phase artifact). A 4-wide window onto the fleet; the Select dial reaches builders past the first four.
   - **Row 2 (4 keys): action palette**, fixed in place, always acting on the selected builder: **[Approve gate] [Run Dev] [Send Fb (N)] [ Next / free ]**. Rejected: activating a builder to re-flow all keys (modal, kills muscle memory, ambiguous state).
2. **Dials collect, key commits.** Repoint each diff dial press from immediate send to submitting the chunk as feedback via a **mode-neutral verb** (`feedback-file` / `feedback-hunk` / `feedback-selection`). VSCode routes it forward-now or enqueue per the workspace setting. Dial rotate = navigate (unchanged), tap = jump-to-first (unchanged).
3. **Send Fb (N) key = flush.** Fires a `send-queue` verb; badge `N` mirrors the per-builder queued count from the overview. In immediate mode `N` stays 0 and the key is inert; in queued mode `N` climbs and the key sends. No deck-side mode inference.
[...]

Issue #1359, opened 2026-08-05
AIR consultation policy is contradictory: protocol says builder-decides, CLAUDE.md says one-consult-at-PR — pick one and align all surfaces


amrmelsayed:
## The contradiction (surfaced 2026-08-05/06 by two AIR runs)

- `protocols/air/protocol.md` (both trees): comparison table says AIR consultation is **'Optional (builder decides)'**; PR-phase step 3 says *'Optionally run CMAP consultation if the builder judges the complexity warrants it.'*
- `CLAUDE.md`/`AGENTS.md` (PIR section): describes PIR's consult footprint as matching **'BUGFIX/AIR's one-consult-at-PR pattern'** — implying AIR consults at PR as standard.
- `lessons-critical.md` (hot tier, injected into every phase): *'Trust the protocol — never skip CMAP/consultation.'*

Field consequence: air-1352's builder skipped consult per the protocol doc; the architect held the gate citing the lessons-tier rule; the builder accepted a 'misread' that wasn't one (corrected on PR #1355). air-1357 then did the same thing — correctly, per the doc. Enforcement is currently whichever surface the architect read last.

## Evidence for the decision

air-1352's architect-requested consult on a 'trivial' build-script change caught two real stale arch.md build-wiring lines that both the builder's sweep and the architect's review missed — consistent with the hot-tier lesson's claim that consultation catches what solo review misses, even on small changes.

## Options

**A. Make AIR's consult mandatory-at-PR** (align to BUGFIX's 'PR phase only'). Rationale: AIR is small *features*, BUGFIX small *fixes* — AIR being LIGHTER on review rigor than BUGFIX is inverted; the pass is one 2-way CMAP, cheap; it removes the ambiguity and matches CLAUDE.md's existing description and the lessons-tier rule. Cost: adds one consult (~1-3 min) to every AIR.

**B. Keep builder-discretion** and fix the other surfaces: correct CLAUDE.md/AGENTS.md's characterization, and qualify the lessons-critical line ('…where the protocol mandates it'). Cost: preserves a standing trap where the always-injected hot-tier lesson contradicts the protocol a builder is running, and leaves architects without a consistent gate-check expectation.

[...]

PR #1452, opened 2026-08-14
[Air #1390] feat: fold Stream Deck plugin design rationale into README

### LINKED ISSUE #1390: Fold the Stream Deck plugin's design rationale in-tree before the pre-migration repo is archived
## Context

From the pr-gate integration review of PR #1387 (claude lane, non-blocking). `apps/streamdeck/README.md` defers architecture and design decisions to the pre-migration repository's `PLAN.md`, which #1347's plan deliberately did not import. That is a live documentation dependency on a repository whose retirement is already a tracked follow-up of the migration.

## Scope

Move the still-relevant design rationale in-tree — either a Design section in `apps/streamdeck/README.md` or an `apps/streamdeck` subsection of `codev/resources/arch.md` (implementer's call) — and update the README pointer so provenance stays in the History section only.

## Sequencing

Must land **before** the pre-migration repository is archived (that retirement is gated on the next release / the sdk's first npm publish). Natural bundle-mate for the retirement work itself.

amrmelsayed:
## Summary

`apps/streamdeck/README.md` deferred the plugin's architecture and design decisions to the pre-migration repository's planning document. This folds that rationale in-tree as a new **Design** section so the README no longer depends on a repository slated for retirement (gated on the sdk's first npm publish).

Closes #1390

## What Changed

- **New `## Design` section in the README** covering the *why*, reconstructed from the system as it actually ships (the README, `store.ts`/`plugin.ts`, `codev/plans/1347-*`, `codev/reviews/1347-*`, and arch.md's Spec 1410 coherence notes) rather than transcribed from the external planning doc:
  - a stateless controller (holds no Tower state, edits no files) and why that makes the device disposable and every surface consistent;
  - canonical verbs over a command relay instead of driving the editor directly;
  - one shared selection binding Row 1 / Row 2 / dials to a single builder;
  - keys-commit / dials-review mapped to the SD+'s two physical controls, and why review (a scan-and-pick motion) belongs on the encoders;
  - the canvas owning composer/queue state while the deck stays mode-neutral;
  - why gate approval is never silent.
- **Removed all three `PLAN.md` pointers** (intro, Status/roadmap, History) and the external repo name/URL from committed prose. Provenance survives in the **History** section, described in codev's own terms (commit `77be3d0`, `packages/streamdeck`, issues #1347/#1189 retained as codev's reference terms).
- **`readme-design.test.ts`** (new) pins the invariant: a Design section exists, no `PLAN.md` deferral, no external repo name, and provenance stays in History. Mirrors the existing `vendored-ui-lib.test.ts` doc-invariant pattern.

## Key Decisions

- **Home = README Design section** (not an `arch.md` subsection). Per the architect's steer: co-located with the code and free of the COLD arch.md hot/cold tier discipline.
- **Reconstructed from in-tree sources**, not the pre-migration planning doc — I did not assume access to the pre-migration repository, and writing the rationale from the shipped system is more accurate than transcribing a stale plan.
- **Named nothing external** in the prose I authored, per the standing "describe in codev's own terms" rule; the pre-migration repository is referred to as such.

## Test Plan

- [x] Unit tests added (`readme-design.test.ts`, 4 cases)
- [x] Build passes (`pnpm --filter @cluesmith/codev-sdk build && ...streamdeck check-types`)
- [x] All tests pass (`pnpm --filter @cluesmith/codev-streamdeck test`: 166/166 across 9 files)

## Review Notes

Documentation change plus a doc-invariant test; no runtime code touched. Focus review on whether the Design section's *why* matches the shipped behavior and reads as rationale rather than a second feature list.

### DIFF
--- a/apps/streamdeck/README.md
+++ b/apps/streamdeck/README.md
@@ -10,8 +10,71 @@ and runs it. The plugin holds no Tower state and never edits files directly.
 - **Auth**: reads `~/.agent-farm/local-key` and sends it as the `codev-web-key`
   header (never generates it — Tower owns the key).
 
-Architecture and design decisions live in the pre-migration repo's `PLAN.md`
-(see [History](#history)).
+The **[Design](#design)** section explains why the plugin is shaped this way; the
+[History](#history) section records where it came from.
+
+## Design
+
+The rest of this README is the *what*; this section is the *why*. These decisions
+shape everything above.
+
+**A stateless controller, not a second client.** The plugin holds no Tower state
+and edits no files. It is a remote: it reads the overview Tower already computes
+and POSTs verbs Tower already routes. Everything authoritative lives elsewhere.
+Tower owns workspace, builder, and gate state plus the auth key; VSCode owns the
+working tree, the review queue, and the composer. The deck only projects that
+state onto keys and dials and fires intents back. The payoff is that the device
+is disposable: unplug it, restart it, or run two of them, and nothing is lost or
+forked, because there was never a second copy of the truth to reconcile. It also
+means every surface (deck, VSCode sidebar, web dashboard) reflects one state, so a
+change made on any of them shows up on all of them.
+
+**Canonical verbs over a command relay, not direct editor control.** The plugin
+never speaks to VSCode directly. It POSTs a small, fixed vocabulary of canonical
+verbs to Tower's command relay (`/api/command`), and the focused VSCode window
+maps each verb to a `codev.*` command. The decoupling is deliberate: the deck does
+not need to know how many editors are open, which one is focused, or how a command
+is implemented. Tower routes the verb to the focused window, and each verb is
+stamped with the active workspace, so one Tower serving several workspaces sends
+each command to the right place. New editor behavior is a new `codev.*` handler,
+and the deck's verb set barely moves.
+
+**One shared selection, because the board is a single instrument.** Row 1 selects a
+builder, Row 2 acts on it, and the dials review it, with all three pointed at the
+same builder. That coherence is the design, not a coincidence: a Row 1 press is
+select-and-open in one gesture, and focusing a builder's diff or canvas in VSCode
+moves the deck's selection back to it (via the `builder-active` activity hook).
+Without one binding selection the three zones would drift and every press would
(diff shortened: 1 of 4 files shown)


amrmelsayed:
**Integration review (streamdeck architect) — APPROVE.**

The section is *why*, not a second feature list, which is the thing this issue could easily have gotten wrong. It argues four decisions — stateless controller, canonical verbs over the relay, one shared selection, keys-commit/dials-review — and each is argued from consequence rather than asserted. The dependency the issue existed to kill is gone: no pointer to an external planning document, and provenance now reads in codev's own terms in History (imported under #1347 at `77be3d0`, retirement gated on the sdk's first npm publish).

**Claims verified against the code**, since the real risk in a reconstructed rationale is confident inaccuracy rather than bad prose:

- *"POSTs canonical verbs … and each verb is stamped with the active workspace"* — accurate. `TowerClient.sendCommand` populates `body.workspace` when given one, and the deck passes `selectedWorkspacePath()` at every call site; the VSCode relay drops any command whose workspace doesn't match its own.
- *"the focused VSCode window maps each verb to a `codev.*` command"* — accurate, and `VERB_COMMANDS` doubles as the allowlist, which is what makes "the deck's verb set barely moves" true rather than aspirational.
- *"focusing a builder's diff or canvas moves the deck's selection back to it"* — accurate as of #1410 §F, including the canvas half.

**By-product worth fixing separately (not in this PR):** `apps/vscode/src/command-relay.ts:78` still carries the comment *"absent today, so this is a no-op until a controller populates it"* about workspace stamping. That comment predates the deck populating the field, which it now does at every call site — so the README is right and the code comment is stale. One-line fix on the vscode surface; flagged to that lane rather than expanded into this docs lane.

**On the consultation**: this lane ran without a CMAP pass. To be precise about provenance, since it has been attributed to me: I did not instruct a skip when scoping — the builder skipped it and I am ratifying that now, deliberately, rather than retroactively. The reasoning: for docs-only prose the failure mode is factual drift from the system it describes, and the check that catches that is verifying each claim against the code, which I have done above and recorded. Nothing here touches executable behaviour. That said, the general question of when AIR requires a consult is genuinely unresolved (#1359) and this lane is now a datapoint there, not a precedent I would extend to any PR that touches code.

PR #1453, opened 2026-08-14
[Air #1444] Stream Deck: re-glyph the catch-all Codev Action off the terminal picture

### LINKED ISSUE #1444: Stream Deck: re-glyph the catch-all 'Codev Action' — its terminal icon now collides with Open Terminal
Surfaced by #1440 (PR #1443) and confirmed by side-by-side comparison of the rendered assets.

**Problem.** The catch-all `Codev Action` uses a **terminal glyph** (`icons/action.*`). #1440 gave the new `open-terminal` action its own dedicated icon rendered from the `terminal` glyph — correctly, since that action literally opens a terminal and its key face draws the same glyph. Result: the two entries look near-identical in the Stream Deck action picker. (Not a regression — before #1440 they *shared* the single `action` asset — but the ambiguity that motivated the icon work survives for this pair.)

**Recommendation.** Re-glyph **Codev Action**, not Open Terminal. The terminal picture semantically belongs to the action that opens a terminal; `Codev Action` is a configurable catch-all that runs any verb (open terminals, view diff, send, spawn, refresh), so a terminal image was always a mislabel for it. Cheapest option worth trying first: the existing Codev brand mark (`icons/plugin.svg`) — a generic Codev action wearing the Codev mark reads correctly and needs no new artwork. Needs a visual check at whatever size it lands.

**Notes for whoever picks it up.** `scripts/render-action-icons.mjs` (from #1440) already does glyph→PNG at the right sizes and fill fractions; extend it rather than hand-rolling. Verify whether the Codev Action key renders a composite face at runtime before assuming the manifest asset is all that shows. Natural home: the next Elgato pack session, alongside the deferred pre-populated SD+ profile.

amrmelsayed:
Closes #1444.

## Problem

The catch-all **Codev Action** drew a **terminal glyph** (`icons/action.*`). #1440 gave the new `open-terminal` action its own dedicated terminal-glyph icon — correctly, since that action literally opens a terminal. Result: the two entries look near-identical in the Stream Deck action picker. The terminal picture was always a mislabel for `Codev Action`, a configurable catch-all that runs any verb (open terminals, view diff, send, spawn, refresh).

## Change

Re-glyph the **catch-all**, not `open-terminal`, onto the existing Codev brand mark (`icons/plugin.svg`) — the architect's first candidate. A generic Codev action wearing the Codev mark reads correctly and needs no new artwork.

- **Extends `scripts/render-action-icons.mjs`** (from #1440) with a brand-mark source (`extractBrandMark`) that flows through the *same* `trim → fit → composite` pipeline as the face.ts glyphs (`KEY_FILL 0.56`, `LIST_FILL 0.94`, sizes 72/144/20/40). No hand-rolling, no new dependency. The glyph path is refactored into a shared `build()`/`emit()` helper so both sources fit and center identically.
- **Regenerates the four `action` assets** (`icons/action{,@2x}.png`, `icons/list/action{,@2x}.png`) from the mark.
- **No manifest change.** `Codev Action` already references `icons/list/action` + `icons/action`; only the pixels behind those filenames change. No other manifest action is touched.

## Verification (issue point 3) — done before coding

Does the Codev Action key render a composite face at runtime (like Builder Action / Send Fb / Run Dev do)? **No.** `CodevAction` (`src/actions.ts:63-66`) extends `VerbKey` and does **not** override `onWillAppear` — only `DevServerAction`/`BuilderAction`/gate keys composite a face via `setImage`. So the Codev Action key face shows the manifest `States[0].Image` (`icons/action`) directly.

**=> This is NOT picker-only.** The re-glyph fixes both the Stream Deck action-picker list icon *and* the physical key face.

## Legibility check (does `plugin.svg` survive?)

Rendered and eyeballed at delivery sizes. The handshake mark is clean and unmistakably not-a-terminal on both the 72/144 key face and — the critical case — the 20×20 picker list icon. It survives; no alternative artwork needed.

| Key face (144) | Picker list icon (20, on picker ground) |
|---|---|
| Codev handshake mark, centered on the rounded `#1C2128` ground | Handshake still recognizable at 20px |

## Diff scope note

Re-running the #1440 render script also re-encodes the `send-queue`/`open-terminal` PNGs with different ImageMagick metadata bytes, but those are **pixel-identical** to the committed versions (verified `compare -metric AE = 0` across all 8 variants). I reverted them so this diff is scoped to the `action` assets alone.

## Tests

- `render-action-icons.test.ts`: `extractBrandMark` pulls the mark's `<g>` group + viewBox, leaves the opaque background `<rect>` behind (transparent render), carries no `${c}` recolor placeholder, and throws loudly if `plugin.svg`'s shape drifts. Plus a source-level regression guard: `action` routes through `BRAND_ICONS`, never the glyph `ICONS` list.
- `manifest-icons.test.ts`: the `action` assets ship at the convention sizes (72/144/20/40), keep their manifest filenames, and are no longer byte-identical to the `open-terminal` assets.
- Full suite: **172 passed**. `check-types` + `build` green.

## Review notes

Straightforward asset regeneration behind a one-source render extension; low blast radius. Given AIR's economy and the purely-declarative, visually-verified nature of the change (with source- and manifest-level test guards), I did not run a full 3-way CMAP.


amrmelsayed:
**Integration review (streamdeck architect) — APPROVE.** Assets + render script only; verified by viewing the rendered PNG, which is the only real check for an image change.

- **Re-glyphed the right side of the collision.** The catch-all moves off the terminal picture; `open-terminal` keeps it. That is the semantically correct assignment — the terminal image belongs to the action that opens a terminal, and `CodevAction` is a configurable runner of any verb (its default is `refresh-overview`), for which a terminal was always a mislabel.
- **Brand mark, no new artwork.** It renders `icons/plugin.svg` — the Codev mark — through #1440's existing `render-action-icons.mjs` at the established sizes and fill fractions. No new dependency, no hand-drawn asset, and the script keeps being the one place icons come from.
- **The verify-first question came back the opposite way to my assumption, and I confirmed it in source.** `CodevAction extends VerbKey`, which implements `onKeyDown` only — no `renderTo`, no `setImage`, no `onWillAppear` override. So unlike Builder Action / Send Fb / Run Dev, this key never paints a runtime composite face, and its manifest `States[].Image` **is** what shows on the hardware permanently. This fix is therefore picker *and* key-face, not picker-only. My framing on #1440 ("a setup-experience fix, not an on-hardware one") was right for the actions that composite their faces and wrong for this one.

Consultation: full 3-way available again (codex restored). 172 tests green, build clean.

Issue #1465, opened 2026-08-15
Stream Deck: size the Row-1 builder window from placed keys — a builder can be selected while shown on none


amrmelsayed:
## Problem: a builder can be selected while rendering on no key

Row-1 selection pages by a hardcoded constant, not by how many builder keys exist. `ROW1_WINDOW_SIZE = 4` (`apps/streamdeck/src/store.ts:28`), `builderWindowStart()` returns `floor(cursor.builder / 4) * 4`, and `windowedBuilder(slotIndex)` returns `builders()[windowStart + slotIndex]`. The page step is fixed at four and is **independent of how many `BuilderAction` keys the user actually placed**.

The consequence is a correctness bug, not a sizing preference. Place three builder keys instead of four and every builder at index ≡ 3 (mod 4) renders on **no key at all** — while the Select dial still walks onto it, so it becomes the selected builder, drives Row 2's entire palette and both review dials, and shows no accent ring anywhere on the board. The reviewer is acting on a builder the deck never displays. (There is no gap only while the fleet is no larger than the number of placed slots.)

It is also wrong for most hardware: Mini is 3×2, Standard 5×3, XL 8×4, and the SD+ keypad 4×2, so a constant of four fits exactly one device.

## Why now

The trigger is concrete rather than hypothetical: #1463 adds an Open Architect key that the owner intends to place on a **Row-1 key** (safe, because its Main mode is selection-independent), which leaves three builder slots against a fixed page of four — the exact configuration that hides every fourth builder. The fix belongs to Row-1 windowing, not to the architect key, so #1463 ships with the fixed-4 behaviour documented and this issue carries the change.

## Feasibility (verified against `@elgato/streamdeck@2.1.0`)

There is **no profile-structure API** — `plugin/profiles.d.ts` states verbatim that plugins "may only switch to profiles distributed with the plugin, as defined within the manifest, and cannot access user-defined profiles", and `switchToProfile` is the only profile call. So the layout cannot be read directly.

It can be *derived from the lifecycle*, which is the workable path: every visible `BuilderAction` fires `willAppear` carrying `KeyAction.coordinates` (`{column, row}`) and its device, and `streamDeck.devices` exposes `Device.size` (`{columns, rows}`) and type.

## Proposed shape

- Window size = the number of currently visible `BuilderAction` instances, replacing the constant.
- Slot order derived by sorting visible instances by `(row, column)`, which could retire the manual `slot` PI field or keep it as an override.
[...]

Issue #1406, opened 2026-08-11
afx spawn silently attributes the builder to 'main' when CODEV_ARCHITECT_NAME is absent, even inside another architect's session


amrmelsayed:
## Observed (2026-08-11, live)

The streamdeck architect ran `afx spawn 1403 --protocol air` from its own architect session. Tower registered builder-air-1403 with owner **main**; the builder's reports landed on main's terminal, and main had to ask who owned it.

## Mechanism

`packages/codev/src/agent-farm/commands/spawn.ts:26-39`: the spawning architect's name is read once at module load from `CODEV_ARCHITECT_NAME` (Spec 755) and **silently** falls back to `'main'` when the var is empty. In the reporting session the var WAS empty, while `afx whoami` run from the very same shell resolved the identity correctly (`type: architect, name: streamdeck`), so Tower-side resolution works where the env-var path failed. Shells spawned by tools/subprocesses inside an architect terminal do not reliably carry the injected var.

## Consequences of misattribution

- Builder reports route to the wrong architect's terminal.
- The builder's `afx send architect` affinity points at the wrong architect, and Tower's spoofing check then actively **blocks** the builder from addressing its real spawner via `architect:<name>`.
- Review/gate flow lands on an architect who doesn't own the work.

## Suggested fix

Resolve the spawning architect the way `afx whoami` does (Tower PID/terminal lookup), with the env var as a fast path, and keep the `'main'` fallback only for genuinely-outside-any-architect-terminal invocations. At minimum: when falling back to `'main'`, print a visible warning naming the fallback, so misattribution is caught at spawn time instead of at first report.

## Workaround until fixed

Prefix spawns with the explicit var: `CODEV_ARCHITECT_NAME=<name> afx spawn …`.

Related: #1391/#1392 (workspace registration edges), #1393 (gate-relay semantics); same family of multi-architect attribution seams.

amrmelsayed:
[...]

PR #1466, opened 2026-08-15
Stream Deck: Open Architect Terminal key (Builder/Main modes) + paired terminal-key rename

### LINKED ISSUE #1463: Stream Deck: key to open the architect that spawned the selected builder
## Gap

The deck can reach every builder — select it, open its artifact, approve its gate, run its dev server, send it feedback, open its terminal — and cannot reach a single architect. But architects are where the conversation happens: when a review raises something the builder can't settle, the next move is talking to whoever owns the lane, and that requires leaving the deck for the mouse.

## Design: builder-scoped, like the rest of Row 2

A new key whose press opens the terminal of **the architect that spawned the currently selected builder**. It follows the Row 2 rule exactly — every key there acts on the selected builder — so architects never enter the selection model, Row 1 stays builder-only, and no key's meaning depends on what kind of thing is selected (the modality #1410 rejected).

You never pick an architect; you pick a builder, and the key follows the ownership the system already records. That scales to any number of architects at zero cost and reads as the natural escalation: *I am reviewing this work, now talk to whoever owns it.*

**Everything needed already exists:**

- `open-architect-terminal` is in the relay allowlist and the deck's PI verb list.
- `codev.openArchitectTerminal` accepts an optional architect name (`extension.ts`).
- `OverviewBuilder.spawnedByArchitect: string | null` is already on the wire the deck reads — **no types, Tower, or relay change.**

**Implementation shape:** a `VerbKey` subclass whose `args()` returns `[selectedBuilder.spawnedByArchitect]`, mirroring how Row 2's Open Terminal passes the selected builder id. It should render its target's name on the face (`labelFaceSvg`), so you can see who you would summon before pressing — note that `VerbKey` has no runtime face, so this key must opt into compositing (see #1444 / #1459).

## Rulings to honour

1. **`spawnedByArchitect` is `null`** (builder spawned outside an architect context): the key is **inert with a status message**, not a silent fallback to `main`. A key that sometimes means "the owner" and sometimes "main" is exactly the quiet inconsistency this board's design avoids.
2. **Architect not live / unresolvable**: defer to the VS Code command's existing handling; the deck must not second-guess it.
3. **No builder selected**: inert, consistent with the rest of Row 2.

## Dependency: #1406 is a prerequisite, not a footnote

`#1406` (spawn attribution silently falls back to `main` unless `CODEV_ARCHITECT_NAME` is set) is today a cosmetic wrinkle in `afx status`. This feature turns it into a **wrong action on hardware** — pressing "talk to the owner" and summoning the wrong architect. `air-1403` is a real instance: spawned by the streamdeck architect, registered as main's. Fix or confirm #1406 before this ships, or accept that the key inherits its inaccuracy.

## Alternatives considered and rejected

- **Statically configured per-architect keys** (PI carries an architect name). Rejected: it breaks the Row 2 invariant by making one key context-free while its neighbours are builder-scoped, and it needs a folder or several slots to cover more than one architect.
- **An architect altitude on the Zoom dial.** Rejected on reading the code: that dial's cursor **is** the shared selection (`cursor.builder` → `selectedBuilder()` → what Row 2 and both review dials act on), so putting architects there either moves the selection off a builder and hollows out Row 2, or needs a parallel cursor plus an altitude whose tap means "summon" where every other altitude's tap means "descend". Its gestures are also fully allocated (rotate walks, tap zooms in, press zooms out).

## Placement

The plugin does not dictate placement — bundled profiles ship `Actions: null` by long-standing precedent (#1404, re-ruled at #1410's dev-approval), so keys are placed by the user in the Stream Deck app. Guidance for the README:

- Recommended home is **Row 2** alongside the other builder-scoped actions.
- The natural donor is **Send Feedback** *while the workspace is in forward mode*, where that key is inert by design (badge pinned at 0, press does nothing). Unplacing a key does not remove the action — it stays available in the picker and can be re-placed in seconds if the workspace moves to queue mode.
- The trade disappears entirely with a larger profile (#1381), which this feature is a concrete argument for.

## Protocol

PIR — the key renders a composite face and its press behaviour is worth seeing on hardware before a PR exists, matching #1428's precedent for face work.

amrmelsayed:
# PIR Review: Stream Deck key to open the architect that spawned the selected builder

Fixes #1463

## Summary

Adds a Row-2 Stream Deck key, **Open Architect Terminal**, that opens an architect's
terminal in VS Code via the existing `open-architect-terminal` verb. It has two
Property-Inspector modes: **Builder** (default — follows the selected builder and
opens the architect that spawned it, `spawnedByArchitect`; inert when nothing is
selected or the builder has no owner) and **Main** (always the workspace's `main`).
The key renders a live face — the constant title `Architect` over the resolved
architect's name — so you see who a press would summon. As a coherence fix, the two
terminal keys were renamed as a symmetric pair: the pre-existing **Open Terminal**
became **Open Builder Terminal** (face `Terminal` → `Builder`, terminal glyph kept).

## Files Changed

Code (`apps/streamdeck`), vs merge-base `3f7061d4a`:

- `src/actions.ts` (+86 / -8) — new `OpenArchitectAction`; builder-key doc + face label rename
- `src/face.ts` (+46 / -8) — `architect` glyph, `capitalizeFirst`, `architectFaceSvg`; optional dim color on the line helpers
- `src/plugin.ts` (+2) — register `OpenArchitectAction`
- `com.cluesmith.codev.sdPlugin/manifest.json` (+20 / -3) — new action entry; paired rename of the two keys
- `com.cluesmith.codev.sdPlugin/ui/open-architect.html` (+26, new) — Property Inspector (Builder/Main select)
- `scripts/render-action-icons.mjs` (+1) — add `open-architect` to `ICONS`
- `icons/{,list/}open-architect{,@2x}.png` (4 new) — rendered from the `architect` glyph
- `README.md` (+39 / -10) — key docs, both renames, layout diagram, known edges, Row-1 caveat
- `src/__tests__/{actions,face,manifest-icons,render-action-icons}.test.ts` (+152) — behavior + face + icon tests
- `codev/resources/arch.md`, `codev/resources/lessons-learned.md` — cold-tier notes (see below)

## Commits

- `44726b1fe` Add architect glyph + architectFaceSvg (title/subtitle, dim None) and its rendered icons
- `af32dba3d` OpenArchitectAction: two-mode (Builder/Main) Row-2 key, dynamic face, inert when unresolved
- `361af2633` Manifest action + Property Inspector (target: Builder/Main) for Open Architect
- `2f23f332c` README: document Open Architect key, its modes, and known edges
- `d0ae5b9ce` Rename the terminal keys as a pair: Open Builder Terminal / Open Architect Terminal (face 'Terminal'→'Builder')
- (plan/thread commits omitted)

## Test Results

- `pnpm --filter @cluesmith/codev-streamdeck build`: ✓ pass
- `pnpm --filter @cluesmith/codev-streamdeck check-types` (tsc): ✓ pass
- `pnpm --filter @cluesmith/codev-streamdeck test`: ✓ pass (197 tests, 11 new)
- `pnpm --filter @cluesmith/codev-streamdeck validate` (Elgato): ✓ pass
- Manual (dev-approval, hardware, approved by Amr): both PI modes, the inert `None`
  face, the missing-main face, and the already-placed Open Terminal key surviving
  the rename (now reads `Builder`).

## Architecture Updates

**HOT (`arch-critical.md`): none.** This is one key's behavior, not a system-shape
invariant, and the hot file is at its cap.

**COLD (`arch.md`, Integration Points → Stream Deck ↔ VSCode coherence):** added a
bullet recording the key, its two modes, and — the decision a future reader would
otherwise re-derive — **why the deck delegates `main`-else-first to VS Code instead
of resolving it itself**: VS Code's `openArchitectTerminal` already special-cases
the literal `'main'`, so keeping the policy there means *one policy, one home*, and
it is precisely what let the deck **stop consuming `OverviewData.architects` (the
live-architect view) entirely** — dropping the first-live fallback and a store
reader that would have coupled the deck to a transiently-wrong view. The bullet also
records the two accepted residuals and the UUID-stability property (below).

## Lessons Learned Updates

**HOT (`lessons-critical.md`): none** — the core reuse decision ("delegate the
policy to its single existing home rather than duplicate it on the client") is
already covered by the standing hot lesson *"Single source of truth beats
distributed state."* This PR is an instance of it, not a new rule.

**COLD (`lessons-learned.md`, UI/UX):** added a lesson that a Stream Deck action's
identity is its **UUID, not its `Name`** — renaming `Name`/`Tooltip`/face leaves
already-placed keys working as long as the UUID is unchanged; change the UUID and you
orphan them. Included the `render-action-icons.mjs` gotchas (glyph line must carry no
trailing comment; the script re-touches all icons, so restore byte-churned
pre-existing PNGs to keep the diff scoped).

- **The face is the safeguard, not decoration.** The resolved architect name on the
  subtitle is the only thing between the user and a wrong-architect press, so it must
  render in every mode including the inert `None` state. `architectFaceSvg(undefined)`
  is the dim/`None` face; `face.test.ts` pins all three (`Main`, a sibling name,
  `None`).
- **`resolve()` returns `string | undefined`.** `undefined` (Builder mode, no
  selection / no owner) is the only inert path; Main mode always resolves to `'main'`.
  A named-but-not-live target is *not* inert — the name flows to VS Code, which owns
  the not-found warning (deliberate; ruling 2).
- **Two accepted residuals (documented, not chased):** (1) in **Main** mode when
  `main` is absent, VS Code opens the first live architect while the face still reads
  `Architect / Main` — the relay is fire-and-forget, so the deck never learns which
  opened; narrow (a workspace with architects but no `main`) and pre-existing VS Code
  behaviour, not introduced here. (2) A live architect registration behind a **dead
  PTY** resolves fine and opens a terminal nobody reads — the deck can't detect it.
  Deck-side verification would reintroduce exactly the live-view coupling this design
  removed, so both are limitations, not oversights.
- **The paired rename is UUID-stable.** Both UUIDs (`open-terminal`,
  `open-architect`) are unchanged; only `Name`/`Tooltip`/face moved. That is what
  protects keys users already placed, and it's invisible in a diff — verified on
  hardware at dev-approval.
- **`#1406` is a live prerequisite, not a footnote.** Spawn mis-attribution (a
  builder registered under `main` when `CODEV_ARCHITECT_NAME` is absent) makes
  **Builder** mode summon the wrong architect. The face shows the (currently wrong)
  stored owner, which makes it *noticeable* before a press but cannot *correct* it.
  This key should not ship broadly until #1406 is fixed or confirmed.
- **Out of scope, tracked in #1465:** Row 1's builder window is a fixed page of four,
  so placing the Main-mode key on a Row-1 key leaves three builder slots and hides
  every fourth builder past a three-builder fleet. Self-sizing the window is #1465.

## How to Test Locally

- **View diff**: VSCode sidebar → right-click builder `pir-1463` → **Review Diff**
- **Run dev**: VSCode sidebar → **Run Dev**, or `afx dev pir-1463`
  - Builder mode (default): the face reads `Architect / <owner>` and the subtitle
    tracks the Row-1 selection across builders owned by different architects; press
    opens that architect.
  - Builder mode, no owner / no selection: face dims to `Architect / None`; press is
    inert.
  - Main mode: face reads `Architect / Main`; press opens `main` regardless of
    selection.
  - The already-placed **Open Terminal** key still works and now reads `Builder`
    (UUID stability).


amrmelsayed:
**Integration review (streamdeck architect) — APPROVE.** Medium tier (new key + governance docs + a rename touching a shipped key); reviewed continuously through this lane's four design revisions, and re-verified at source here.

**The design invariant holds where it matters most.** `resolve()` is shared by the press path and `renderTo`, so what the face shows and what a press sends cannot diverge — that is the whole safeguard against a wrong-architect press, and it is structural rather than a matter of keeping two code paths in step. Resolution reads exactly as designed: `target === 'main'` fires the literal name and lets VSCode apply main-else-first; otherwise the selected builder's `spawnedByArchitect`; otherwise `undefined` → visibly inert.

**Verified rather than accepted:**

- **UUID stability.** `com.cluesmith.codev.open-terminal` and `.open-architect` are untouched; the manifest diff on the renamed key is `Name` and `Tooltip` only. Already-placed keys survive, which is the sole user-visible risk in the paired rename.
- **The deck never consumes the live-architect view.** No `OverviewData.architects` reader was added — which is what makes the three registry failure modes raised during review (empty list, transiently missing `main`, live-row-behind-dead-PTY) *not this key's problem*.
- **Icons** come from `render-action-icons.mjs` rather than new artwork, keeping #1440's single-source pipeline intact.

**On claude's non-blocking note** (the paired rename sat outside the approved plan's file list): correct as an observation, and worth stating why it is nonetheless the right call rather than merely a permitted one. This feature *created* the ambiguity between the two terminal keys, so the lane that introduced it is the honest place to resolve it; leaving a newly-confusing pair behind and filing a follow-up would have shipped a known defect for process tidiness. It was owner-directed at the hardware gate, UUID-stable, and documented — which is the shape a mid-lane scope addition should have.

**Governance docs are correctly tiered**, which is the part a consult pass does not police. Both entries went to the COLD tier: the `arch.md` entry records the *decision* (why main-else-first lives in VSCode — one policy, one home) plus the two accepted residuals and the #1406 prerequisite, rather than restating the feature; the `lessons-learned.md` entry generalises past this lane (UUID is identity, `Name` is a label; the property is invisible in a diff, so the hardware check is what proves it). Neither belongs in the capped hot tier, and neither was put there.

The lessons entry also carries a genuinely reusable trap the lane discovered — the icon script regenerates *all* icons and parses glyph lines out of `face.ts`, so a new glyph must be a single line with no trailing comment, and pre-existing PNGs it re-touches should be restored to keep a diff scoped. That is the kind of thing that costs the next person an hour and is invisible until it bites.

Consultation: full 3-way, all APPROVE (codex availability restored).

Issue #1446, opened 2026-08-13
porch: a completed project's status.yaml on main never records its own merge (post-merge stamp orphans by construction)


amrmelsayed:
Every lane's final porch write is a `chore(porch): <id> PR #<n> merged` commit that touches only `codev/projects/<id>-<slug>/status.yaml`. Because it necessarily happens **after** the PR merges, it lands on a branch that is already merged and is then deleted at cleanup — so it can never reach `main`. Observed independently on pir-1191 and pir-1410; by construction it applies to every completed lane.

**Effect on main's record.** The shipped copy of a completed project is frozen one step short of done. For `1191` (merged, issue closed, protocol verified) main carries:

```yaml
phase: verified
build_complete: false
pr_history:
  - { phase: review, pr_number: 1427, branch: builder/pir-1191, created_at: ... }   # created, never merged
pr_ready_for_human: false
```

Sampling completed projects on main, 12 of the first 40 with a `pr_number` show `build_complete: false`. So anyone reading project history from `main` — a person, a future dashboard, or an audit script — sees a repo full of lanes that appear to have stalled at 'PR opened'.

**The question to settle is which of two things `codev/projects/*/status.yaml` is:**

1. *Ephemeral orchestration state* — in which case shipping it to main at all is arguably noise, and the durable record is the PR + issue + review artifact (already true today).
2. *Part of the shipped project record* — in which case the merge outcome belongs in it, and the final stamp needs to land somewhere reachable: write the merge record **before** merging (the PR carries it), or have the merge step amend it on main, or drop the fields that can never be accurate post-merge.

Not urgent and nothing is lost today (the PR and issue are the real record), but the current shape quietly publishes a wrong answer, and it is cheap to make honest either way. Surfaced by the streamdeck architect during pir-1410 cleanup; filed here as porch's own design call.

amrmelsayed:
**Supporting data from the streamdeck lane — the record on main is not uniformly wrong, it is *inconsistently* wrong, which sharpens the diagnosis.**

Sampling the six lanes I shipped in the last 48h (all merged, all issues closed, all verified):

| project | `phase` on main | `build_complete` | merge recorded |
|---|---|---|---|
| 1425 | verified | false | no |
| 1414 | verified | false | no |
| 1428 | verified | **true** | no |
| bugfix-1437 | verified | false | no |
| bugfix-1433 | **pr** | false | no |
| 1410 | verified | **true** | no |

[...]

PR #1467, opened 2026-08-15
[PIR #1463 follow-up] Review/arch: capture the key reuse lesson (avoid live-architect dependency; resolve() = structural safeguard)

amrmelsayed:
Doc-only follow-up to #1466 (merged). Adds the reuse lesson the architect asked to preserve in the durable artifacts, for the next person adding a Stream Deck key:

- **Avoiding the live-architect dependency beat handling it carefully.** The deck fires a *name* and delegates `main`-else-first to VS Code, so the three registry failure modes (empty list, transiently-missing `main`, live row behind a dead PTY) are simply not this key's problem rather than mitigated by it — reaching for `OverviewData.architects` would re-import all three.
- **The wrong-architect safeguard is structural, not conventional.** `resolve()` is shared by the press path (`onKeyDown`) and `renderTo`, so face and action cannot drift — one resolution, the name you see is the name you open.

Added to `codev/reviews/1463-*.md` and the `arch.md` Stream Deck integration bullet. No code change.

Refs #1463.

### DIFF
--- a/codev/projects/1463-stream-deck-key-to-open-the-ar/status.yaml
+++ b/codev/projects/1463-stream-deck-key-to-open-the-ar/status.yaml
@@ -21,10 +21,12 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-15T02:54:37.170Z'
-updated_at: '2026-08-15T08:45:59.435Z'
+updated_at: '2026-08-15T08:50:09.703Z'
 pr_history:
   - phase: review
     pr_number: 1466
     branch: builder/pir-1463
     created_at: '2026-08-15T08:13:46.320Z'
+    merged: true
+    merged_at: '2026-08-15T08:50:09.703Z'
 pr_ready_for_human: false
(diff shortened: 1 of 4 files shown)


amrmelsayed:
**Integration review (streamdeck architect) — APPROVE.** Docs only: the review artifact, the `arch.md` Stream Deck bullet, and protocol bookkeeping. No code, no manifest, no assets.

It carries the two decisions this lane produced that a future contributor would otherwise re-derive — **avoiding the live-architect dependency beat handling it carefully**, and **the wrong-architect safeguard is structural rather than conventional because `resolve()` is shared by the press and render paths**. Both are worth the words precisely because the next person adding a key will reach for `OverviewData.architects` by default, and because a safeguard that holds by construction reads identically in a diff to one that holds by discipline.

**On why this is a second PR at all — that is my sequencing, not the builder's.** I asked for these additions *after* #1466 had already merged, so the artifact they belong in had shipped. Asking a phase earlier, before the pr gate, would have carried them in the original PR. This is the same shape as the post-merge porch stamp discussed on #1446: an artifact the protocol produces after its own completion has no authorised way home, and it lands as an unaccompanied follow-up PR with no gate record to corroborate against. Two live instances of it now sit open at once (this and #1464).

Practice change on my side: review-artifact content requests go to the builder **before** the pr gate, not after the merge.

Issue #1483, opened 2026-08-17
Tracking: mailbox-first send follow-ups — gate correctness, interrupt semantics, attribution, visibility (post-1313)


mohidmakhdoomi:
## What

Umbrella for the 13 follow-on issues to Spec 1313's mailbox-first `afx send` (PR #1330, merged). They are not independent: they sit on one pipeline — **sender identity → mailbox row → render-gate classify → write edge → visibility surfaces** — and several share root causes, files, or a required landing order. This issue holds the map, the dependencies, and the suggested sequence.

```mermaid
graph LR
  A[sender identity<br/>#1478 #1475] --> B[mailbox row / schema<br/>#1476 enabler]
  B --> C[render-gate classify<br/>#1482 #1361 #1474 #1473 #1471]
  C --> D[write edge<br/>#1365 → #1481]
  D --> E[visibility surfaces<br/>#1450 #1478 #1472]
```

### A — Gate correctness (the false-verdict family)

The gate's verdict rests on three legs: the mirror's **dims** match the TUI's real layout, the mirror's **content** is coherently seeded, and the **profile** matches the app. Each issue here is one leg failing; all but one fail toward indefinite `held(busy)` (mail delayed forever), and one fails toward false-CLEAN (misdelivery — the worse direction).

- [x] #1482 — **dims leg.** Tower-side `cols/rows` can silently diverge from the kernel PTY: `PtySession.resize()` (`packages/codev/src/terminal/pty-session.ts:567-580`) assigns `this.cols/this.rows` **and** resizes the gate mirror (`_gateScreen?.resize`) *before* the shellper resize's drop-boolean is known, and all callers ignore that boolean — so a dropped resize re-wraps the classification mirror at dims the real TUI never adopted, and a small disagreement flips the verdict to indefinite BUSY. Also carries the **diagnostic sub-item** (see "first slice" below). **(PIR #1482 → PR #1604, merged 2026-09-04: migration v18 persists `GateVerdict.detail` as a `reason:detail` sub-code across `afx inbox`/`afx send`/dashboard/VSCode via one shared sdk formatter, and `resize()` now confirms the shellper resize before moving Tower dims or re-wrapping the gate mirror.)**
[...]

Issue #1365, opened 2026-08-07
Serializer convergence: route mailbox write edge through submitToSession (serialize gated deliveries vs interrupt/escape)


mohidmakhdoomi:
## Summary

The gated mailbox delivery path and the `escape`/`interrupt` submission path take **two disjoint locks**, so a gated mailbox delivery can interleave with a concurrent `--interrupt` / `--escape` write to the same terminal. This is a known, deliberately-accepted boundary today (a gated delivery only ever writes onto a render-verified empty prompt, and `interrupt`/`escape` are explicit human gate-bypasses), but it is worth converging so the write edge is serialized end-to-end.

Split off from PR #1330 (Spec 1313 maintainer round) as agreed — **not** in scope for that PR.

## Detail

- Mailbox deliveries run through a **per-agent** write serializer (`write-queue.ts`, keyed by `agentKey`; completion-chains a message's text + Enter). Two mailbox deliveries to one agent cannot interleave.
- `escape` / `interrupt` submissions take a **per-terminal** submission lock (`submitToSession`, `session-submit.ts`).
- These are **disjoint** locks. A gated mailbox delivery is therefore *not* serialized against a concurrent `escape`/`interrupt` to the same terminal.

The boundary is already flagged in-code:

- `packages/codev/src/agent-farm/servers/session-submit.ts` (the "Exactly what it covers — this is NOT blanket per-session atomicity" comment) documents that the mailbox delivery path is deliberately not covered by the submission lock, and why the residual cross-path race is accepted rather than a regression.
- `packages/codev/src/agent-farm/servers/tower-routes.ts` (the immediate `escape`/`interrupt` write path, incl. the non-writable-session `TERMINAL_NOT_WRITABLE` 503 handling).

## Proposed fix

Route the mailbox write edge through `submitToSession` so a gated delivery and an `escape`/`interrupt` submission take the **same** per-terminal lock. The per-terminal lock would be acquired as a **leaf inside** the per-agent serializer, so there is **no lock-cycle hazard** (the ordering is always per-agent → per-terminal, never the reverse).

## Notes

- Priority is low: the render gate already guarantees a gated body only lands on a verified-empty prompt, and `interrupt`/`escape` are explicit operator actions — so the practical corruption surface is small. This is a robustness/convergence cleanup ("one mechanism, not two"), not a live corruption bug.
- Refs #1313 / PR #1330.

## Evaluate first (absorbed from #1480, 2026-08-17)

[...]

PR #1484, opened 2026-08-17
Fix #1472: bound the mailbox escalation-toast dedupe set

### LINKED ISSUE #1472: VSCode: bound the mailbox escalation-toast `seen` Set (dedupe by mailboxId with eviction)
## What

The VSCode mailbox escalation toast (`apps/vscode/src/notifications/mailbox-escalation-toast.ts`, Spec 1313 Phase 8) dedupes "already shown" escalations with a `seen` Set that only ever grows — entries are added on show and never evicted. Over a long-lived extension host it accumulates one entry per escalated mailbox row for the lifetime of the window.

Bound it: dedupe by `mailboxId` with eviction (e.g. drop an id once its row leaves the escalated set, or cap the Set with LRU eviction).

## Why

It's an unbounded-growth leak. Negligible in practice today — escalations are rare and each entry is tiny — but it's real, it was flagged as a non-blocking note in the Phase 8 round-2 CMAP, and an unbounded Set over an all-day extension host is the kind of thing that should just be bounded rather than reasoned about.

## Notes

- Purely local to the toast module; no protocol or server change.
- Natural eviction key: the mailbox row's transition out of the escalated state (the same signal that would let a re-escalation legitimately re-notify).
- Refs #1313 (PR #1330), Follow-up section of `codev/reviews/1313-afx-send-mailbox-first-delivery.md`.

mohidmakhdoomi:
## Summary

The VSCode mailbox escalation toast deduped "already shown" escalations with a `seen` Set that only ever grew — one permanent entry per escalated mailbox row for the lifetime of the extension host. It is now bounded, evicting on the mailbox row leaving the escalated set.

Fixes #1472

## Root Cause

`apps/vscode/src/notifications/mailbox-escalation-toast.ts` — `const seen = new Set<string>()` is activation-scoped, and `seen.add(payload.mailboxId)` was its only mutation: no `delete`, no `clear`, anywhere in the module. Add-only container + extension-host lifetime = unbounded growth. Negligible per entry, but it never drains.

## Fix

Evict on **the mailbox row leaving the escalated set** — the same signal that would let a legitimate re-escalation re-notify — mirroring how `activateGateToasts` prunes a builder that leaves the blocked set:

- The toast now takes the `OverviewCache` (already in scope at the single call site in `extension.ts`) and subscribes to `onDidChange`. `OverviewData` carries no per-row mailbox ids, only the workspace-level `mailboxEscalated` flag, so `false` — nothing held in this workspace is escalated — is the finest-grained "left the escalated set" signal available client-side. At that point every id in the set has left it, so the set is dropped whole. A null/absent cache read says nothing about the escalated set and never prunes.
- `MAX_SEEN = 500` with oldest-first eviction (a `Set` iterates in insertion order) backstops the pathological window that stays escalated all day and so never observes that `false`.

No protocol or server change; the module and its call site are the whole surface.

Two notes checked against the server rather than assumed:

- `db/mailbox.ts` sets `escalated = 1` once, guarded by `escalated = 0`, and never resets it; terminal rows are pruned. So the same `mailboxId` cannot legitimately re-escalate — eviction here is purely about memory, never about suppressing a notification.
- The prune cannot fire on a stale snapshot: `OverviewCache.refresh()` is last-write-wins by sequence, and the escalation SSE event itself triggers a refresh whose request starts after Tower flagged the row, so an older in-flight `mailboxEscalated: false` response can never commit after it.

## Test Plan

- [x] Regression test added — 4 new cases in `src/__tests__/mailbox-escalation-toast.test.ts`: an id re-toasts after de-escalation; the cap evicts oldest-first while the workspace never de-escalates; no prune while still escalated; no prune on an empty cache. Verified failing without the fix (2 failed / 10 passed with the eviction neutered) and passing with it (12/12).
- [x] Build passes — `pnpm check-types` clean; porch `build` check green.
- [x] All tests pass — 72 files / 850 tests in the vscode unit suite. `pnpm lint` reports one pre-existing warning in the unrelated `src/commands/tunnel.ts`.

### DIFF
--- a/apps/vscode/src/__tests__/mailbox-escalation-toast.test.ts
+++ b/apps/vscode/src/__tests__/mailbox-escalation-toast.test.ts
@@ -18,7 +18,7 @@ vi.mock('vscode', () => ({
   },
 }));
 
-const { activateMailboxEscalationToasts } = await import('../notifications/mailbox-escalation-toast.js');
+const { activateMailboxEscalationToasts, MAX_SEEN } = await import('../notifications/mailbox-escalation-toast.js');
 
 type SSEHandler = (e: { type: string; data: string }) => void;
 
@@ -51,10 +51,35 @@ function escalationEvent(overrides: Record<string, unknown> = {}): string {
   return JSON.stringify({ type: 'mailbox-escalation', body: JSON.stringify(payload) });
 }
 
-function activate(cm: ReturnType<typeof makeConnectionManager>) {
+/**
+ * Minimal OverviewCache double: the workspace-level `mailboxEscalated` flag plus a
+ * manual change hook, so a test can drive the de-escalation prune (#1472) the same
+ * way Tower's overview refresh would.
+ */
+function makeOverviewCache(initial: { mailboxEscalated: boolean } | null = null) {
+  let data = initial;
+  const listeners: (() => void)[] = [];
+  return {
+    getData: () => data,
+    onDidChange: (fn: () => void) => {
+      listeners.push(fn);
+      return { dispose() {} };
+    },
+    /** Simulate an overview refresh committing a new escalated state. */
+    set: (next: { mailboxEscalated: boolean } | null) => {
+      data = next;
+      for (const fn of listeners) { fn(); }
+    },
+  };
+}
+
+function activate(
+  cm: ReturnType<typeof makeConnectionManager>,
+  cache: ReturnType<typeof makeOverviewCache> = makeOverviewCache(),
+) {
   const ctx = makeCtx();
-  // Structural fakes stand in for vscode.ExtensionContext / ConnectionManager.
-  activateMailboxEscalationToasts(ctx as any, cm as any);
+  // Structural fakes stand in for vscode.ExtensionContext / ConnectionManager / OverviewCache.
+  activateMailboxEscalationToasts(ctx as any, cm as any, cache as any);
   return ctx;
 }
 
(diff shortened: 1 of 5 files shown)


mohidmakhdoomi:
## Architect integration review (risk tier: Low — isolated VSCode module)

**Verdict: APPROVE** (pending maintainer review — we are not maintainers; maintainer merges.)

Verified directly against the diff:

- **Root cause is real and the fix is complete.** Pre-PR `seen` Set at `mailbox-escalation-toast.ts:30` had zero delete/clear paths; this PR adds both the semantic eviction (clear on `mailboxEscalated: false` via `OverviewCache.onDidChange`) and a `MAX_SEEN=500` oldest-first backstop for a continuously-escalated window.
- **Eviction key matches the tracking issue's guidance** (#1483 workstream D): the row leaving the escalated set is the same signal that would legitimize a re-notification. Mirrors the `gate-toast.ts` pruning precedent, so the two toast modules now age their dedupe state the same way.
- **Server-side one-shot escalation semantics** (`escalated=1` guarded by `escalated=0`, never reset) means whole-set clearing on the workspace-level flag cannot cause duplicate toasts — verified the builder's analysis against `db/mailbox.ts`.
- **Stale-snapshot race** is closed by `OverviewCache` last-write-wins sequencing; the no-data guard correctly treats an empty cache as no signal.
- One-argument signature change, one call site (`extension.ts:1522`); four targeted regression tests cover both eviction paths, the still-escalated case, and the empty-cache guard.
- Interlock with #1365 (workstream B): that issue concerns write-path locking, not escalation semantics — no coupling risk to this eviction key.

Builder CMAP: gemini/codex/claude all APPROVE. No findings to address.

Parked for maintainer approval + merge.

waleedkadous (approved):
Excellent, tight fix — thank you. Both integration reviews (codex + claude) verified every load-bearing claim against source: the server-side one-way `escalated` flag means a cleared id has no second event to dedupe against, the overview cache's last-write-wins refresh closes the stale-snapshot race, and the shape matches `gate-toast.ts` so `notifications/` is now consistent. Approving.

Two small things you might consider, neither blocking:
- The identical add-only `seen` set lives in `builder-spawn-handler.ts:11` (keyed on terminalId, no eviction) — the very module this PR's comment cites as its model. Worth a follow-up issue under #1483 rather than widening this PR.
- A half-sentence noting why this module deliberately does *not* persist its seen-set to `workspaceState` the way `gate-toast.ts` does (no replay → nothing to re-toast) would save a future reader the question.

PR #1486, opened 2026-08-17
[Air #1478] feat: carry architect:<name> through to inbox and composer attribution

### LINKED ISSUE #1478: afx send: carry architect:<name> through to inbox and composer attribution; stop truncating names in afx inbox
## What

An architect's specific name is dropped everywhere a sent message is attributed — one root cause, two visible surfaces, plus a formatting defect on one of them. *(Consolidates #1479 — same defect, second symptom.)*

**Root cause (send side):** `commands/send.ts:317` sets `from = detectCurrentBuilderId() ?? 'architect'` — the specific architect name (`currentArchitectName()` / `TOWER_ARCHITECT_NAME`) is discarded at send time, so every architect sender lands in the mailbox as the generic string `architect`.

**Surface 1 — `afx inbox`:** the FROM → TO column renders `architect → <builder>` with no way to tell *which* architect sent it. Show `architect:<name>`. Additionally, the cell is hard-sliced to a fixed 22-char column (`commands/inbox.ts:139`, `fromTo.slice(0, 22)` against `widths = [..., 22, ...]`), so a long builder/agent name is silently cut off mid-name — size the column to content (or widen / wrap) instead.

**Surface 2 — the delivered composer message:** a builder receives a generic `### [ARCHITECT INSTRUCTION | …] ###` header (`formatArchitectMessage`, `utils/message-format.ts`) with no sender name. Here a second collapse compounds the first: `formatMessageForTarget` (`servers/tower-routes.ts:1517-1526`) ignores `from` entirely on the any → builder path, so even a corrected `from` wouldn't surface. Thread the name through (e.g. `### [ARCHITECT:main INSTRUCTION | …] ###`).

## Why

Both surfaces exist to answer "who sent this?" — the inbox for the operator triaging held mail, the composer header for the builder deciding whose direction it's acting on. In a multi-architect workspace, bare `architect` is ambiguous on both, and a truncated builder name means the recipient can't be identified either. The asymmetry is telling: builder → architect messages already carry the sender (`### [BUILDER <id> MESSAGE] ###`); architect → builder attribution is just an oversight.

## Notes

- One PR: fix the `from` collapse once in `send.ts`, add an optional sender param through `formatMessageForTarget` → `formatArchitectMessage` (the any → builder branch currently discards it), and fix the inbox column width.
- `afx inbox show <id>` (`commands/inbox.ts:171`, the `From → To` kv) inherits the corrected `from` for free; the mailbox row's `from_agent` is the carrier.
- Keep the raw-mode path (`raw: true`) unattributed as today; keep the inbox list metadata-only (Spec 1313 Redaction) — this is identity + formatting, not body display.
- Refs #1313 (PR #1330). Consolidates and closes #1479. Tracked by #1483 (workstream C); the dashboard surface (#1450) inherits the corrected identity from the shared store.

### LINKED ISSUE #1479: afx send: attribute delivered composer messages to architect:<name>, not generic 'architect'
## What

When an `afx send` message is delivered into a builder's composer, it's framed with no sender name — a generic `### [ARCHITECT INSTRUCTION | …] ###` header (`formatArchitectMessage`, `utils/message-format.ts`) — so the builder can't tell *which* architect sent it. Show `architect:<name>` in the delivered framing (e.g. `### [ARCHITECT:main INSTRUCTION | …] ###`).

Two spots collapse the identity:

1. `commands/send.ts:317` sets `from = detectCurrentBuilderId() ?? 'architect'` — the specific architect name is dropped at send time.
2. `formatMessageForTarget` → `formatArchitectMessage` (`servers/tower-routes.ts:1524`) ignores `from` entirely on the any → builder path, so even a correct `from` wouldn't surface.

Both need to carry the architect name through to the rendered header.

## Why

A builder receiving an instruction should know who it's from — same as builder → architect messages already show `### [BUILDER <id> MESSAGE] ###`. In a multi-architect workspace, "ARCHITECT INSTRUCTION" with no name is ambiguous, and it matters for the builder deciding whose direction it's acting on. The asymmetry (builder sends are attributed, architect sends aren't) is just an oversight in the formatter.

## Notes

- Shares the `from = 'architect'` root cause with the `afx inbox` display issue — the send-side half can be fixed once for both surfaces.
- Thread the architect name through `formatMessageForTarget` into `formatArchitectMessage` (add an optional sender param) rather than only fixing `from`, since the any → builder branch currently discards it.
- Keep the raw-mode path (`raw: true`) unattributed as today.
- Refs #1313 (PR #1330).

mohidmakhdoomi:
## Summary

An architect's specific name was dropped everywhere a sent message is attributed. `afx send` now carries the sender as `architect:<name>`, so `afx inbox` shows *which* architect sent a held message and the builder's composer header reads `### [ARCHITECT:main INSTRUCTION | …] ###`. The inbox's FROM → TO column is sized to its content instead of being truncated at 22 chars.

Closes #1478
Closes #1479

## What Changed

One root cause, two visible surfaces, plus a formatting defect on one of them:

1. **`commands/send.ts`** — `from = detectCurrentBuilderId() ?? 'architect'` discarded the architect's name at send time. New exported `architectSenderId()` returns `architect:<name>` from `CODEV_ARCHITECT_NAME`. This is the carrier: it lands in `mailbox.from_agent`, which both surfaces read.
2. **`servers/tower-routes.ts`** — `formatMessageForTarget`'s `any → builder` branch ignored `from` entirely, a second and independent collapse: even a corrected sender could not have surfaced. It now threads `from` into the formatter.
3. **`utils/message-format.ts`** — `formatArchitectMessage` takes an optional `sender`; `architectHeaderLabel()` renders `ARCHITECT:<name>` for a *validated* architect identity, and `senderHeaderLabel()` labels any sender by role so no direction mislabels an architect as a builder.
4. **`commands/inbox.ts`** — the FROM → TO cell was `fromTo.slice(0, 22)` against a fixed 22-wide column. Cells are now rendered first and the column sized to `max(header, widest cell) + 2`.
5. **`commands/interrupt.ts` / `commands/reset.ts`** — both call the shared `architectSenderId()`, so one actor has one `from_agent` form across all three commands.

`afx inbox show <id>`'s `From → To` row inherits the corrected identity through the same mailbox row.

Verified against the built CLI, not just tests:

```
  ID                                      AGE     REASON         FROM → TO                                                                 WORKSPACE
  ────────────────────────────────────    ─────   ────────────   ───────────────────────────────────────────────────────────────────────   ─────────────
  abcdef01-2345-6789-abcd-ef0123456789    1m      busy           architect:integration-review → builder-aspir-1478-carry-architect-name    project
  ffffffff-0000-0000-0000-000000000000    10s     no-live-pty!   architect:main → spir-2                                                   project
```

```
architect->architect : ### [ARCHITECT:main MESSAGE | …] ###
builder->architect   : ### [BUILDER builder-spir-109 MESSAGE | …] ###
architect->builder   : ### [ARCHITECT:feedback INSTRUCTION | …] ###
forged name          : ### [ARCHITECT INSTRUCTION | …] ###
```

## Key Decisions

- **The carrier is the address form `architect:<name>`, not a bare name.** It is what Tower already accepts as an architect address, it stores directly as `mailbox.from_agent`, and it sits outside `looksLikeBuilderId` (which early-returns `false` for anything starting with `architect`). Confirmed by reading `tower-messages.ts`: a non-builder sender yields `lookupBuilderSpawningArchitect → undefined`, so Spec 755 sender-affinity routing and the #1094 anti-spoofing warning behave exactly as before. `sender` is never passed to `parseAddress` — only `target` is — so there is no path where a colon-bearing sender is read as an address. A test pins that `resolveTarget` still receives the sender verbatim.
- **Attribution derives from the sender's shape, not from the code branch.** `senderHeaderLabel()` yields `ARCHITECT[:<name>]` for an architect sender and `BUILDER <id>` for everything else, so the architect → architect direction can't render an architect under a `BUILDER` prefix and cron's `af-cron` framing is untouched.
- **The absent-`CODEV_ARCHITECT_NAME` fallback is bare `architect`, deliberately** (this reverses my first cut, which used `architect:main`). Tower injects the var into **every** architect terminal it starts — `main` included (`tower-instances.ts:584` uses `DEFAULT_ARCHITECT_NAME`; `tower-terminals.ts:692` re-injects `role_id || 'main'` after a shellper restart). So a missing value does not mean "the main architect", it means "not an architect terminal" (a plain shell, a script, CI). Defaulting those to `architect:main` would convert today's honest ambiguity into a specific **false** attribution — the laundering of an unverified identity that #1094 exists to prevent. Those callers keep the bare `architect` they send today; every real architect terminal gains its name. This is why `architectSenderId()` reads the env directly rather than through `currentArchitectName()`, whose default is `main`.
- **The header name is validated, not just trimmed.** `from` arrives from a `POST /api/send` body, so `architectHeaderLabel` checks it against `ARCHITECT_NAME_PATTERN` (anchored `[a-z][a-z0-9-]*`, 64-char cap) before interpolating; a crafted `architect:x] ###\n### [ARCHITECT` degrades to the bare label instead of forging framing. (`validateArchitectName` is deliberately not used — it rejects the reserved `main`, the most common real sender.)
- **Raw mode stays unattributed** and the inbox list stays metadata-only (Spec 1313 Redaction) — this is identity and formatting, not body display.
- **The column is sized to content, not widened to a bigger constant.** A fixed cap is the defect; a larger fixed cap is the same defect further out.

## Test Plan

- [x] Unit tests added — `air-1478-architect-attribution.test.ts` covers the label rules (validated names, forged-framing rejection, `architect:` with no name, builder and `af-cron` senders, raw mode, and the architect → architect header); the four route-level cases live in `tower-routes.test.ts`'s existing `POST /api/send` block, which already drives `handleRequest` against a real in-memory `global.db` with `message-format` unmocked.
- [x] `send.test.ts` — named terminal (`feedback`), explicit `main`, absent env (bare `architect`), and a malformed env value; `--all` carries the same identity. `beforeEach` clears `CODEV_ARCHITECT_NAME` so the suite is deterministic when the runner inherits a Tower-injected env.
- [x] `inbox-cli.test.ts` — long FROM → TO rendered in full, one shared width across header/separator/rows, never narrower than the header, `?` for a missing sender.
- [x] Build passes (`tsc` clean).
- [x] All tests pass — **4884 passed, 48 skipped, 0 failed**. (A first run showed 67 failures; all were an unbuilt worktree — missing `skeleton/` and `dist/` — and disappeared after `pnpm --filter @cluesmith/codev build`.)
- [x] `porch check` passes (`pr_exists`, `e2e_tests`) — but note the AIR protocol's `e2e_tests` check is `npm run test:e2e … || echo 'e2e tests skipped (not configured)'`, which cannot fail and found no root-level `test:e2e` script, so it ran nothing. I ran the e2e test that actually covers this change instead — `send-integration.e2e.test.ts` (POST /api/send → `/ws/messages` broadcast), which spawns its own Tower on port 14600: **7 passed**. I deliberately did not run the full e2e suite: its harness defaults to port 4100, the live Tower hosting this workspace's architect and builder sessions, which I can't restart without human permission.
- [x] Verified end-to-end against the built CLI — both output blocks above come from `dist/`, not from tests.

## Review round 1 — what changed

CMAP: **gemini APPROVE** (HIGH, no issues) · **codex COMMENT** (HIGH, no functional or security defects) · **claude COMMENT** (HIGH, four items). The maintainer's integration review independently found the same two defects claude did. All are fixed in `acd26ceb`:

| Finding | Resolution |
|---|---|
| architect → architect rendered `### [BUILDER architect:main MESSAGE …] ###` | `senderHeaderLabel()` — role from sender shape; architect → architect now reads `ARCHITECT:main MESSAGE`, builder senders unchanged. Tested. |
| `from` unvalidated into the composer header | Validated against `ARCHITECT_NAME_PATTERN`; degrades to bare `ARCHITECT`. Tested with a forged name. |
| `architect:main` asserted when the env names nobody (#1094 tension) | Reversed to bare `architect`, with the injection-site evidence above. Tested. |
| `interrupt.ts` / `reset.ts` still sent bare `architect` | Both call `architectSenderId()`; their "reused verbatim from `afx send`" comments now name the shared functions. |
| 263-line standalone route-test file was a duplicated harness (codex) | Route tests folded into `tower-routes.test.ts`; net −152 lines, same coverage (`9ee8ffb7`). |

### Round 2

CMAP: **gemini APPROVE** (HIGH, no issues — verified every round-1 fix and the fallback reasoning) · **codex COMMENT** (HIGH, two items, both fixed in `c3453417`) · claude pending at time of writing.

- **The framing-injection hole was wider than this body first claimed.** codex was right: it is not limited to crafted *builder* ids. A sender that merely *looks* architect-shaped (`architect:x] ###…`) fails `architectHeaderLabel`'s name validation and falls through to `senderHeaderLabel`'s builder branch, which interpolated the identity verbatim. The hole predates this PR on the builder → architect path, but `senderHeaderLabel` is the single labeling chokepoint this PR introduces, so that is where it is closed: the builder branch now checks `SAFE_SENDER_ID` (no newline, `#`, bracket or whitespace; 128-char cap — every real id passes, including `af-cron` and bare worktree names) and degrades to `BUILDER <unknown>`. Both branches validate, so the chokepoint is total. Tested, including that a forged sender cannot open a second header block.
- Stale `beforeEach` comment in `send.test.ts` (still described the absent-env fallback as `main`) corrected.

claude's round-2 review (**APPROVE**, HIGH) verified routing-inertness, the injection defense, and consumer safety at source, and raised four non-blocking items — all now addressed in `9d15deec`:

- **The `interrupt`/`reset` change was unasserted.** Both suites mock `architectSenderId`, so reverting either file to the inline `?? 'architect'` would have kept the suite green. `afx interrupt` now asserts the resulting `from` end to end; `afx reset` — whose `from` reaches a terminal port the mocked `runReset` never invokes — pins the shared-helper call instead.
- **Case sensitivity.** `architectHeaderLabel` matched a literal lowercase `architect:` while `parseAddress` is case-insensitive, so `from: 'Architect:main'` was labelled `BUILDER Architect:main`. The prefix match now follows addressing; the *name* stays strictly lowercase-validated, so `architect:Main` degrades to the bare label rather than inventing a name.
- **Docs**: the `FROM → TO` row in `agent-farm.md` now mentions the `architect:<name>` identity and content sizing, mirrored across `codev/` and `codev-skeleton/`.
- **The symmetric half of the fallback argument, for the record:** "env present" does not *prove* an architect terminal — `tower-terminals.ts:686-688` notes that a Tower-spawned process can inherit Tower's own `CODEV_ARCHITECT_NAME`. The case that could bite is a builder shell that has `cd`'d out of its worktree (so `detectCurrentBuilderId()` returns `null`): it previously sent an anonymous false `architect` and now sends a *named* false `architect:<name>`. Routing is inert either way and the value is display-only, so this is no worse in kind than today's behavior — but it is the same laundering running the other direction, and it is a real limit of an env-asserted identity rather than something the fallback choice removes.

### Test-run honesty note

One full-suite run showed a single failure; three consecutive full runs before and after are green at 4884 passed. That run's stderr was discarded, so I cannot name the test — I am reporting it as an unidentified transient rather than claiming the suite was clean throughout. It was not in any file this PR touches (those pass in every targeted run), and the repo carries a known-flaky `session-manager` suite.

## Review Notes

The judgement call worth your attention is the **absent-env fallback** (bare `architect` vs `architect:main`), reasoned from the injection sites under Key Decisions — it trades a small loss of specificity for never asserting an identity we can't source.


mohidmakhdoomi:
## Architect integration review (risk tier: Medium — shared agent-farm command/server code)

**Verdict: REQUEST_CHANGES** — the root-cause fix is correct and verified-safe, but two findings should land in this PR. Both verified by the architect against the code, not just the reviewer's claims.

### What checks out (verified)
- `architect:<name>` as `from` is routing-inert: `looksLikeBuilderId()` early-returns false on `architect*` (`tower-messages.ts:51`), the affinity lookup misses, neither the Spec 755 branch nor the #1094 spoofing branch is entered, and `from` never passes through `parseAddress`.
- No production code parses the `[ARCHITECT INSTRUCTION` header; header change is safe.
- `logger.row` never truncates — the `.slice(0,22)` was the whole defect; content-sizing is the right fix.
- Test approach (driving the real `/api/send` registry hold path and asserting `from_agent` + `formatted_message` on the row) is the right shape.

### Changes requested
1. **Architect→architect sends now mislabel harder.** `formatMessageForTarget`'s `isArchitectTarget && from` branch routes to `formatBuilderMessage`, whose header is a hardcoded `### [BUILDER ${builderId} MESSAGE…]` — so a named architect renders as `BUILDER architect:feedback`. Cross-architect messaging is documented and used. The PR's own principle (attribution derives from sender shape) fixes it in ~2 lines: when `from` is an `architect:<name>` identity, use `architectHeaderLabel`; keep `BUILDER <id>` for builder senders. Add the missing architect→architect test case.
2. **`from_agent` goes bimodal for the same actor.** `interrupt.ts:40` and `reset.ts:56` still write bare `architect` while `send.ts` now writes `architect:main` — one architect, two identities in the same inbox table, and both files' "reused verbatim from send.ts" comments are now false. `architectSenderId()` is already exported; call it from both. (Original scoping to send.ts was reasonable and flagged transparently — but defining the helper once and using it at the existing consumers is what keeps the column meaningful.)

### Consider / rebut (non-blocking)
- When `CODEV_ARCHITECT_NAME` is absent, `architectSenderId()` manufactures `architect:main` — in tension with `whoami`'s documented "no implicit fallback to main" stance (#1094). Every Tower-started architect terminal has the var injected, so the fallback fires only where the claim is least warranted. Falling back to bare `architect` would keep attribution honest at zero routing cost. Judgment call — decide deliberately and say why.
- Follow-up candidates (not this PR): content-size `REASON`/`WORKSPACE` columns on the same principle; give the VS Code extension's send an architect `from` (currently renders `? → builder`); a line in the skeleton's `afx send` docs + CHANGELOG for the new user-visible header.

Parked for maintainer approval + merge after changes; we are not maintainers.

waleedkadous:

mohidmakhdoomi:
## Rebased onto current `main` — conflicts resolved, architect integration check

This branch had gone `CONFLICTING` (770 commits behind). It is now **`MERGEABLE`** at `291987e7f`, via a **merge** of `origin/main` (`d86ea25d1`), not a rebase — per this repo's merge-never-squash convention.

### The one substantive re-shape

Four files conflicted, but only one needed real thought: **#1574's self-attesting frames landed on the same wrapper this PR edits.** The two changes turned out to be orthogonal rather than redundant —

- **#1574** names the **recipient**: `recipient(toAgent)` in the header.
- **#1478** (this PR) names the **sender**: `architect:<name>` instead of a generic `architect`.

Neither carried the other, so both are still needed. Merged header:

```
### [ARCHITECT:main INSTRUCTION → spir-9 | 2026-09-03T…] ###
```

Post-#1574 the production any→builder path routes through `formatArchitectToBuilderMessage`, which on `main` takes no `sender` and passes none down (`message-format.ts:131-135`) — so the sender name **stopped one hop short of the only surface this issue is about**. That hop is now threaded and pinned by a unit test.

### Architect verification (independent of the builder's report)

- Root cause still live on `main` before this PR: `commands/send.ts:328` is `detectCurrentBuilderId() ?? 'architect'`, no `currentArchitectName()`. Confirmed.
- `formatMessageForTarget`'s **signature is unchanged**, so all six call sites (`tower-routes.ts:1769, 1966, 2011, 2042, 2092` + `:1550` def) already passed `from` — no call-site fallout.
- **No collateral revert.** I diffed every deletion this branch makes against `main`: each one is exactly the line its replacement supersedes, and every rewritten header keeps `recipient(toAgent)`. #1574's and #1494's work is preserved.
- Source delta is 179 insertions / 27 deletions across 6 files; test delta adds a new 156-line attribution suite plus updates to five existing ones.

Test fallout from the merge was mechanical: `main` renamed `reset` → `refresh`, and two header regexes predated the `→ recipient` segment (they now assert both halves).

### Status

Parked for maintainer review — **we are not maintainers of this repo, so we are not merging this.** Please approve and merge when it suits you. The issue stays open and the worktree stays intact until you do.

waleedkadous (changes requested):
Thanks for the rebase — the core is in good shape: both lanes verified the attribution threads consistently through send, interrupt, refresh and persistence, `formatMessageForTarget` stays the single funnel, and the #1573/#1584 echo-needle derives from the formatted first line at runtime so delivery verification is untouched. I'd like to see three changes before merge, all small:

1. **Architect-shaped senders that fail validation fall through to `BUILDER …`.** `senderHeaderLabel()` (`message-format.ts:72-77`): `architect:Main`, `architect:`, and the default auto-numbered `architect-3` all end up as `BUILDER architect:Main` / `BUILDER architect-3`, contradicting the documented "invalid architect name degrades to the bare ARCHITECT label". Detect the architect prefix independently and return `architectHeaderLabel(sender)` for every architect-shaped sender; `looksLikeBuilderId` already excludes `architect*` for exactly this reason. Tests through `senderHeaderLabel()` and `formatBuilderMessage()` for the malformed cases.
2. **Bound sender metadata at the API boundary.** `/api/send` persists `body.from` with no length/identity validation (`tower-routes.ts:~1901`), and the new content-sized inbox column (`inbox.ts:140-151`) pads every row to the widest sender — a local caller can store a ~1 MiB sender and make `afx inbox` allocate accordingly. Validate `from` against the same safe-identity contract (or cap its length) while keeping full display for legitimate ids.
3. **arch.md §sender attribution (~1862-1868)** still describes the pre-PR two-class header mapping, and its "do not build on this header as proof of identity" caveat covers only `[USER via VS Code]` — `ARCHITECT:<name>` is equally caller-controlled and needs the same explicit non-provenance clause (the #1094 laundering concern).

Worth a follow-up rather than this PR: give the VS Code extension's send an architect `from` (it renders `? → builder` today), and thread `senderHeaderLabel` through the any→builder path so builder→builder frames stop claiming `ARCHITECT` (overlaps #1576). Also noting `message-format.ts` now depends on the built sdk — a new build-order edge on a file the Tower server loads; fine, just worth knowing.

mohidmakhdoomi:
Thanks — all three were real, and the second one was partly created by this PR. Pushed in 871161e88.

## 1. Architect-shaped senders falling through to the BUILDER branch

Confirmed, and worse than an edge case: `architect-3` is the **default auto-numbered architect name**, so this was the common path.

Root cause was routing on `architectHeaderLabel`'s *return value*. Its fallback `'ARCHITECT'` answers two different questions with one string — "not an architect" and "an architect whose name is unshowable" — and the second fell through to the builder branch. The prefix is now detected independently of whether the name validates, mirroring `looksLikeBuilderId` as you suggested:

```ts
export function senderHeaderLabel(sender: string): string {
  const bare = sender.toLowerCase();
  if (bare === 'arch' || bare.startsWith('architect')) return architectHeaderLabel(sender);
  return SAFE_SENDER_ID.test(sender) ? `BUILDER ${sender}` : 'BUILDER <unknown>';
}
```

Tested through **both** `senderHeaderLabel()` and `formatBuilderMessage()` for `architect:Main`, `architect:` and `architect-3` — the second test asserts the rendered header, since a label is only correct if it reaches the frame a recipient reads.

One existing assertion moved with this: an architect-shaped forgery (`architect:x] ###…`) now degrades to the bare `ARCHITECT` label rather than `BUILDER <unknown>`. Nothing of the payload is interpolated either way, but the old test asserted `not.toContain('### [ARCHITECT ')` — which the legitimate degraded label now contains. It counts header frames instead:

```ts
const out = formatBuilderMessage(forged, 'spir-9', 'hi');
expect(out.match(/^### \[/gm)).toHaveLength(1);
expect(out).not.toContain(forged);
```

## 2. Unbounded sender metadata

Confirmed on both halves, and the amplification is on this PR — the content-sized column is mine.

`isSafeSenderId` and `MAX_SENDER_ID_LENGTH` (128) are now exported from `message-format.ts`, so the boundary and the renderer enforce **one** contract rather than two that can drift. `POST /api/send` refuses a non-conforming `from` with a 400, following the `MAX_MESSAGE_BYTES` precedent immediately above it. Rejected rather than dropped or capped on purpose: a message delivered under a silently-discarded sender is exactly the unattributed frame this PR exists to remove.

`inbox.ts` gets a display ceiling of `MAX_SENDER_ID_LENGTH * 2 + ' → '.length` as defence in depth for rows persisted before the route check existed. Explicitly **not** the 22-char truncation returning — there is a test asserting a pair of *maximum-length* legitimate ids renders whole with no ellipsis, alongside one asserting a 200,000-char stored sender is capped and does not set the padding for every other row.

The only real callers are the CLI (`architectSenderId()` / `detectCurrentBuilderId()`) and the extension's `vscode-user`; all conform.

## 3. arch.md sender attribution

Rewritten to the three classes actually shipped, including that the architect → architect label follows the sender's *shape* rather than a hardcoded role.

On the caveat — you're right, and I've made it explicit for both classes: shape validation bounds what can be **rendered**, it does not prove **authenticity**. `[ARCHITECT:<name>]` no more proves that architect sent it than `[USER via VS Code]` proves a human clicked. The section now names the #1094 laundering concern directly and says a recipient must not treat a header as authorization for what the body asks.

No skeleton mirror — `codev-skeleton/resources` has no `arch.md`.

## Follow-ups acknowledged, not in this PR

Both noted and deliberately left out of scope: giving the VS Code extension's send an architect `from` (it renders `? → builder` today), and threading `senderHeaderLabel` through the any → builder path so builder → builder frames stop claiming `ARCHITECT` (overlaps #1576).

## Verification

`pnpm test` — 5497 passed, 0 failed (10 new tests). `tsc --noEmit` clean. Nothing under `apps/` touched, so its separate suite wasn't needed.

waleedkadous (dismissed):
Re-verified: `senderHeaderLabel` detects the architect prefix independently of name validity (mirroring `looksLikeBuilderId`), so `architect-3` — the default auto-numbered name, i.e. the common path — no longer falls through to `BUILDER`; `from` is bounded to `MAX_SENDER_ID_LENGTH` (128) through the same safe-id regex with tests either side of the limit, which also caps the inbox column; and the arch.md non-provenance caveat now covers `[ARCHITECT:<name>]` explicitly. CI 7/7 on 9275a1036. Approving and merging.

Issue #1474, opened 2026-08-17
Render gate: tighten the loose agy prompt marker (AGY_MARKER = /^> /)


mohidmakhdoomi:
## What

The agy (Gemini CLI) render-gate profile keys its prompt detection on `AGY_MARKER = /^> /` — any line beginning with "> " is treated as the composer prompt. That's a loose match: agy's profile was derived empirically (color-keyed placeholder palette) and this marker is the weakest link, since "> " is common in ordinary output (quoted text, diffs, shell transcripts, agent prose).

Tighten it so the marker identifies the *actual* agy composer prompt rather than any "> "-prefixed line — e.g. anchor on additional profile signals (cursor row, placeholder-fg palette, region boundary) already used elsewhere in the agy classifier.

## Why

A too-loose prompt marker is a false-clean risk: if a "> " line in agy's output is mistaken for an empty composer, the gate could classify CLEAN and deliver onto a busy screen — the exact corruption the mailbox-first design rules out. It hasn't bitten in practice, but it's the least-grounded part of the agy profile and worth hardening while the fixture-capture tooling is fresh.

## Notes

- agy classification is the least-covered of the three profiles (Gemini skips non-blockingly when unauthenticated), so a captured real-agy fixture across states (idle / draft / trust-dialog / quoted-output) should back the tightening.
- Distinct from #1365 (serializer convergence) — that bullet in the review pairs them, but this is specifically about the agy prompt marker, not the interrupt-vs-mailbox lock.
- Refs #1313 (PR #1330), Technical Debt / Follow-up sections of `codev/reviews/1313-afx-send-mailbox-first-delivery.md`.

Issue #1473, opened 2026-08-17
Render gate: fuller close of the gate→write input race (R7 staleness) and input-echo-lag residual


mohidmakhdoomi:
## What

The render gate samples a ring change-token, classifies the screen, then re-checks the token before writing — so a change in the gate→write window forces a HOLD rather than a write onto a stale-clean prompt. That guard closes the **output**-side staleness window. Two related **input**-side residuals remain open:

1. **Human keystroke between snapshot and write (the R7 staleness case).** A user can type into the composer in the same gate→write window; the change-token catches most cases, but the guard should fully cover a keystroke that lands after classify but before (or during) the paced write.
2. **Input-echo lag.** A just-typed character may not yet be echoed into the ring at snapshot time, so the gate can read "empty" a beat before the echo lands.

Tighten the gate so both are covered — i.e. the "verified-empty prompt" guarantee holds against in-flight human input, not just in-flight agent output.

## Why

The whole point of the gate is that a body only ever lands on a genuinely empty prompt. The output-side token guard already enforces that against agent repaints; these two input-side windows are the remaining ways a human keystroke can race the write. They're narrow, but they're the same corruption class the spec exists to eliminate, so closing them fully is worth doing.

## Notes

- Fail-safe today: the residual biases toward HOLD, never toward force-injecting onto a draft — this is hardening, not a live corruption bug.
- Likely lives around the classify → re-check → write sequence in `render-gate.ts` / `mailbox-delivery.ts`; consider extending the change-token to also reflect pending input / echo state, or an explicit post-write echo settle.
- Refs #1313 (PR #1330), Follow-up section of `codev/reviews/1313-afx-send-mailbox-first-delivery.md` (the "gate→write input race / R7 staleness guard / input-echo-lag residual" item).

mohidmakhdoomi:
## Pre-spawn verification (architect, 2026-09-04 @ `03bc5213e`)

The body was written 2026-08-17. **The premise still holds and the residual is still open** — but a lot landed underneath it since (#1573 settle-before-write + echo verification, #1584 point-of-no-return, #1365 the per-terminal write lock, #1482 the `detail` sub-code), so the *shape* of what is left is narrower and better-defined than the body describes. Verified refs at HEAD:

| Claim | Body said | Actually at HEAD | Verdict |
|---|---|---|---|
[...]

PR #1487, opened 2026-08-17
Fix #1471: replace the render-gate perf wall-clock bound with a deterministic op-count check

### LINKED ISSUE #1471: Render gate: replace the perf wall-clock assertion with a deterministic op-count check
## What

The render-gate perf test (`render-gate.test.ts`) asserts classification stays under a wall-clock budget (best-of-5 min). Because it's timing-based, it flakes on loaded CI runners — the whole-ring/seed-cap render measured 125–142ms against a 75ms local ceiling. It's currently mitigated (not fixed) with a **CI-aware bound** (`process.env.CI ? 800 : 250` ms), which keeps a tight local steady-state signal while letting CI assert only a catastrophic-regression ceiling.

Replace the wall-clock assertion with a **deterministic op-count check** — count the work the classifier actually does (cells scanned / ring bytes rendered / viewport size) and assert on that, so the test measures algorithmic cost rather than machine speed.

## Why

A wall-clock budget is inherently non-deterministic: it either flakes under load or is loosened until it no longer catches the regression it exists for. The gate's real cost guarantee is "O(viewport), not O(ring size)" — an op-count assertion pins that invariant directly and can't flake on a busy runner.

## Notes

- The CI-aware bound stays as-is until this lands; it's a mitigation, not the intended end state.
- The relevant cost property post round-2 is the persistent bounded `SessionScreen` mirror (classify is O(viewport)), plus the per-`ringToken` verdict memo + cost-aware backstop backoff — an op-count check should exercise the mirror path, not the retired whole-ring path.
- Refs #1313 (PR #1330), Follow-up / Flaky Tests sections of `codev/reviews/1313-afx-send-mailbox-first-delivery.md`.

mohidmakhdoomi:
## Summary

The render-gate perf test asserted a best-of-5 **wall-clock** measurement against a CI-aware budget, so it flaked on loaded runners. This replaces it with a deterministic **op-count** check that asserts the classifier's algorithmic cost — one classify is O(viewport), not O(ring size) — in integers that cannot flake.

Test-only change; no production file is touched.

Fixes #1471

## Root Cause

`render-gate.test.ts` measured `performance.now()` deltas around `classifyScreen` over a 4 MB replay and asserted `best < (process.env.CI ? 800 : 250)` ms. Two problems:

1. **It measures the machine, not the algorithm.** Identical code that passes comfortably on an idle box measured **391.7ms** when pinned to one contended core (`taskset -c 0` + busy loops) — reproduced locally, `AssertionError: expected 391.73237999999947 to be less than 250`. Best-of-5-min narrows the spread but cannot remove it: a fully contended core has no lucky run. So the bound either flakes or gets loosened until it no longer catches the regression it exists for (its history: 75ms → 250 → a CI-aware 800).
2. **It measures the wrong path.** It timed `classifyScreen`, the transient whole-ring entry that parses the entire replay into a throwaway `Terminal` per call — genuinely O(ring size). Production (round 2) classifies the persistent bounded `SessionScreen` mirror via `classifyBuffer`, which only *reads* an already-parsed viewport.

## Fix

New suite — *"deterministic op count: one classify is O(viewport), not O(ring size) (#1471)"* — exercising the **mirror** path (`SessionScreen.feed` chunked as `PtySession.onPtyData` does → `read()` → `classifyBuffer`).

`classifyBuffer` takes the terminal as a parameter and only reads it, so the test hands it a `Proxy` facade counting `getLine` / `getCell` calls and any bytes written. The real classifier runs; nothing is stubbed and no production instrumentation is added.

| Test | Asserts |
|---|---|
| 4 MB vs ~200 B of history, same repainted final screen | op counts are **byte-identical** — 20000× the history, identical work |
| one classify | `lineReads ≤ 2·rows`, `cellReads ≤ cols·rows`, `bytesParsed === 0` |
| three classifies of a static screen | flat per-classify cost (the backstop re-checks on a timer) |
| negative control | the retired whole-ring path re-parses ~4 MB **per classify** — so the counter demonstrably discriminates the two cost models |

The existing 4 MB whole-ring test keeps its correctness half (renders whole, no slice/size cap, classifies its busy tail); only the timing loop and the CI-aware budget are removed.

**Verified against simulated regressions** (both reverted afterward):

| Simulated regression in `render-gate.ts` | Result |
|---|---|
| `screenLines` scans from history start instead of `viewportY` | 3 op-count tests fail |
| pure cost regression — walk all history, verdict unchanged | 2 fail: `lineReads` 1131 vs the ≤130 viewport bound, 1131 vs 66 across history sizes |

The second matters most: the verdict stays correct, so *only* the op count catches it — exactly what the wall-clock bound was proxying for.

## Test Plan

- [x] Regression test added (fails under a simulated O(history) regression, passes on `main`'s classifier)
- [x] Build passes (`pnpm --filter @cluesmith/codev build`)
- [x] All tests pass (4860 passed / 48 skipped / 0 failed)
- [x] Deterministic under load: all 46 tests in the file pass pinned to the contended core that broke the old assertion
- [x] Test file typechecks clean (the package build excludes `__tests__`, so this was checked under a temporary tsconfig)

### DIFF
--- a/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
+++ b/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1471
+title: render-gate-replace-the-perf-w
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-08-17T23:26:16.744Z'
+updated_at: '2026-08-17T23:26:16.745Z'
--- a/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
+++ b/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1471
 title: render-gate-replace-the-perf-w
 protocol: bugfix
-phase: pr
+phase: verified
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -13,5 +13,5 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-17T23:26:16.744Z'
-updated_at: '2026-08-17T23:48:29.209Z'
+updated_at: '2026-08-17T23:48:33.936Z'
 pr_ready_for_human: false
--- a/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
+++ b/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1471
 title: render-gate-replace-the-perf-w
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-17T23:26:16.744Z'
-updated_at: '2026-08-17T23:26:16.745Z'
+updated_at: '2026-08-17T23:30:06.987Z'
--- a/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
+++ b/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1471
 title: render-gate-replace-the-perf-w
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-17T23:26:16.744Z'
-updated_at: '2026-08-17T23:30:06.987Z'
+updated_at: '2026-08-17T23:36:24.616Z'
--- a/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
+++ b/codev/projects/bugfix-1471-render-gate-replace-the-perf-w/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-08-17T23:45:28.418Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-08-17T23:26:16.744Z'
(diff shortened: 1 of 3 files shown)


mohidmakhdoomi:
## Architect integration review (risk tier: Low — test-only, one file)

**Verdict: APPROVE** (parked for maintainer review + merge; we are not maintainers).

- **Root cause is deeper than the issue framed, and the fix is aimed correctly.** The flaky assertion didn't just measure wall-clock — it measured the *retired* whole-ring path (`classifyScreen`, genuinely O(ring size)) rather than production's persistent-mirror path (`classifyBuffer`, O(viewport)). The new op-count suite pins the invariant the wall-clock bound was proxying for, on the path production actually runs — exactly what #1483's guidance asked for.
- **The counting facade measures the real classifier.** A `Proxy` over the mirror's live terminal counts `getLine`/`getCell`/bytes; nothing is stubbed. Integer comparisons cannot flake on a loaded runner.
- **Regression evidence is the strongest part:** an injected pure-cost regression (verdict unchanged, cost O(history)) is caught *only* by the op counts (1131 line reads vs the ≤130 geometric bound) — demonstrating the new test detects what the old one existed for but couldn't reliably catch. Claude's CMAP lane independently re-ran the experiment and reproduced the numbers.
- The 4 MB test keeps its correctness half; only the timing loop and CI-aware budget are removed. Negative control shows the counter distinguishes the two cost models.
- Builder CMAP: 3× APPROVE (HIGH). 7/7 GitHub CI checks green, including the shared runners the old assertion flaked on.

Deliberately scoped tight so the gate-hardening issues (#1473, #1474) touching this file later rebase cleanly.

waleedkadous (approved):
Really nice piece of test engineering — the counting-facade technique is the right answer to wall-clock flakiness, and both reviewers verified the bounds are exact (`lineReads ≤ 2·rows`, `cellReads ≤ cols·rows` as the true worst case) with a negative control that keeps the point. Approving.

One small loose end you could fold in before merge, or I'll pick it up: issue #1471 names `codev/reviews/1313-afx-send-mailbox-first-delivery.md`, whose lines ~341 and ~348 still describe the CI-aware wall-clock bound as live and the op-count check as pending. That file's own convention (line 347) is a strike-through + DONE pointer — two lines pointing at this PR.

Worth knowing rather than changing: the branch is well behind main, but the intervening render-gate work doesn't alter `classifyBuffer`'s measured behavior, and the facade's coupling to xterm's descriptor shape fails loudly rather than silently on an upgrade. Claude's lane also suggested the same technique could retire several sibling wall-clock assertions (tunnel-e2e, shellper-client, session-manager, spec-1365) — a good follow-up issue if you're inclined.

Issue #1361, opened 2026-08-06
Render gate: adopt/reconnect mirror seed capped at 1 MiB → idle adopted alt-screen agent can hold mail until next repaint


mohidmakhdoomi:
## Summary

The render-gate persistent mirror (`SessionScreen`, Spec 1313 round-2) is seeded on **adopt/reconnect** from a **1 MiB-capped** replay tail. For a long-lived alt-screen agent whose current frame started **> 1 MiB back** in the byte stream, the mirror is **born torn** (missing the alt-screen enter / composer marker / bounding rule) and classifies BUSY. An **unattended, idle** adopted agent that then emits **zero further output** has no repaint to heal the frame, so its mailbox can stay held until its next repaint.

This is a **pre-existing residual, not a round-2 regression** — pre-round-2 the gate classified `ringBuffer.getAll().join('\n')`, and that ring was seeded from the *same* capped replay, so the adopt-path tear existed identically before. Round-2 strictly *fixes* the live-session tear (the common outage) and leaves this adopt edge where `main` already had it.

**It is fail-safe:** torn → HOLD, never misdeliver. It **self-heals** on any post-adopt output *or* any viewer attach (the WebSocket connect resize nudge triggers a repaint). Mail is **delayed, not lost.**

Reachability is non-trivial in this workflow because a **Tower restart is routine** (every `pnpm -w run local-install` restarts Tower → every active session is adopted via the reconcile path).

## Verified code locations (spir-1313 branch / PR #1330)

- `packages/codev/src/agent-farm/servers/tower-terminals.ts:775` (reconcile) and `:1034` (reconnect) — both call `capRingSeed(await client.waitForReplay(), …)` before `attachShellper`.
- `capRingSeed` (`tower-terminals.ts:40-44`) — returns the last `RING_SEED_MAX_BYTES` (`= 1024*1024`, 1 MiB) suffix when replay exceeds it.
- `packages/codev/src/terminal/pty-session.ts:227-231` (`attachShellper`) — seeds the mirror (`feedGateScreen(replay)`) with the **identical capped bytes** the ring gets (`ringBuffer.pushData(replay)`).
- Shellper retains up to `REPLAY_PAYLOAD_MAX = 8*1024*1024` (8 MiB) of replay (`shellper-protocol.ts:40`), so the uncapped replay is available.

## Fix options

[...]

Issue #1488, opened 2026-08-17
Protocols: e2e_tests check can never fail — '|| echo skipped' swallows real e2e failures (AIR, SPIR, ASPIR)


mohidmakhdoomi:
## What

The `e2e_tests` check in three protocol definitions is a no-op that cannot fail:

```
npm run test:e2e 2>&1 || echo 'e2e tests skipped (not configured)'
```

- `codev/protocols/air/protocol.json:72` (and skeleton twin)
- `codev/protocols/spir/protocol.json:142` (and skeleton twin)
- `codev/protocols/aspir/protocol.json:139` (and skeleton twin)

The `|| echo` exists to tolerate repos with no `test:e2e` script, but it swallows **every** non-zero exit — including a real, red e2e suite in a repo that has one. The check reports green in ~0.1s having run nothing (no root `test:e2e` script exists in this repo), or having ignored genuine failures elsewhere.

Discovered during AIR #1478: the builder noticed its porch `e2e_tests` check passed instantly without executing anything, and ran the relevant e2e suite (`send-integration.e2e.test.ts`) manually instead.

## Why it matters

A check that cannot fail is worse than no check: it stamps a green `e2e_tests` line into every porch run, which reads as verified coverage. Spec 550 already shipped the sanctioned mechanism for repos without e2e — `porch.checks` overrides in `.codev/config.json` (`"e2e_tests": { "skip": true }`), which prints an explicit skip rather than a counterfeit pass.

## Fix shape

Distinguish "script missing" from "script failed": run `test:e2e` if it exists and propagate its exit code; emit an explicit skip only when the script is absent (e.g. probe `npm run --if-present`'s semantics or check the script's existence first). Apply to all three protocol.json files **and their codev-skeleton twins**.

Found while working #1483's program; not part of that program's scope.

PR #1491, opened 2026-08-18
[Air #1474] Render gate: anchor the agy prompt marker to the cursor row and marker palette

### LINKED ISSUE #1474: Render gate: tighten the loose agy prompt marker (AGY_MARKER = /^> /)

mohidmakhdoomi:
Implements #1474.

## What was actually wrong

`AGY_MARKER = /^> /` (`gate-profiles.ts:72`) treated **any** line starting with `> ` as agy's
composer prompt, and `findMarkerRow` is last-match-wins. The issue called this "the weakest link"
in an empirically-derived profile. Measured against a real, authenticated **agy 1.1.13** under a
PTY at 110×32, it is worse than a theoretical looseness — non-composer `> ` rows are *routine*:

| real screen | last `> ` row the old marker picked | that row's fg | cursor row | the actual composer |
|---|---|---|---|---|
| idle (accept-edits) | 11 | palette 12 | 11 | 11 ✅ |
| idle (no-hint mode) | **none** — bare `>` never matched `/^> /` | palette 12 | 11 | 11 ❌ |
| draft | 24 | palette 12 | 24 | 24 ✅ |
| slash menu (`/`) | 13 — **the menu's selection cursor** | palette 12 | 11 | 11 ❌ |
| trust dialog | 8 — **the selected option** | palette 12 | 12 (off-row) | none ❌ |
| settled after an answer | 20 | palette 12 | 20 | 20 ✅ |
| torn mid-repaint frame | 10 — **the transcript echo of the sent turn** | palette **4** | varies | absent ❌ |

Three findings drove the design:

1. **agy echoes every submitted turn into the transcript as a `> ` line** (SGR 34;1 → palette 4).
   Non-composer `> ` rows accumulate one per conversation turn.
2. **The slash menu's selection cursor is also `> `, also palette-12, and renders BELOW the
   composer** — so it won the marker scan outright. A color anchor alone does *not* separate
   these; the cursor row does.
3. **In agy's no-hint mode the composer is a bare `>`**, which right-trims to `">"` and never
   matched `/^> /` — so the gate held **every** message to an agy in that mode forever. A
   pre-existing false-HOLD, fixed here, since the issue's ask is that the marker identify the
   *actual* composer.

Markdown blockquotes turned out to render as `│`, not `> ` — so the issue's "quoted output" risk
arrives via the turn echo rather than via blockquotes.

**Honest scope of the defect.** I could not reproduce an actual false-CLEAN from the captures: on
every mis-bounded frame the wrong region still happened to contain counted text, so the verdict
landed `busy` anyway. What is demonstrated is that the classifier **bounds the wrong region on
real screens**, and that getting `busy` out of a wrong region is luck rather than a guarantee. The
corruption risk is latent, not observed — stated plainly rather than overclaimed.

## The change

Two optional, per-app `GateProfile` anchors — profile *data*, in keeping with the spike's
constraint 9 — set only on agy, both measured:

- `markerRequiresCursorRow: true` — the marker row must hold the buffer cursor. Only the live
  input row does; menu rows, dialog options and transcript echoes never do.
- `markerFgPalette: 12` — the marker glyph's own color, which separates the composer from the
  palette-4 turn echo.

Plus the marker separator relaxed to `/^>(\s|$)/` for the bare-`>` mode. claude/codex set neither
anchor and are byte-for-byte unaffected (pinned by a test).

Every failure direction is toward **HOLD**: a row that fails an anchor is simply not a marker, so
drift yields `no-composer-marker` → held and re-checked, never a false clean. Sustained holds
already escalate through the mailbox liveness telemetry (`recordStreak`).

## Verdict changes on the real captures

| fixture | before | after |
|---|---|---|
| `agy-idle.clean` | clean | clean |
| `agy-draft.busy` | busy / user-text | busy / user-text |
| `agy-menu.busy` | busy / **no-region-end** (marker was the menu item — right verdict, wrong reason) | busy / **user-text** (right region: the `/` typed in the composer) |
| `agy-trust.busy` | busy / **no-region-end** (incidental — no rule under the option) | busy / **no-composer-marker** (honest: there is no composer) |
| `agy-torn-echo.busy` | busy / no-region-end (marker was the palette-4 echo) | busy / **no-composer-marker** |
| `agy-turn-echo.clean` | clean | clean |
| `agy-baremarker.clean` | **busy / no-composer-marker (false HOLD)** | **clean** |

## Cost of the tightening, measured

Sweeping every byte-prefix of a real stream showed 70 CLEAN frames before vs 9 after — alarming
until you notice byte-prefix sampling cuts *mid-repaint*, which production never does. Sampling
the way production actually classifies (wall-clock, against a live agy driven through boot → idle
→ streaming → settled, classifying the mirror every 200 ms) the two profiles are **identical**:

```
idle       samples=  30   OLD clean=100%   NEW clean=100%
streaming  samples= 205   OLD clean= 98%   NEW clean= 98%
settled    samples=  40   OLD clean=100%   NEW clean=100%
```

No measurable false-HOLD cost, and the bare-`>` mode goes from permanently held to deliverable.

## Fixtures

The issue asked for captured real-agy fixtures across states, and agy **is** authenticated in this
environment, so the three synthesized Phase-3 fixtures are replaced by **seven real captures**
(idle / bare-marker / draft / menu / trust / turn-echo / torn frame). agy's banner embeds the
account email and session cwd; both are replaced with **same-length** placeholders so the rendered
screen stays byte-for-byte equivalent — no attribute is retouched. Each file is 1.7–8.7 KB, so
they are committed plain rather than gzipped (gzip is used for the multi-hundred-KB claude
replays). The torn-frame fixture is real bytes cut mid-repaint, which is the tear shape #1361
documents for the adopt seed.

## Tests

55 pass in `render-gate.test.ts`. New coverage: per-fixture verdicts for all seven states; the
*reason* for each (a mis-bounded region that returns `busy` is not the same as a correctly-bounded
one); synthetic branch tests for each anchor independently, including the palette-12 dialog option
with the cursor **on** it — the one shape the anchors cannot reject, which pins that the occupancy
count still catches it; and a guard that claude/codex are unaffected by cursor position.

`porch check 1474`: build ✓ (13.1s), tests ✓ (28.7s).

## Notes for the reviewer

- The cursor-row anchor assumes `buffer.cursorY` is viewport-relative, the same convention
  `isGhostCursorCell` already relies on. True whenever `viewportY === baseY` (no manual scrollback),
  which holds on both gate paths.
- The one shape neither anchor rejects is a dialog that parks the cursor on a palette-12 `> ` option.
  The trust dialog does not (measured), and the occupancy count plus the region-end guard both still
  catch it — but it is the seam to watch if agy adds dialogs.


mohidmakhdoomi:
## Test status, stated precisely

`porch check 1474`: build ✓ (13.5s), unit tests ✓ (28.6s) — 55 pass in `render-gate.test.ts`.

**The `e2e_tests` gate check is a no-op here.** AIR defines it as `npm run test:e2e 2>&1 || echo 'e2e tests skipped (not configured)'` and the repo root has no `test:e2e` script, so it reports ✓ in 0.1s having executed nothing. It is marked `optional` in the protocol, so that is intended — but it is not evidence, so I ran the real suite:

```
pnpm --filter @cluesmith/codev test:e2e
Test Files  1 failed | 19 passed | 5 skipped (25)
     Tests  3 failed | 171 passed | 21 skipped (195)
```

All 3 failures are in `tower-api.e2e.test.ts` — `POST /api/terminals` returning **500** where 201 is expected.

**They are pre-existing and unrelated to this PR.** Verified rather than asserted: I reverted both changed source files to the branch base (141b4935), rebuilt, and re-ran that file — **identical 3 failures**. Files restored afterwards (all committed, so lossless; working tree verified clean against HEAD, unit tests re-run green).

I did not skip or annotate them as flaky, because they are neither mine nor intermittent — they reproduce consistently. Flagging for a maintainer instead. Plausibly environmental: this machine is running four-plus concurrent builder sessions plus Tower, and all three failures are PTY-spawn-through-the-API.

mohidmakhdoomi:
## Architect integration review — 3-way CMAP (risk tier: High — render-gate classifier, false-CLEAN direction)

**Verdicts: gemini APPROVE · codex APPROVE · claude COMMENT ("merge it, with follow-ups") — all HIGH confidence, zero blocking correctness issues.** Architect read concurs: the two positive-evidence anchors (`markerRequiresCursorRow`, `markerFgPalette`) fit the existing per-app-profile seam, every failure direction lands on `no-composer-marker` → HOLD, and claude/codex profiles are untouched. The real-capture measurement table is the load-bearing artifact, and the PR's refusal to overclaim (mis-bounding demonstrated, false-CLEAN latent; no-hint 'held forever' rests on live observation) is noted and appreciated.

Independently verified across the reviews: viewport-relative `cursorY` is sound on the production gate (`viewportY === baseY` for `SessionScreen`); the tightening's failure direction is observable (`no-composer-marker` feeds `isClassifierStuck` → escalation at streak 10); fixture sanitization is complete (no emails/paths survive); the bare-`>` relaxation is gated behind both anchors so it cannot loosen anything — and it fixes a real pre-existing total delivery outage in agy's no-hint mode.

### Requested before the gate (one item)
1. **Add an agy case to the production-mirror-path suite** (`render-gate.test.ts:364`). These are the first anchors that depend on **cursor state**, the one dimension where the transient path and the `SessionScreen` mirror path could conceivably diverge — a parity test (same fixture, same verdict via both paths) closes the only untested seam between what the tests prove and what production runs.

### Follow-ups (this PR or noted for later — builder's call, say which)
- `arch.md:1795` still describes the agy gate as a `placeholderFgPalette` rule; the definition of "marker present" changed for agy. Two-line touch or explicit MAINTAIN deferral.
- Commit the capture/sanitization harness (Spec 1313 precedent: `codev/spir-1313-captures/*.mjs`) so re-measuring against a future agy doesn't start from the PR description.
- Cosmetic: required-fixtures test title still says "idle/draft/trust" while requiring seven.
- Keep the dual 4-bit/256-color palette-12 matching — it is deliberate, not redundant (real fixtures use `[94m`, synthetic `[38;5;12m`).

### For the maintainer
- **Merge-order note:** this PR and parked PR #1487 (issue #1471) both touch `render-gate.test.ts` in different regions; whichever merges second may need a trivial rebase.
- The 3 pre-existing `tower-api.e2e.test.ts` failures the builder flagged were proven unrelated by revert-rebuild-rerun at the branch base; also note issue #1488 (the AIR `e2e_tests` check is a no-op that masked them).

Parked for maintainer approval + merge; we are not maintainers.

mohidmakhdoomi:
## CMAP items addressed — all four taken in-PR, none deferred

### Requested before the gate ✅
**agy case in the production-mirror-path suite** (`render-gate.test.ts`). Agreed on the reasoning: these anchors are the first classifier input that depends on cursor *state*, so the transient-vs-mirror seam is newly load-bearing. Added a parity case **per agy fixture** (7 of them) rather than one representative — same bytes through both paths, `toEqual` on the whole verdict, plus an assertion that the transient path's own verdict matches the filename, so the parity check can't pass by both paths being wrong together.

Each fixture is fed **twice**: once in production-sized 64 KiB chunks, and once in **7-byte chunks that deliberately split escape sequences across `feed()` calls**. That second pass is the one with teeth here — the cursor-positioning CSI (`ESC[<row>;<col>H`, and the relative `ESC[2A`/`ESC[2C` agy actually emits) is exactly what a torn chunk boundary would corrupt, and a mis-parsed cursor is now a *verdict* change rather than a cosmetic one. Both pass. 62 tests in the file; full package suite green (4876 passed / 48 skipped).

### Follow-ups — taken, not deferred
- **`arch.md`** ✅ — the agy entry now records that "marker present" is more than a text match, names both anchors, and states the fail-toward-`no-composer-marker` → HOLD direction.
- **Capture/sanitization harness** ✅ — committed at `codev/air-1474-captures/` (capture script, sanitizer, README covering auth prerequisite, per-state recipes, the 110×32 geometry constraint, and how the torn fixture is derived). Verified rather than assumed: re-ran the committed sanitizer against the raw captures and re-derived four committed fixtures **byte-identically**.
- **Cosmetic test title** ✅ — now lists all seven agy states.
- **Dual 4-bit/256-color palette matching** — kept as-is, per your note that it is deliberate.

### One thing the harness work turned up
The sanitizer's leak check was **self-defeating**: it tested for the bare prefixes `/home/` and `/tmp/`, but the replacement path is `/home/agent/project`, so it reported a leak on every successful run — including on the fixtures already in this PR. A check that always fires checks nothing, and I'd rather not ship that as the documented safety step. It now matches paths that are *not* the placeholder, and **refuses to write** the fixture if anything survives. Re-verified against all four captures: `leaks=[]`, output byte-identical.

To be clear about what this did and didn't mean: the committed fixtures were always clean — I verified that independently with a direct grep before the first push, and the byte-identical re-derivation confirms it again. The defect was in the *check*, not in the fixtures.

Ready for the `pr` gate.

waleedkadous:

mohidmakhdoomi:
## Brought up to current `main` — conflicts resolved, CI green, architect integration check

This branch had gone `CONFLICTING` (770 commits behind). It is now **`MERGEABLE`** at `470d49b2d`, with **all 7 checks passing**, via a **merge** of `origin/main` (`818337fa2`) — not a rebase, per this repo's merge-never-squash convention. (`mergeStateStatus: BLOCKED` is now only the missing maintainer approval; no conflict, no failing check.)

### The conflict was adjacency, not disagreement

Exactly one file conflicted: `packages/codev/src/agent-farm/servers/render-gate.ts`. The sole commit to touch it since the merge-base was #1573 (`35f2e637f`), whose change there is **purely additive** — one new exported `bufferLines(term)`, inserted directly above `findMarkerRow`, whose doc comment and signature this branch had rewritten for the cursor-row/palette anchors. Both sides kept verbatim.

The two functions now sit adjacent and are worth keeping straight:

- **`screenLines()`** — viewport only. The gate's classification path; #1474's anchoring lives here.
- **`bufferLines()`** — viewport **plus scrollback** (#1573), for post-delivery echo verification only. Not a classification input, and the marker anchoring does not touch it.

### Architect verification (independent of the builder's report)

- **No collateral loss.** `bufferLines` is preserved at `render-gate.ts:195`; #1573's `watchEchoOnScreen` and `SETTLE_BEFORE_WRITE_MS` are intact in `mailbox-delivery.ts`. I audited every deletion this branch makes against `main` — each is a line this feature supersedes, nothing of #1573's.
- **The feature is intact and is the point of the diff.** `findMarkerRow` now takes `(lines, profile, buf, top, cursorRow, cell)`, reading `buf.cursorY`/`buf.cursorX` at `:348-349` — the cursor-row anchoring this issue asked for.
- **Premise re-verified on current `main`:** `AGY_MARKER = /^> /` is still at `gate-profiles.ts:72`. The only commit to that file since the merge-base was the unrelated `afx reset` → `afx refresh` rename (`f0fa43afe`), which nothing here references.

Post-merge suite: **275 files passed / 3 skipped / 0 failed; 5477 tests passed**, typecheck clean. The render-gate suite is 66 cases now rather than 55 — `main`'s own additions to that file merged in cleanly alongside these.

### Two notes for whoever merges

1. **Merge-order:** PR #1487 (issue #1471) also touches `render-gate.test.ts`, in a different region. Whichever of #1487 / #1491 merges second may need a trivial rebase.
2. **Resuming any long-parked branch in this repo:** the first post-merge `pnpm test` here reported 14 failures in `request-auth.test.ts` / `tower-routes.test.ts` (`WS_KEY_PROTOCOL_PREFIX`, `TOWER_KEY_HEADER` arriving `undefined`). **All false** — `packages/types/dist` predated the constants `main` had since added, and `node_modules` predated a new `three` dependency. `pnpm install && pnpm build` first, *then* believe a test result.

### Status

Parked for maintainer review — **we are not maintainers of this repo, so we are not merging this.** Please approve and merge when it suits you. Issue #1474 stays open and the worktree stays intact until you do.

waleedkadous (changes requested):
Really thorough work — seven real agy captures, per-fixture mirror-vs-transient parity tests (including the 7-byte chunking that splits cursor CSIs), and a failure direction that stays HOLD rather than deliver. The render-gate change itself is approve-quality and claude's lane approved it outright; the one thing I'd like fixed before merge is in the capture tooling, because it's a leak-safety guarantee that currently isn't one:

1. **`codev/air-1474-captures/sanitize.py` writes the output before running its leak check** (the `open(dst, …).write(data)` at ~line 137 precedes the `# Leak check` block). If identifiers survive, it raises only after the unsafe fixture is already on disk — contradicting its own "REFUSING to sanitize" message and creating an accidental-commit risk. Run every check before writing, or delete `dst` on failure. Relatedly, its path regex misses `/Users/…`, so a macOS capture would pass the check while leaking a username (the committed fixtures are clean; this is about the next capture).

Two small things you might consider alongside, neither blocking:
- `buf.cursorY` is baseY-relative in xterm's public contract; the code relies on `viewportY === baseY`. `buf.baseY + buf.cursorY - buf.viewportY` removes that assumption for free, and it's now load-bearing for delivery.
- Idle-session drift is silent — `onLiveness` gates on recent output, so a re-themed agy sitting idle holds forever with no alarm, the same shape as the bare-`>` bug this fixes. A follow-up on profile-drift observability (classifier-stuck escalation for idle sessions, or an `afx doctor` check that classifies each live mirror once) would close that.

Also noting `markerFgPalette` will want to accept multiple colors before agy ships a truecolor theme — follow-up territory.

mohidmakhdoomi:
Thanks — the blocking item was right, and chasing it turned up a second defect in the same file.

**Blocking: `sanitize.py` wrote before checking.** Fixed by check-then-write rather than delete-on-failure, for the reason you gave: a guarantee that depends on cleanup running is weaker than one that never creates the file. `PATH_RE` gains `/Users/…` — you were right that the committed fixtures are clean and this is about the next capture; the underlying rule is that the prefixes the scrubber rewrites and the prefixes the leak check inspects have to be one list, or the check is narrower than the risk.

Writing the self-check exposed a **second** defect, in the predicate I had already "fixed" once. Same-length substitution truncates a cwd shorter than the placeholder, so `/home/ab/x` → `/home/agent/p`, which failed `startswith(PLACEHOLDER_PATH)` and was reported as a leak — every short-cwd capture would have been unsanitizable. Fail-safe, so it would have surfaced as a mysterious refusal rather than a leak, which is the kind of thing that gets worked around by disabling the check. `is_placeholder()` now accepts prefixes of the placeholder too, and all seven committed fixtures re-verify `leaks=[]` under the widened predicate.

Per your note about proving rather than asserting: `sanitize.py --selftest` stubs out the redaction, pushes a leaking capture through, and stats the output path to confirm no file exists. `air-1474-sanitize-ordering.test.ts` runs it in CI. The rest of the harness still stays out of CI (it needs an authenticated `agy` and a PTY), but this part's failure mode is committing someone's username, so it runs. Mutation-checked: restore write-then-check and the test fails with `leaking capture left /tmp/…/fixture.txt on disk`.

**Non-blocking, `cursorY` vs `viewportY`: taken, and it was not hygiene.** I went to write a defensive test and measured a real false clean. Scroll the viewport up and the unconverted row lands on a *stale composer still in scrollback* — empty, palette-12, rule beneath it — so the region bounds and classifies `empty`. CLEAN, while the live composer is off-view holding a half-typed draft. Confirmed against the built old code (`{clean: true, detail: "empty"}`) and the fix (`no-composer-marker`). `buf.baseY + buf.cursorY - top` removes it; comments at both sites updated.

The regression test pins the rigging itself — that the viewport is genuinely off the bottom, that the two conventions disagree, and that the row the old reading picked really was a marker+rule pair — so it fails loudly rather than quietly becoming a tautology if xterm changes. Mutation-checked the same way.

**Non-blocking, not done here, and I agree it is a real gap:** silent idle-session profile drift. `onLiveness` gates on recent output, so a re-themed agy sitting idle holds forever with no alarm — structurally the same failure as the bare-`>` bug this PR fixes, which is what makes it worth its own issue rather than a rider on this one. Same for `markerFgPalette` accepting multiple colors before agy ships a truecolor theme. Both are being filed separately.

`pnpm test`: 276 files / 5479 tests pass, 0 failed.

waleedkadous (dismissed):
Re-verified: `sanitize_file` now finds leaks before any write and refuses with nothing on disk, backed by a self-test that stats the path; `/Users/…` added to `PATH_RE` so the scrub set and the leak-check set are one list; and the short-cwd false-positive you found while writing the self-test is exactly the kind of second defect a real check surfaces. CI 7/7 on 8d9d666c8. Approving and merging.

Issue #1489, opened 2026-08-17
Rename `afx reset` to `afx refresh` (keep `reset` as a deprecated alias for one release)


waleedkadous:
## What

Rename the builder context-reset command `afx reset` (shipped in Spec 1273, PR #1320) to **`afx refresh`**. Same behavior, same flags — this is purely a naming change.

## Why

"reset" miscommunicates the command's core guarantee. In git-adjacent muscle memory *reset* means "throw work away," while this command's whole point is the opposite: it refuses to clear until the builder's saved state is verified (nonce, min-bytes, quiet-window — every gate fails safe without clearing). A name that reads as destructive works against a feature built to be safe. "refresh" says what actually happens: same state, fresh context window. It also matches the vocabulary of the upcoming porch-triggered automatic flow (#1470, "builder context refresh"), so the manual command and the automatic flow share a vocabulary.

## Prescribed design (no open decisions)

1. `afx refresh <builder>` becomes the canonical command — help text, docs, and skill files all lead with it.
2. `afx reset` remains a **working alias for one release**: it runs identically but prints a one-line notice (`afx reset has been renamed to afx refresh; this alias will be removed in a future release`) to stderr. Remove it in a later release — file nothing for that now, just note it in the deprecation line.
3. `--dry-run` output, log lines, and error messages that name the command say `refresh`.
4. Sweep every reference across **both trees** (`codev/` and `codev-skeleton/`) and the Claude skill files (`.claude/skills/afx/`): docs, resources, prompts, lessons. Grep for `afx reset` repo-wide before claiming done.
5. Internal identifiers (function names, file names like `reset.ts`) may be renamed or left as-is at the builder's discretion — the user-facing surface is what matters; don't let an internal-rename diff balloon the PR.
6. Tests: existing reset tests keep passing against `refresh`; add one test pinning the alias (runs + prints the notice).

Issue #1481, opened 2026-08-17
afx send: add --interrupt-after <seconds> (hold, then force-deliver after a bounded wait)


mohidmakhdoomi:
## What

Add an `afx send --interrupt-after <seconds>` flag: attempt normal gated delivery, **hold** (as today) while the recipient's prompt is busy, and if it's still undelivered after `<seconds>`, **escalate to an interrupt delivery** (Ctrl+C the turn, then deliver) — a bounded, opt-in timeout that force-delivers only after patience runs out.

This sits between the two existing shapes:

- `--interrupt` — force **now** (urgent: deliver immediately, claim-first).
- `--delay <s>` — wait `<s>`, then gated-deliver, **never** force.
- `--interrupt-after <s>` (new) — try gated delivery immediately, keep holding, then force **after** `<s>` if still held.

Guidance to document: use `--interrupt-after` for **time-sensitive** messages (they should land soon, but a clean prompt is preferred if one appears in time); use `--interrupt` for **urgent** messages that must land right now.

## Why

Today the only escape from an indefinite hold is `--interrupt`, which force-injects immediately even when the recipient might reach a clean prompt a few seconds later. For time-sensitive-but-not-urgent mail (a cron nudge, a "wrap up soon"), the operator wants "prefer a clean delivery, but don't wait forever." There's no way to express that bounded-patience semantic now — it's all-or-nothing between force-now and wait-forever.

## Notes

- Builds directly on existing machinery: the `not_before`/eligibility columns and escalation-age clock from the 1313 durable-`--delay` round, plus the delayed-interrupt reshape (fire Ctrl+C at due time, body delivers through the gated drainer after the turn). `--interrupt-after` is essentially "hold now, arm an interrupt at `now + s`."
- This is a **deliberate, bounded exception** to the no-force principle — opt-in per message, and only *after* the gate has been given `<s>` to find a clean prompt. Keep the immediate-gated path the default; document the force semantics on the flag.
- Interacts with the interrupt-vs-mailbox interaction under evaluation in #1365 (which absorbed the evaluation mandate from #1480) — settle that model first (or design `--interrupt-after` to route its forced write through the same serialized path).
- Consider the escalation/starvation-alarm interplay: an `--interrupt-after` row is "will self-resolve," so it shouldn't trip the held-mail starvation owner-notice before its deadline.
- Refs #1313 (PR #1330).

waleedkadous:
[...]

PR #1492, opened 2026-08-18
[PIR #1365] Serializer convergence: one lock at the terminal write edge

### LINKED ISSUE #1365: Serializer convergence: route mailbox write edge through submitToSession (serialize gated deliveries vs interrupt/escape)

mohidmakhdoomi:
# PIR Review: Serializer convergence — one lock at the terminal write edge

Fixes #1365

## Summary

The gated mailbox delivery path and the `--interrupt` / `--escape` paths held **disjoint**
locks (per-agent vs per-terminal), so they could interleave on one terminal. The failure that
mattered was not a garbled composer but a **false `delivered`**: a `^C` landing inside a
delivery's own text→Enter window cleared the composer, the delivery's Enter submitted nothing,
every byte still reached the PTY so the write reported success, and the row was marked
delivered for a message the agent never saw. This PR routes the delivery's write edge through
the same per-terminal submission lock, taken as a leaf inside the per-agent serializer
(order: per-agent → per-terminal, no cycle), and lands the resulting model as one documented
boundary instead of three separately-reasoned decisions.

The issue asked for an evaluation *before* a remedy. That evaluation is in
`codev/plans/1365-serializer-convergence-route-m.md` Part 1; it ratified convergence, and
this is its implementation.

## Files Changed

Implementation:

- `packages/codev/src/agent-farm/servers/session-submit.ts` — `trySubmitToSession`,
  `isSubmissionInFlight`, `OPERATOR_SUBMIT_WAIT_CEILING_MS` + `SubmitOptions`, `SubmissionKind`
  + `pendingOperators`, `unserializedWriteCount` / `watchBypasses`, and the rewritten boundary
  comment
- `packages/codev/src/agent-farm/servers/mailbox-delivery.ts` (+144 / −20) — `DeliverySession.id`,
  `WriteAbort` / `WriteResult`, the in-lock precheck, the outcome mapping
- `packages/codev/src/agent-farm/servers/message-write.ts` — `submitMessagePaced`
  (replaces `writeMessagePaced`), `PacedSubmitResult`
- `packages/codev/src/agent-farm/servers/tower-routes.ts` — wait ceiling at the three
  operator call sites, `logCeilingExpired`, `degraded` on the send response, updated scope comments
- `packages/codev/src/agent-farm/servers/mailbox-wiring.ts` — binds the new write edge
- `packages/codev/src/agent-farm/commands/send.ts` — warns the sender on a degraded write
- `packages/sdk/src/tower-client.ts` — `degraded` / `degradedReason` on the sendMessage result

**Two of those are outside the 21-file scope this PR originally stated** (`commands/send.ts`
and the SDK client), added by the review round below. They are the minimum needed to make a
degraded operator write visible to the *sender* rather than only to the Tower log, which
required crossing the server→client boundary. Flagged so the diff holds no surprises; the
boundary rule itself is respected (`codev-sdk` still imports only `codev-types`).

Tests:

- `packages/codev/src/agent-farm/__tests__/spec-1365-serializer-convergence.test.ts` — new
- `packages/codev/src/agent-farm/__tests__/spec-1313-paced-write-drop.test.ts` — re-pointed
- `tower-routes.test.ts`, `send-delivery.test.ts`, `send-mailbox-repro.test.ts`,
  `cron-delivery.test.ts`, `send-architect-identity.test.ts` — fakes updated

Evidence + docs:

- `packages/codev/scripts/spec-1365-e2e-evidence.mts` (+420 / −0) — dev-approval evidence script
- `codev/evidence/1365-dev-approval-transcript.txt` (+80 / −0) — its transcript
- `codev/resources/arch.md` (+10 / −2), `codev/resources/arch-critical.md`,
  `codev/resources/lessons-learned.md`
- `codev/plans/1365-...md`, `codev/state/pir-1365_thread.md`

## Commits

- `2adbe0b9` [PIR #1365] Plan draft: evaluation of the three write paths + convergence design
- `a6cdbe27` [PIR #1365] Plan revised (rev 2): all 5 blocking review items + interrupt-latency ceiling
- `30af22b2` [PIR #1365] Lock primitives: try-acquire, wait ceiling, degraded-write counter
- `e9fd2d42` [PIR #1365] Route the mailbox write edge through the per-terminal submission lock
- `194685e1` [PIR #1365] Tests: interleaving, in-lock precheck, liveness, ceiling, key hygiene
- `dcf22a5f` [PIR #1365] Document the converged write-edge model in one place
- `54d57008` [PIR #1365] Thread log: implement phase
- `ee3a17df` [PIR #1365] dev-approval evidence: 4 scenarios against an isolated live Tower
- `1483d65c` [PIR #1365] Thread log: dev-approval evidence
- `ad824bf2` [PIR #1365] Review + retrospective
- `ece06a5e` [PIR #1365] Fix codex/claude finding: kind-aware ceiling + report degraded writes
- `0dc75d8e` [PIR #1365] Review round 2: byte-accurate bypass count, counter eviction, claim sweep

## Test Results

- `pnpm --filter @cluesmith/codev-sdk build`: ✓ pass
- `pnpm --filter @cluesmith/codev test`: ✓ pass — **4934 passed / 0 failed / 48 skipped**,
  248 files, after merging `origin/main`. 28 new tests in
  `spec-1365-serializer-convergence.test.ts`, plus the degraded-interrupt response test in
  `tower-routes.test.ts`. (Before the merge, on the #1365 work alone: 4885 / 0 / 48, 246 files
  — the delta is main's own tests arriving, not tests changing behaviour here.)
- **Merged `origin/main` after the `pr` gate was approved** (merge `ebbc495dc`, not a rebase —
  the reviewed history is preserved). `main` had advanced 25 commits during the review round
  (AIR #1489's `afx reset` → `afx refresh`, secfix-1's Tower auth hardening, PIR #1495). One
  conflict, in `codev/resources/lessons-learned.md` § Architecture: an append-only collision
  (three #1365 entries vs one secfix-1 entry at the same point), resolved **keep-both**, all
  four entries present and unmodified. Every code file auto-merged; every #1365 hunk survives
  byte-identical, and `message-write.ts`, `mailbox-delivery.ts`, `commands/send.ts` and
  `spec-1365-serializer-convergence.test.ts` were untouched by the merge.
- **Manual verification** (dev-approval gate, human-approved): `afx dev` was not usable —
  4100 is shared by design and restarting the live Tower kills every builder session — so the
  running-worktree evidence was scripted against an **isolated Tower on port 14650** with real
  shellper-backed PTYs and real HTTP endpoints (routes → mailbox → render gate → locks → PTY,
  nothing stubbed). **66/66 checks.** Full transcript:
  `codev/evidence/1365-dev-approval-transcript.txt`; script:
  `packages/codev/scripts/spec-1365-e2e-evidence.mts`.
  - S1 — 10× long multi-line send raced by `--interrupt`: 10/10 bodies reached the wire
    **whole**, zero fragmentation, zero duplication.
  - S2 — `--delay 5 --interrupt` mid-turn: scheduled; `^C` at due time; body did **not** land
    mid-turn; landed **exactly once** after the prompt cleared; `^C` before body.
  - S3 — interrupt vs a busy line: returned in **2156 ms** rather than waiting out a ~4.1 s
    paced write; Tower logged the degradation at WARN; the raced delivery reported
    `preempted` and **held its row** (`delivered=false held=true reason=busy`).
  - S4 — `--escape` unchanged (ESC + Enter on the wire); a dead terminal is refused, never
    silently dropped.
  - Live Tower on 4100 verified healthy and untouched afterwards; no orphan processes.

## Architecture Updates

**COLD — `codev/resources/arch.md` §7 item 5** (rewritten). The old text described the two
locks as disjoint with the cross-path race as an accepted boundary; that is now false. The
replacement carries the whole model in one place: which writers take the lock and which stay
deliberately uncovered, the per-agent → per-terminal order and why there is no cycle, the
deliveries-decline / operators-block asymmetry and the reason each side differs, the wait
ceiling and its degradation, the delayed-interrupt sequencing, and — stated honestly — that
serialization is the structural guarantee while the in-lock precheck only *narrows* the
echo-lag residual (#1473).

**HOT — `codev/resources/arch-critical.md`**: the existing mailbox-first fact already governs
"any new message writer", so the lock-order invariant was **appended to that fact** rather
than added as an eleventh. This keeps the tier at its 10-fact cap with **no displacement** —
the hot tier gains the one clause a future author actually needs at decision time ("take
`submitToSession`; order is per-agent → per-terminal"), not a second entry on the same
subject.

## Lessons Learned Updates

**COLD — `codev/resources/lessons-learned.md` → Architecture**, three entries:

1. *"Every byte reached the PTY" is not "the message landed."* A write-success boolean sees
   transport acceptance, not semantic loss — a `^C` that clears the composer leaves every
   write returning `true`. Any success signal derived from "did the transport accept the
   bytes" needs a second question, answered from a source the transport can't lie about.
2. *A structurally-typed port makes an omitted field compile **and** pass.* When a value
   becomes a **key** (lock, cache, registry), assert its presence at the boundary — type-checking
   the shape does not check the key. Here a double without an `id` keyed every per-terminal
   lock on `undefined`: a silently global lock, no failing assertion anywhere.
3. *Converging two locks is not just "take the same lock" — the acquisition **policy** has to
   match each caller's liveness needs.* Blocking would have regressed both sides (the
   sequential drainer, and the human's escape hatch). The shape that works is asymmetric:
   the background writer declines contention and retries on its existing schedule; the
   operator waits, but boundedly, degrading to documented prior behaviour rather than a hang.

**Considered for HOT and deliberately not promoted**: both are real but narrower than the
current ten hot lessons, and promoting either would require *displacing* an existing one.
Displacement at the cap is the maintainer's call, not a builder's — flagged here rather than
taken unilaterally.

## Review Round: two REQUEST_CHANGES, and what happened to each

PIR's consultation is **single-pass** — there is no second automated round — so the human at
the `pr` gate is the only remaining reviewer of these dispositions. They are written out in
full rather than summarised.

Two independent review sets ran, and they did not agree:

| Reviewer | Protocol CMAP (`codev/projects/1365-.../`) | Architect's CMAP on PR #1492 |
|---|---|---|
| gemini | APPROVE | APPROVE |
| codex | APPROVE | **REQUEST_CHANGES** |
| claude | **REQUEST_CHANGES** | **REQUEST_CHANGES** |

My own codex lane approved; the architect's codex lane found a real bug. I verified every
finding against the code before acting on it, and **none was dismissed on the strength of
another lane's APPROVE**. Both REQUEST_CHANGES lanes converged independently on the same two
blocking findings.

**Blocking 1 — the ceiling could bypass another *operator's* submission.** ACCEPTED, real,
fixed in `ece06a5e`. `bounded` keyed only off "is anything in flight" without asking *what
kind* of writer was ahead, so a second `--interrupt` could skip a first one carrying a long
body after 2 s. Operator-vs-operator was **always** fully serialized before #1365
(`submitToSession` had no ceiling at all — it is Spec 1273's `/clear` fusion bug), so my
ceiling made that one pair strictly *worse* than the status quo. That also falsified this
document's own "never worse" claim. Fix: chain entries carry a `SubmissionKind`, a
`pendingOperators` count tracks operators **queued as well as in flight**, and the ceiling arms
only when nothing ahead is an operator. Queued has to count — bypassing an operator that has
not started yet is the same violation as bypassing one mid-write. Pinned by *"operator vs
operator NEVER degrades"* and *"a THIRD operator does not bypass a QUEUED one"*. Note the
pre-existing ceiling test needed its holder changed from an operator to a delivery: **that
fixture change is the behaviour change**, not a workaround for it.

**Blocking 2 — a ceiling-degraded `--interrupt` still reported unqualified success.**
ACCEPTED, real, fixed in `ece06a5e`. The row is claimed `delivered` before the write, so a
degraded interrupt returned `delivered: true` with only a Tower-side WARN — the same
lying-success-signal class this whole issue exists to remove, relocated from the delivery path
to the operator path. Claim-first is *kept* (un-claiming risks a double delivery, reasoned
through at CMAP round 3 of the implement phase); what changed is that the truth is now
surfaced: `/api/send` returns `degraded: true` + `degradedReason`, threaded through the SDK
client and warned about by `afx send`. An indicator nobody surfaces is half a fix. Pinned by
*"a body-bearing interrupt that crosses the wait ceiling reports degraded"* in
`tower-routes.test.ts`.

**Non-blocking, taken anyway (this commit):**

- *The bypass counter was bumped on ceiling expiry regardless of whether bytes went out.* The
  delayed `^C` re-checks `isStillLive()` / `writable` **inside** the lock and can return having
  written nothing; that no-op was still counted, forcing a concurrent delivery into a spurious
  `preempted` re-delivery. The counter answers "did bytes bypass the lock while I held it?", so
  only bytes may bump it: `SubmitOptions.wroteBytes` is consulted straight after the write
  callback, with no `await` in between, so the ordering guarantee the old placement provided is
  unchanged. Pinned by *"a degraded write that writes NOTHING is not counted as a bypass"*.
- *`unserializedWrites` was never pruned* — one entry per session that ever degraded, retained
  for the life of the Tower. The leak class #1472 just fixed. It cannot self-delete on drain
  the way `chains` and `pendingOperators` do, because it must **outlive** the submission whose
  watcher is about to compare against it: a reset landing between a watcher's two reads would
  read as "nobody raced me" — the exact false `delivered` this issue exists to eliminate. So
  eviction is interlocked with an explicit `watchBypasses` window: refused while a watch is
  open, attempted from *both* the chain's drain cleanup and the last watch's release, so
  whichever runs second is the one that evicts and no ordering leaks. This needs **no
  session-teardown hook** and therefore no `terminal/` → `agent-farm/` layer crossing. Pinned by
  *"the degraded-write counter is evicted once the session goes idle"* and *"eviction cannot
  land inside a watcher window and mask a race"*.
- *Stale `{@link writeMessagePaced}`* in `message-write.ts` — repointed at `submitMessagePaced`.
- *`DEGRADED_SUBMIT_REASON` was inserted between `logCeilingExpired`'s JSDoc and its function*,
  orphaning the comment — moved above it.
- *The residual "never worse than the status quo" claim-sites* — swept and rewritten to state
  the guarantee **per pair** (op↔op unchanged and unbounded; op↔delivery serialized under the
  ceiling and degraded to the old disjoint-lock behaviour above it; delivery↔delivery
  unchanged). `arch.md` §7 item 5 and the `session-submit.ts` boundary comment were corrected in
  `ece06a5e`; `tower-routes.ts`'s `logCeilingExpired` doc comment and the degraded-path inline
  comment in this one.

**Non-blocking, NOT taken — flagged instead:**

- *The ceiling timer is not cancelled when the predecessor wins the race.* `Promise.race`
  leaves a ≤2 s `setTimeout` pending whose resolution is then discarded. Cancelling it means
  adding abort semantics to the injected `SubmitClock` interface, which every test double
  implements. The cost of leaving it is one short-lived timer per *contended* operator
  submission; the cost of fixing it is a broader interface change late in a review round. A
  reviewer who disagrees should say so — it is a small change, just not a free one.
- *`waited < 100` / `tickMs < 250` are timing-sensitive under CI load.* Real, and deliberate:
  these are the assertions that make "the drainer does not stall" and "the escape hatch stays
  responsive" *testable* claims rather than prose. Both have ≥2.5× headroom over the behaviour
  they exclude. If they flake in CI, raising the bounds preserves the property.
- *Hot-tier `arch-critical.md` was appended to directly, where the plan said it would be
  proposed.* Disclosed below; it is the human's call, and reverting it is a one-line edit.

1. **The in-lock precheck's honest status.** With try-lock semantics the delivery never waits,
   so in production today the precheck cannot observe a state change a pre-lock check missed —
   no macrotask can interleave between them. I kept it (both plan reviewers asked for it, the
   human ratified it) but documented its *actual* value rather than implying it closes a live
   race: it backstops the injected port boundary, and it is what keeps the acquisition policy
   a free choice if #1481 later wants the delivery to wait behind an interrupt. If a reviewer
   would rather not carry code whose value is conditional on a future change, this is the
   place to say so.
2. **The wait ceiling is a judgment call** (`OPERATOR_SUBMIT_WAIT_CEILING_MS = 2000`, human-
   ratified at the plan gate). It exists because `--interrupt` previously never waited, and a
   paced write runs `(lines−1)×10+80` ms against a body capped only by `parseJsonBody`'s
   1 MiB — a 48 KB `--file` of short lines is ~8 minutes. It arms **only against a delivery
   write**: behind another operator the wait stays unbounded, exactly as before #1365 (see the
   review round above — a ceiling that could skip an operator made that pair strictly worse,
   and that was a real blocking finding, not a hypothetical). So past the ceiling the operator
   write falls back to precisely the pre-#1365 operator-vs-delivery behaviour — two disjoint
   locks, no serialization — which is no worse for that pair, only no longer silent. The
   guarantee is **per pair**, and the 2 s value itself remains a judgment call.
3. **`preempted` trades a possible duplicate for never falsely reporting delivery.** A
   delivery raced by a ceiling-expired write holds its row instead of marking it delivered, so
   if the message *did* land intact the gate may deliver it again later. That is the same call
   the existing dropped-write branch already makes, and the opposite of the interrupt path's
   claim-first tradeoff — the asymmetry is deliberate (an operator's own message vs an
   autonomous background delivery), but it is worth a second opinion.
4. **Port signature churn.** `writeMessage` gained a 4th parameter and a typed result across
   six test files. One override (`send-delivery.test.ts:604`) previously returned `undefined`
   and relied on falsy ⇒ hold; it is now explicit.
5. **`writeMessagePaced` was removed**, not deprecated — its only live caller was the mailbox
   wiring. Its drop-semantics test is re-pointed at `submitMessagePaced` so the #1198
   silent-loss guard stays on the live write edge rather than on a function nothing calls.
6. **The dev-approval transcript is committed under `codev/evidence/`, a new directory.** That
   placement is a deliberate choice, not an accident: the evidence is part of the PIR record
   for this project, the way `codev/specs/`, `codev/plans/` and `codev/reviews/` are, and a
   gate approved on evidence that then vanishes leaves the approval unauditable. It is
   nonetheless a new top-level convention in the repo, and **the maintainer may veto it** —
   moving or dropping the file changes nothing else in the PR (the generating script,
   `packages/codev/scripts/spec-1365-e2e-evidence.mts`, is re-runnable and is the durable
   artifact).

**Interlock for #1481 (`--interrupt-after`)**: "interrupt, then deliver this body" is now
expressible as ordered acquisitions of *one* lock rather than a race between two. Two
residuals it must design against, both documented: the `^C`→body gap is gate-mediated and
deliberately **not** atomic (the delayed interrupt guarantees "the turn was ended", never
"this body is next"), and a no-op `^C` is only logged.

## How to Test Locally

- **View diff**: VSCode sidebar → right-click builder `pir-1365` → **Review Diff**
- **Run dev**: `afx dev pir-1365` — but note it will contend for the live Tower's port; the
  isolated-Tower script below is why the dev-approval evidence took that route instead
- **Re-run the evidence**: `pnpm --filter @cluesmith/codev build && node
  --experimental-strip-types packages/codev/scripts/spec-1365-e2e-evidence.mts`
  (isolated Tower on 14650, ~90 s, exits non-zero on any failed check)
- **Unit**: `pnpm --filter @cluesmith/codev test spec-1365-serializer-convergence`
- **What to verify**: a long multi-line send raced by `afx interrupt` never fuses and never
  leaves an `afx inbox show` row reading `delivered` whose text is absent or partial; another
  agent's mail keeps flowing while a large body is mid-delivery; `afx interrupt` stays
  responsive against a busy line; `--escape` behaviour is unchanged.

## Flaky Tests

None. No tests were skipped or quarantined, and no pre-existing unrelated failures were
touched.


mohidmakhdoomi:
## Architect integration review — 3-way CMAP (risk tier: High — core write edge, session-submit/mailbox-delivery)

**Verdicts: gemini APPROVE · codex REQUEST_CHANGES · claude REQUEST_CHANGES — all HIGH confidence, and the two RC lanes independently converged on the same finding and the same fix.**

### The blocking finding (verified by the architect against the branch)
The 2s operator wait ceiling **narrows Spec 1273's operator-vs-operator guarantee**: pre-#1365, two operator submissions to one terminal were serialized unboundedly; with the ceiling, a second operator action can degrade past a long body-bearing first one and land inside its text→Enter window — and since the interrupt path claims `markMailboxDelivered` before its write (claim-first, correctly retained) and has no preempted detection of its own, the false-`delivered` failure this PR removes for deliveries is reintroduced for operators. The PR's "never worse than the old status quo" claim is therefore wrong for the operator-operator pair (it remains true for operator-vs-delivery, which was never serialized pre-PR).

**Directed fix (both lanes agree):** record the lock holder's kind in the chain and arm the ceiling **only when a delivery write holds the lock** — that bounds exactly the new exposure D3 was aimed at (the escape hatch stalling behind a long *delivery*) while leaving Spec 1273's operator serialization intact and making the "never worse" claim provably true per-pair. Plus: surface a degraded/raced indicator on the operator response alongside the existing WARN; correct the four claim-sites (`session-submit.ts:249-253`, `:161-171`, `tower-routes.ts:1623-1625`, arch.md §7 item 5); add tests for a body-bearing interrupt crossing the ceiling and for two long operator submissions.

### Non-blocking (in-PR if cheap, else flagged)
- Bump the bypass counter only on bytes actually written (a no-op `^C` currently forces a spurious redelivery hold); evict `unserializedWrites` on session teardown (unbounded map on a long-lived Tower).
- `codev/evidence/` + the committed e2e-evidence script: deliberate precedent (evidence lives with the PIR record, like `codev/reviews/`) — maintainer may veto.

### What all three lanes and the architect agree is right
The convergence design itself: leaf per-terminal lock inside the per-agent serializer (cycle-free, all PTY writers accounted for), delivery fail-fast preserving drainer liveness, in-lock precheck with row-status re-check, the preempted counter sampled around the delivery write, honest narrowing (not closing) of the #1473 echo-lag residual, 23 paired tests each with a bypass-the-lock control, and 66/66 live-wire dev-approval evidence (`codev/evidence/1365-dev-approval-transcript.txt`) demonstrating the ceiling and the preempted-hold end-to-end on real PTYs.

Awaiting the fix push; the pr gate follows. Parked for maintainer approval + merge; we are not maintainers.

mohidmakhdoomi:
## Architect fix-verification — review iteration 1 dispositions (pre-gate)

Verified against the branch at `618925f1` (commits `ece06a5e` + `0dc75d8e`), not from the builder's summary. All accepted findings are actually fixed:

**1. Kind-aware ceiling (converged codex/claude blocking finding) — verified fixed.**
`session-submit.ts`: submissions carry a `SubmissionKind` (default `operator` — the never-bypassable kind, the safe default); a `pendingOperators` per-session count includes *queued* operators (counted before any `await`, after the self-check); `bounded` now requires `!behindOperator`. Operator-vs-operator keeps its pre-#1365 unbounded serialization; the ceiling can only ever bypass a delivery. `trySubmitToSession` tags the delivery path `delivery-write`. The conservative direction (an operator queued behind a delivery *and* an operator waits unbounded) errs toward more serialization, never less.

**2. Degraded interrupt no longer reports unqualified success (blocking) — verified fixed.**
`/api/send` returns `degraded: true` + `degradedReason: 'submit-wait-ceiling-expired'` on both the interrupt and escape paths (additive, present only when it happened — older clients unaffected); threaded through `packages/sdk/src/tower-client.ts` (types + passthrough; SDK still imports only types) and surfaced by `afx send` as a loud, actionable warning. Claim-first is kept and the tradeoff is now told to the sender instead of hidden.

**3. Bytes-only bypass counting — verified.** The bump moved inside the chain: `if (bypassed && (options.wroteBytes?.() ?? true))` immediately after `write()` with no `await` between, so a delivery holding the line cannot observe the bytes without also observing the bump. The delayed-^C path passes `wroteBytes: () => fired`, and `fired` is set exactly where the ^C bytes go out (`tower-routes.ts:1768`).

**4. Counter eviction — verified.** `watchBypasses`/`bypassWatchers` interlock: eviction refused while any watch is open, attempted from both the chain drain and the last watch release; `submitMessagePaced` holds the watch across its write and releases in a `finally`. No session-teardown hook, no `terminal/` → `agent-farm/` crossing. `bypassCountedSessions()` gives tests the leak assertion.

**5. Claim-site sweep — re-ran the grep myself.** Every surviving `never worse` / `status quo` / `never waited` site states the corrected per-pair guarantee (op↔op unchanged/unbounded; op↔delivery ceiling-bounded, degraded to old disjoint-lock behaviour above it; delivery↔delivery unchanged).

**6. Pinning tests — all present, and I re-ran both files: 129/129 green** (`spec-1365-serializer-convergence.test.ts` + `tower-routes.test.ts`, Vitest 4.1.4). Includes `operator vs operator NEVER degrades`, `a THIRD operator does not bypass a QUEUED one`, `a degraded write that writes NOTHING is not counted as a bypass`, `eviction cannot land inside a watcher window and mask a race`, and the `degradedReason` route assertion.

**Refusals — accepted as reasonable.** Ceiling-timer cancellation would force abort semantics onto the injected `SubmitClock` every test double implements, for one ≤2 s timer on the rare contended-operator path; the timing assertions are load-bearing responsiveness claims with ≥2.5× headroom.

**Escalated to the human at the gate (not resolved here):** the hot-tier `arch-critical.md` clause was applied directly rather than proposed. It extends the existing mailbox-first fact (no new bullet; cap holds) and reverting is a one-line edit — ratification is the human's call and is being asked alongside the pr gate.

Two out-of-scope files (`commands/send.ts`, `sdk/tower-client.ts`) are the minimum server→client surfacing for finding 2 and are disclosed in the PR body.

*Contributor-workflow reminder: this PR parks for the maintainer — no self-merge.*

waleedkadous (approved):
Excellent work — this is one of the strongest PRs this subsystem has seen. The control tests that reproduce the pre-fix bug before asserting the fix, the honest degraded-write surfacing all the way to the sender, and collapsing three separately-reasoned write paths into one documented boundary are exactly what this edge needed. Approving.

For the record, what I verified during review:

- **Lock ordering is sound** (per-agent → per-terminal, taken as a leaf; operators never enter the per-agent serializer) — no cycle, and re-entrancy is ruled out since `PtySession.write()` emits no submit signal.
- **The in-lock precheck re-checks the row's own status** (`getById` → `row-resolved`), so a row dismissed while the delivery waited can no longer reach the wire.
- **Drainer liveness holds** under a contended terminal (asymmetric try-lock; verified by the <250ms tick test).
- **The SDK/response changes are additive typing only** — no server/client isolation breach.
- **Field-bug relevance**: this structurally closes the tail-fragment truncation mechanism behind the recent long-message loss report (the write window scales with message length, so any lock-taking writer landing mid-write shredded exactly the long sends) and ends false `delivered` reports from those paths.

I handled the rebase (merge commit, one keep-both docs conflict in `lessons-learned.md`); CI is green on the new head.

One design corner to carry forward — non-blocking, I'll file it as a follow-up issue: `bounded` in `submitToSession` is computed once at enqueue (`contended && ceiling && !behindOperator`). With a long delivery in flight and one operator queued (ceiling armed), a *second* operator arriving meanwhile gets `bounded=false` and can wait out the delivery's entire paced write — unbounded, where the pre-convergence behavior was immediate. It needs a >2s delivery plus two operator actions inside its window, so it's rare; re-arming the bound when the operator ahead drains looks like the right shape, and it belongs with the `--interrupt-after` work that's already sequenced after this PR.

Two accepted tradeoffs I'm explicitly endorsing rather than flagging: `preempted` → possible duplicate delivery (the right direction of error vs silent loss), and the bypass counter's benign scheduling-time false-positive window.

Issue #1139, opened 2026-07-06
vscode: Backlog Reference issue in architect action ignores QuickPick selection (always sends to main)


amrmelsayed:
## Problem

When a user clicks the inline "Reference issue in architect" button on a Backlog row in a workspace with more than one architect, a QuickPick appears to choose the target architect. Whatever the user picks, the reference text (`#<id> "<title>" `) is injected into `architect:main`, not the picked architect. From the user's perspective the picker is a no-op.

Single-architect workspaces are unaffected — no picker, everything correctly routes to main.

## Root cause

Two independently-correct pieces layered wrong:

1. **Spec 786 Phase 6** (`packages/vscode/src/terminal-manager.ts:140` docstring) intentionally defaulted `injectArchitectText`'s `architectName` param to `'main'` to keep the Backlog inline button working without modification while sibling architects were being introduced. Documented explicitly as *"the Backlog button always targets `main` regardless of how many sibling architects exist."*
2. **Issue 841 Gap 2** later added a QuickPick to `codev.openArchitectTerminal` (`packages/vscode/src/extension.ts:782`) so that when the command is invoked with no name AND >1 architects exist, the user is prompted to choose.

The two changes cross wires. The `codev.referenceIssueInArchitect` command (`packages/vscode/src/extension.ts:1019`) invokes `openArchitectTerminal` with no argument (triggering the picker in the multi-architect case), then calls `injectArchitectText(text)` with no name (defaulting to `main`). The picker's selection is captured only for the "which terminal to open," never for "which terminal to inject into."

Concrete chain:

- `extension.ts:1029` — `terminalManager?.injectArchitectText(buildArchitectReferenceInjection(issueId, title))` — no name arg
- `terminal-manager.ts:140` — `injectArchitectText(text: string, architectName: string = 'main')` — defaults, injects into `architect:main`
- Same shape at `extension.ts:1041` for `codev.referencePRInArchitect` (PR sidebar mirror per #1043)

## Fix sketch

Two surgical changes:

1. **`codev.openArchitectTerminal` returns the resolved architect name.** Currently returns void. Return `string | undefined` = the name that was actually opened (arg → picker → default). Keep the single-architect path returning that architect's name so callers can rely on the return value uniformly.

[...]

Issue #907, opened 2026-05-28
vscode: builder briefly appears under 'Uncategorized' area group after cleanup before disappearing


amrmelsayed:
## Symptom

When a builder is terminated (via `afx cleanup -p <id>`, the right-click `Cleanup Builder` action, or any other removal path), the Builders tree shows a transient intermediate state for several seconds:

1. Builder is in its correct area group (e.g. `VSCODE` or `TOWER`) — normal state
2. `afx cleanup` runs → builder's worktree directory is removed, terminal killed, etc.
3. **For ~2-5 seconds: the builder reappears under the `UNCATEGORIZED` group** at the bottom of the Builders tree
4. Eventually disappears entirely from the tree

The intermediate `UNCATEGORIZED` appearance reads as a real reclassification event to the user — they may briefly wonder "why did this builder lose its area?" — when in fact the builder is just being cleaned up.

## When it started

After #818 (Builders area grouping) merged. Before #818, builders rendered as a flat list, so a cleanup-in-progress builder briefly appeared in the flat list before disappearing — no surprising classification jump. After #818, the area-grouping projection means cleanup-in-progress builders with missing/empty `areas[]` get dropped into `Uncategorized` until they're fully gone.

## Plausible root causes (for investigation)

The builder should pick one of these via reproduction with logging, before fixing:

1. **Stale overview data during cleanup window** — The overview cache returns a `OverviewBuilder` record where `areas: []` for the cleanup-in-progress builder. The `groupByArea` helper in `packages/core/src/area-grouping.ts` correctly puts `areas: []` into `Uncategorized`. Fix would be to either filter out cleanup-in-progress builders from the projection, or preserve the original `areas[]` until the builder is fully removed.

2. **Cleanup-induced wire-payload partial state** — `afx cleanup` may emit overview SSE events where the cleanup-in-progress builder has been partially mutated (some fields cleared, others not yet). If `areas[]` gets cleared first while the builder is still in the overview list, it briefly classifies as Uncategorized.

3. **Race between `afx cleanup`'s removal phases** — `afx cleanup` does several things in sequence (kill terminal, remove worktree, drop state.db row). The overview cache refreshes via SSE at various points. A refresh that fires AFTER the worktree directory is removed but BEFORE the `OverviewBuilder` record is dropped from overview would have `areas[]` empty (no labels available because the underlying issue lookup may key off the worktree's existence).

[...]

Issue #1494, opened 2026-08-18
vscode: approval gates should relay through the spawning architect instead of invoking porch approve directly


amrmelsayed:
## Problem

The VS Code approval surfaces write porch state directly. `runPorchApprove`
(`apps/vscode/src/commands/approve.ts:148`) shells out to
`porch approve <id> <gate> --a-human-explicitly-approved-this` with `cwd` set to the main
workspace path. Every approval surface funnels through it: the sidebar ✓, `Cmd+K G`, and the
gate-pending toast fast path (`:114`).

Two consequences:

1. **The spawning architect never learns the gate was approved.** Its model of the builder goes
   stale, and its review at the *next* gate starts from confusion rather than context. This is the
   reported friction.
2. **Porch runs against the main checkout, not the builder's worktree.** Related to #1235
   (status.yaml resolution picks an arbitrary worktree copy).

## Requirement

Replace the direct CLI invocation with a relay to the builder's spawning architect. The human still
decides by clicking; the architect receives the decision and runs `porch approve` itself. This
mirrors how humans already work with codev today: they ask the architect to approve the gate, and
the architect runs the command.

```
today:     sidebar ✓  ──> porch approve            (extension shells out; nobody is told)
proposed:  sidebar ✓  ──> architect ──> porch approve   (architect executes, and is in the loop)
```

The builder is not in this chain. It is woken by porch as it is today.

This is deliberately the *small* version. It reuses the send path the extension already has
(`client.sendMessage(target, msg, { workspace })` → Tower `/api/send`, mailbox-first per Spec 1313).
It does **not** introduce a typed gate-event frame, a system sender type, or a new bus. See
"Relationship to #1194" below.

### Routing

- Target the builder's spawning architect. `spawnedByArchitect` is already on the roster
  (`packages/types/src/api.ts:45` and `:201`) and already reaches the extension
  (`apps/vscode/src/views/builder-grouping.ts:122`), so no new Tower surface is needed.
- **Do not inherit the `|| 'main'` fallback from `builder-grouping.ts:122`.** That is #1406, and a
  misrouted *approval* is materially worse than a misrouted status line: it sends a human's gate
  decision to an architect who did not spawn that builder. Decide the null / unregistered case
  explicitly. Suggested: fall back only when no named architect is registered at all, state the
  fallback in the message body, and surface it in the UI rather than silently rerouting.
- Mailbox-first returns `delivered` or `held(reason)`. A held relay must be visible to the human who
[...]

PR #1500, opened 2026-08-18
vscode: refuse non-live 'main' in open-architect instead of substituting architects[0]

### LINKED ISSUE #1497: vscode: open-architect silently substitutes architects[0] for a non-live 'main', strips the disambiguating label, and captures the text-injection path
## Problem

`extension.ts:920-924` resolves an architect for the open-architect path:

```ts
const match = architects.find(a => a.name === targetName);
const fallback = targetName === 'main' ? architects[0] : undefined;
const target = match ?? fallback;
if (target?.terminalId) {
  await terminalManager?.openArchitect(target.terminalId, targetName, true);
```

When `targetName === 'main'` and `main` is not in the live architect list — a restart window, a
crashed session, a transiently stale overview — this opens **`architects[0]`**, a different
architect, while passing `targetName` (`'main'`) as the identity. Three consequences follow, and
they compound.

### 1. The disambiguating label is stripped, not merely wrong

`terminal-manager.ts:135`:

```ts
const label = architectName === 'main' ? 'Codev: Architect' : `Codev: Architect (${architectName})`;
```

The fallback passes `'main'`, so the wrong architect's terminal opens with the **unqualified**
label — the exact form that denotes `main`. A named architect would normally render
`Codev: Architect (security)`. So the one artifact that could reveal the substitution instead
removes the qualifier that would have revealed it.

### 2. The mis-cached terminal captures the text-injection path

`terminal-manager.ts:152-159`:

```ts
injectArchitectText(text: string, architectName: string = 'main'): boolean {
  const key = `architect:${architectName}`;
  const entry = this.terminals.get(key);
  ...
  entry.terminal.sendText(text, false);
```

`openArchitect` caches under `architect:${architectName}`, so the wrong architect's terminal is
registered as `architect:main`. Until that entry is replaced, **any text injected at `'main'` is
typed into the wrong architect's terminal.** `codev.referenceIssueInArchitect` and
`referencePRInArchitect` resolve through the same name and inherit it.

This is the message-misdelivery hazard, reachable in shipped code with no Stream Deck involvement.
Note also that this path is a direct `sendText`, so none of the Spec 1313 mailbox-first guarantees
apply: there is no persistence, no render-gate, and no `held` result. `sendText(..., false)` does not
submit, but the human sees text appear in a terminal labelled as main and completes the send
themselves.

### 3. It self-corrects only on the next press

`terminal-manager.ts:126-133`: a later `openArchitect` with the same key and a different
`terminalId` disposes the stale terminal and opens the correct one. So the window is "until the next
press", not indefinite. Stated so the severity is not overclaimed — the injection capture lasts as
long as that cache entry does.

## Requirement

The `targetName === 'main'` fallback should not silently substitute a different architect while
claiming to be `main`. Either:

- open the fallback architect under **its own name and label**, so the substitution is visible and
  the cache key matches the actual occupant; or
- **refuse**, and surface the existing "No 'main' architect found — is the workspace activated?"
  warning.

Silently substituting while presenting as `main` is the one behaviour that should not survive,
because it is invisible at the moment it happens and it poisons a name-keyed lookup afterwards.

## Notes

Found while reviewing #1495, when a proposal to pin a permanent `main` key on the Stream Deck's
architect board was declined specifically because pressing it during a `main` flicker would trigger
this path. The defect is in the extension and exists independently of that feature.

Related: #1463 (documented this residual for the single-target key), #950 (message-key hazard),
#1406 (silent owner-fallback misrouting — same "fall back to main rather than refuse" instinct),
#1214 (`role_id || 'main'` silent fallback in Tower identity). The pattern across all of these is
worth naming: a `|| 'main'` convenience that is harmless for display and harmful the moment the
resolved name keys an action.

## Acceptance criteria

- [ ] Pressing open-architect for `main` while `main` is not live either opens the alternate
      architect under its own name and label, or refuses with the existing warning.
- [ ] No terminal is ever cached under `architect:<name>` while hosting a different architect.
- [ ] `injectArchitectText('…', 'main')` cannot reach a non-`main` architect's terminal.
- [ ] Regression test covering the not-live-`main` resolution path.

amrmelsayed:
# PIR Review: open-architect must not substitute a different architect while claiming to be `main`

Fixes #1497

## Summary

`codev.openArchitectTerminal` resolved a non-live `main` to `architects[0]` (a different architect) while flowing the requested name `'main'` onward, so in `terminal-manager.ts` — where the architect name is an address twice over (the `architect:${name}` terminal-cache key and the `injectArchitectText` lookup) — the wrong architect's terminal was cached under `architect:main`, wore the unqualified "Codev: Architect" label, and captured any text later injected at `'main'`. This fix removes the `|| 'main'` fallback so a non-live `main` refuses with the existing warning, and flows the resolved occupant's *own* name onward so the cache key, label, and injection always address the real occupant.

## Files Changed

- `apps/vscode/src/open-architect.ts` (+83 / -0) — new module: `resolveArchitectTarget` (exact-name, no fallback) + `openResolvedArchitect` (opens under the occupant's own name, or refuses)
- `apps/vscode/src/extension.ts` (+11 / -9) — the command delegates to `openResolvedArchitect`
- `apps/vscode/src/terminal-manager.ts` (+6 / -0) — invariant comment at the `openArchitect` cache-key line
- `apps/vscode/src/__tests__/open-architect-not-live-main.test.ts` (+163 / -0) — new regression tests
- `apps/vscode/src/__tests__/extension-architect-commands.test.ts` (+8 / -1) — source guard updated for the delegation
- `codev/resources/lessons-learned.md` (+2 / -0) — name-as-address lesson + vitest-needs-built-dist lesson

## Commits

- `bb9314444` [PIR #1497] Refuse non-live 'main' in open-architect instead of substituting architects[0]
- `6bba74ee8` [PIR #1497] Regression tests for not-live-'main' resolution and injection routing
- `6d04d3dbf` [PIR #1497] Update builder thread: implement phase
- (+ this review commit)

## Test Results

- `pnpm check-types`: ✓ pass (after building `codev-types`, `codev-sdk`, `artifact-canvas` — see below)
- `pnpm test:unit` (vitest): ✓ pass — 856 tests, 73 files (10 new in this PR)
- `pnpm lint`: ✓ clean on changed files
- `node esbuild.js`: ✓ bundles
- Porch phase checks (`build`, `tests`): ✓ pass
- Manual verification (dev-approval gate): approved by the human on their machine — the live round-trip (real Tower pty + VSCode UI) is what the builder shell cannot run; see "How to Test Locally".

## Architecture Updates

No HOT arch change (`arch-critical.md` untouched): this is a localized bug fix inside `apps/vscode`, adding one small helper module without changing a module boundary, a system-shape invariant, or any cross-package contract. The one new file (`open-architect.ts`) is an extraction of a single command's resolution step for testability, not a new subsystem — below the threshold for an `arch.md` entry.

## Lessons Learned Updates

Routed two entries to COLD `codev/resources/lessons-learned.md` (HOT `lessons-critical.md` is at its 10-entry cap and neither entry is universal enough to justify displacing an existing hot lesson):

- **Architecture** — the `|| 'main'`-as-address pattern: a name that is only a caption can carry a convenience fallback safely, but the moment it keys an action (cache key, lookup, routing target) the fallback becomes a misdelivery bug. Records the two structural defenses (flow the occupant's own name; drop a fallback whose entire realized behaviour is the failure window) and the deliberate non-divergence from pir-1494 (both refuse). Sibling to the existing #1139 lesson on the same command.
- **Testing** — a vitest test that value-imports a workspace package needs that package's `dist` built first; companion to the #907 esbuild lesson, on the vitest side.

- **The design decision (refuse vs substitute).** The plan flagged that sibling lane pir-1494 answers the same "fall back to `main`?" question by refusing, and the architect asked for the choice here to be deliberate. This lane also **refuses** (removes the fallback), argued on: the fallback has no healthy execution path (only ever consulted when `main` is absent), caller intent (a request for `main` wants `main`), convention-convergence with pir-1494, and the issue thread's stated deck preference. **Option B** (open the substitute under its own name/label) was considered and is a one-line change in `resolveArchitectTarget` — because `openResolvedArchitect` already flows `target.name`, switching to B needs no other change. If a reviewer prefers B, it is a trivial, safe switch.
- **The injection-capture test imports the real `TerminalManager`.** It is the first vscode unit test to do so, and only resolves because CI's `test.yml` runs `pnpm build` before vitest (so `codev-types`/`codev-sdk` `dist` exist). Porch's own `tests` check passed, confirming the ordering holds in this harness. If you run `vitest` in isolation without a prior build, this file will fail to load — build the workspace first.
- **Structural safeguard vs policy.** `openResolvedArchitect` passing `target.name` (not the requested `targetName`) is what holds the invariant "no terminal is cached under `architect:<name>` while hosting a different architect" under *any* policy — worth a look to confirm it reads clearly.

## How to Test Locally

- **View diff**: VSCode sidebar → right-click builder `pir-1497` → **Review Diff**
- **Run dev**: VSCode sidebar → **Run Dev**, or `afx dev pir-1497`
- **What to verify** (needs a real workspace — a live `openArchitect` dials a Tower pty, so this cannot run headless):
  - Workspace with a named architect (e.g. `web`) plus `main`. Stop/restart so `main` is momentarily absent → invoke **Open Architect Terminal** for `main` (or the #1463 deck key set to Main). Expect: the "No 'main' architect found — is the workspace activated?" warning and **no terminal opens**. Then trigger **Reference Issue in Architect** on a backlog row → the reference text appears in **no terminal**, and specifically not in `web`'s. (Pre-fix: `web`'s terminal opened under the bare "Codev: Architect" label and received the text.)
  - With `main` live: the same two actions open `main`'s terminal and land the reference text there.

## Flaky Tests

None. One pre-existing, unrelated `check-types` failure (`src/markdown-preview/webview/main.ts` cannot resolve `@cluesmith/codev-artifact-canvas`) is a missing-build-artifact, not a flake and not caused by this diff — it clears once `artifact-canvas` is built, which CI does in a dedicated step. Untouched file; noted here rather than fixed (out of scope).


amrmelsayed:

**Risk tier: medium.** `+692/-10 across 10 files` overstates it — docs and state are ~421 lines (plan 253, thread 76, review 61, status 29, lessons 2), leaving **~281 lines of code across 5 files**. CMAP coverage (3-way, all APPROVE) already exceeds what medium requires.

### I verified the whole call path, not just the extracted module

The module being correct proves nothing if the caller ignores it, so I traced it end to end.

- **`resolveArchitectTarget`** is exact-match with no fallback. The `targetName === 'main' ? architects[0] : undefined` line is gone.
- **`openResolvedArchitect` flows `target.name`, never `targetName`.** This is the part I want to call out as good design rather than merely correct: it holds the invariant **structurally**. Even if a fallback were reintroduced in the resolver later, the name flowed onward would still be the actual occupant's, so the wrong terminal could not be cached under another's key. That is a defence against a future edit, not just against today's bug.
- **`extension.ts:1183` and `:1196`** both do `if (!resolvedName) { return; }` before injecting, and inject at `resolvedName`. So a refusal genuinely skips injection rather than falling through to `'main'`.
- **`terminal-manager.ts`** is comment-only, stating the invariant at the site where it binds (the cache key and the `injectArchitectText` lookup). This is the #1454 defect class handled preventively — a constraint the code cannot show, written where the next reader will trip over it.

### The framing worth keeping in the record

**The correct architecture was already here; the function was lying about its own contract.** `extension.ts:878-882` documents, from #1139, that the command *"returns the architect name that was actually opened"* so injection callers can target the same terminal. The old implementation returned `targetName` — the name that was *asked for*. Every downstream consumer was correct and was fed a wrong value.

So, as with #1494, this restores an existing design rather than introducing one. The bug was never the absence of a contract; it was an implementation that quietly violated the contract it advertised.

### On the design divergence

I told this builder that substituting under the alternate's own name was probably right here, and that #1494's *refuse* answer probably should not apply — and invited it to argue the opposite. **It argued the opposite and I accept it.** Its strongest point is that my recoverability distinction argues "refuse is cheap here" exactly as much as it argues "substitute is safe here", and supplies no positive reason to substitute. The two lanes now converge on *refuse*, deliberately and with the reasoning written down in the plan rather than left implicit.

### Two open items, neither blocking

1. **Test-harness fragility, flagged by the builder against itself.** The new test is the **first to import the real `TerminalManager`**, which resolves only because CI runs `pnpm build` before vitest. `terminal-manager.test.ts` deliberately avoids that import for this reason. The suite now carries a build-ordering dependency it did not have. Worth an explicit decision rather than silence — either accept it with a comment naming the dependency, or follow the existing file's avoidance pattern.
2. **The dev-approval demonstration needs a real host.** The builder stated plainly that `openArchitect` dials a Tower pty and cannot be driven headlessly, and named exactly what must be exercised: with `web` and `main` registered, stop `main`, invoke open-architect, confirm warning + no terminal, then confirm `referenceIssueInArchitect` lands nowhere rather than silently in `web`. That demonstration *is* the gate for this lane — the defect is which terminal receives text, and no assertion about a resolver reaches it.

**No blocking findings from me.** Merge decision rests with the repo owner.

amrmelsayed:
## Correction to open item 1 in my review above — I was wrong, and it is not a fragility

I wrote that "the suite now carries a build-ordering dependency it did not have." **That is false in both halves**, and I am retracting it rather than leaving it for a reviewer to act on.

- **CI already builds before tests.** `.github/workflows/test.yml` has dedicated steps building `sdk`, `core`, `types` and `artifact-canvas` ahead of any test run. The ordering is pre-existing infrastructure, not something this PR newly leans on.
- **20 existing tests already require built packages.** Twenty files under `apps/vscode/src/__tests__/` import `@cluesmith/codev-sdk` or `@cluesmith/codev-types`, all of which need `dist`. The suite has depended on the build for a long time.

**What is actually new is narrower, and it is an improvement rather than a risk.** `terminal-manager.test.ts` does not import `TerminalManager` at all — it reads the source file as text and asserts against it with regexes, noting at `:155` that the "heavy TerminalManager harness [is] avoided" by that file's convention. The new test imports the real class and captures real `sendText`, departing from that local convention.

That departure is the right direction. A regex-on-source guard asserts that code *looks* a certain way; it cannot catch a defect whose entire nature is **which terminal receives text**. The existing approach demonstrably would not have caught this bug. Importing the real class and observing the real call is the test that does.

So there is no open item here. The only thing worth a line is a comment in the new test noting that it intentionally diverges from `terminal-manager.test.ts`'s source-text convention and why, so the next person does not "fix" it back.

The builder's own statement was accurate and narrow — that its test is the first to import `TerminalManager` and relies on build-first ordering. The overstatement was mine, added on top of a correct claim without checking either half of it.

**Open item 2 (the dev-approval demonstration requiring a real host) stands unchanged.**

Issue #1501, opened 2026-08-18
Stream Deck: the Scroll dial can't scroll a spec/plan under review (canvas webview, not a text editor)


amrmelsayed:
## Problem

The Scroll dial rotates to scroll the focused editor's viewport (keyboard-free reading), but it does nothing when the reviewer is looking at a **spec or plan** — the exact long document where scrolling without the keyboard is most useful. Its two neighbours (Review: Files / Headings and Review: Changes / Blocks) can navigate a spec/plan; the Scroll dial cannot.

Discovered during #1498's dev-approval hardware testing. #1498 deliberately left rotation unchanged and scoped itself to `apps/streamdeck` only, so this is a separate follow-up rather than part of that lane.

## Root cause

The Scroll dial's rotation relays the `scroll` verb, which maps to VSCode's built-in **`editorScroll`** command (`apps/vscode/src/command-relay.ts:59`). `editorScroll` acts only on the **active text editor's** viewport.

A spec/plan under review is not a text editor — it opens in the **artifact-canvas**, a `WebviewPanel` (`apps/vscode/src/markdown-preview/preview-provider.ts`, the shared artifact-canvas surface). When that canvas is focused there is no active text editor, so `editorScroll` has nothing to act on and silently no-ops.

The review dials work on a spec/plan because they **switch channels by phase**: in canvas mode `ReviewNav` stops sending editor/diff verbs and drives the canvas over `sendCanvasCommand` (`heading-next`, `block-next`, etc.). The Scroll dial has only the single `editorScroll` channel, so it can never reach the canvas.

## Proposed direction

Give the canvas a viewport-scroll capability and phase-switch the Scroll dial's rotation the way the review dials already switch:

- Add a canvas command that scrolls the artifact-canvas viewport by N lines (up/down) — the canvas currently exposes heading/block/comment navigation and composer commands, but no raw viewport pan.
- In `ScrollNav.onDialRotate`, when the selected builder's phase is canvas mode (`reviewMode(...) === 'canvas'`), route rotation to that canvas command via `sendCanvasCommand` instead of `editorScroll`; keep `editorScroll` for diff/text-editor mode. This mirrors the existing `ReviewNav` mode split.
- Leave `revealCursor: false` semantics intact (viewport move only).

## What "fixed" means — the two gestures have different scopes

[...]

Issue #1462, opened 2026-08-15
porch: bookkeeping commits during the pr gate restart CI and invalidate the merge window (self-sustaining loop)


amrmelsayed:
Porch records state by **committing**, and CI keys off the branch head. So any protocol bookkeeping during the merge window invalidates the merge window: recording the pr gate pushes a commit, the push re-triggers every check, and the green board that justified the gate no longer describes the head being merged.

Observed end-to-end on pir-1070 / PR #1461 (2026-08-15), where it became a loop rather than a one-off delay:

```
02:36:20  chore(porch): 1070 pr gate-approved            <- gate recorded, CI restarts
02:36:23  chore(porch): 1070 protocol complete           <- second bookkeeping commit, CI restarts
02:37:11  [PIR #1070] Thread: pr gate approved,
          merge hit branch-protection wall               <- builder narrates the block, CI restarts
```

The architect measured all seven checks green, recorded the gate, and by the time the merge request reached the executing architect the board was pending against a head three commits newer. Twice more the same thing happened while the state was being reported.

**The loop is self-sustaining and is caused by good behaviour, not bad.** The builder narrates that the merge was blocked → the narration is a commit → the commit restarts CI → the merge defers → there is a fresh blocked state to narrate. The better a builder's thread hygiene, the tighter the loop. It only broke because two architects independently traced it in the same minute and told the builder to stop writing, including stopping `porch done`, which also commits.

**Two consequences worth separating:**

1. **A stated "CI is green" cannot survive the gate that follows it.** On a porch-driven lane those two facts can never be true of the same commit. Every merge-request presentation this week has carried a green claim that expired between writing and arrival — harmless when the executor re-verifies, and a silent hazard when they do not.
2. **The merge window is only the gap between the last bookkeeping commit and the next one**, and nothing today defines or protects that gap.

**Wanted — a rule rather than better manners.** Something in the protocol that makes the branch quiet from the pr gate until the merge lands: no thread writes, no porch state writes, no pushes of any kind, with `porch done --merged` explicitly deferred until after the merge commit exists. The builder-facing prompt should say so, because a well-behaved builder will otherwise keep its journal faithfully and re-trigger CI each time.

Worth considering alongside: whether porch's gate write and its phase-transition write need to be two commits (they were, here, 3 seconds apart), and whether bookkeeping-only commits could be marked so CI can skip them — though a rule is cheaper than a CI-config change and does not depend on every workflow honouring the marker.

[...]

PR #1504, opened 2026-08-18
[PIR #1498] Stream Deck: the Scroll dial narrates itself (mode, builder, empty state, canvas honesty)

### LINKED ISSUE #1498: Stream Deck: the Scroll dial does not narrate itself — no live mode, no builder, no empty state
## Problem

The Scroll dial is the only dial on the board that does not narrate itself, and it is the one whose press behaviour is mode-dependent.

`ScrollNav` (`apps/streamdeck/src/actions.ts`) calls `setTitle('Scroll')` once in `onWillAppear` and never touches the touchstrip again. It holds no `store.onChange` subscription and never calls `setFeedback`. Its three neighbours — Zoom, Review: Files / Headings, Review: Changes / Blocks — all render the house layout (line 1 the live semantic, line 2 the builder under review, a progress bar) and re-render on every store change. On the current board three dials describe themselves and the fourth shows one static word.

**It also under-reports what it does.** Since #1410 the press sends the mode-neutral `feedback-selection`, so it forwards to the builder immediately *or* stages into the review queue depending on the workspace's `codev.diffCodelensMode`. The dial whose press is mode-dependent is the only dial that never names its mode — which is precisely the surprise #1410's touchstrip work existed to eliminate ("Files · queue" vs "Files · send"). It also acts on the **selected builder**, and unlike every other builder-scoped surface it never shows which builder that is.

## Proposed change

Bring it onto the pattern the review dials already use — `setFeedback` plus a `store.onChange` subscription — rather than inventing anything:

- **Line 1**: the axis paired with the press meaning and the live delivery mode, in the established `A · B` form (e.g. `Scroll · queue` / `Scroll · send`), so a press is never a surprise.
- **Line 2**: the selected builder (`#issue title`), matching its neighbours, because the press acts on that builder.
- **Bar**: the selected builder's progress, as the other dials do.
- **Empty state**: `No builder` rather than an unchanging `Scroll` when there is nothing to act on — visibly inert, never live-looking and dead.

Rotation is unchanged (viewport scroll of the focused editor, `revealCursor: false`).

## Two decisions to make in the plan, not at implement time

1. **Does the press earn its place?** Under the board's own acceptance criterion — a key or gesture earns its place if the press completes an intention or opens a loop the deck can close — a selection press is weak: to have a selection your hand is already on the keyboard, where `Cmd/Ctrl+K B` lives. The counter-argument is that under queue mode the press is cheap and reversible (it stages rather than interrupts), which is a genuinely deck-shaped loop. Decide deliberately; if the press is dropped, line 1 becomes just the axis and the dial keeps its rotation.
2. **What does the bar show?** Builder progress is honest and consistent with the neighbours. Scroll position is *not* available — VS Code owns the viewport and the deck deliberately never mirrors it (`ReviewNav`'s comment records why the deck shows what the dial does rather than a counter). Do not invent a scroll-position bar.

## Scope

`apps/streamdeck` only. No wire, types, relay or VS Code change — every value needed is already on the overview the deck holds. Protocol: **PIR** — the payoff is legibility on the physical touchstrip, which wants confirming on hardware in both delivery modes and with no builder selected.

amrmelsayed:
# PIR Review: Scroll dial narrates itself (mode, builder, empty state, canvas honesty)

Fixes #1498

## Summary

The Stream Deck Scroll dial was the only touchstrip dial that did not narrate itself — a
static `setTitle('Scroll')`, no store subscription, no `setFeedback` — even though its
press is the one mode-dependent gesture on the board (forward-now vs stage-into-queue per
`codev.diffCodelensMode`). It now renders the house pattern: line 1 pairs the axis with a
qualifier (`Scroll · send` / `Scroll · queue` in diff/editor mode; `Scroll · editor only`
in canvas mode), line 2 names the selected builder (`No builder` when none), and there is
no progress bar (the dial's rotation and any builder-progress bar are unrelated axes). The
root fix has two halves — a manifest layout declaration and the render code — because the
dial was *declared* title-only, not merely un-rendered.

## Deviation From the Approved Plan

The approved plan specified only the `send` / `queue` line-1 qualifiers. The third
qualifier, **`Scroll · editor only` for canvas mode, was added after plan approval** — the
owner (Amr) ruled during the dev-approval hardware session that the dial silently doing
nothing on a spec/plan violated this lane's own "inert must be visibly inert" principle, so
the canvas-mode honesty was absorbed here rather than deferred. It landed *before*
dev-approval was granted, was flagged to the architect in the thread, and is unit-tested —
human-visible, not silent scope creep. Naming it explicitly here so the `pr` gate is clean.
(The separate *capability* gap — actually scrolling a canvas — stayed out of scope as
follow-up #1501.)

## Files Changed

Against `git merge-base main HEAD` (`9129ab81c`):

- `apps/streamdeck/src/actions.ts` (+79 / -) — `ScrollNav` rewritten to a store-subscribed
  dial; shared `selectedBuilderLine` helper (also adopted by `ReviewNav`); canvas-mode
  qualifier.
- `apps/streamdeck/src/__tests__/actions.test.ts` (+54 / -) — ScrollNav render coverage.
- `apps/streamdeck/com.cluesmith.codev.sdPlugin/layouts/title-value.json` (+24 / -0) — new
  title+value layout (no bar).
- `apps/streamdeck/com.cluesmith.codev.sdPlugin/layouts/label.json` (−15) — deleted (orphaned).
- `apps/streamdeck/com.cluesmith.codev.sdPlugin/manifest.json` (+4 / −) — scroll-nav layout
  repoint + Push description.
- `codev/reviews/1495-stream-deck-architect-action-k.md` (+21 / -0) — folded #1462 protocol
  note from the merged pir-1495 lane (separate commit, per architect).
- Plan, thread, porch status.

## Commits

`git log main..HEAD --oneline` (implementation, excluding porch chores):

- `8773d096e` [PIR #1498] Scroll dial: title+value touchscreen layout (no bar), retire label.json
- `46f881652` [PIR #1498] Scroll dial narrates itself: mode-aware line 1, selected builder line 2, inert empty state
- `69a149a2b` [PIR #1498] docs: land stranded #1462 protocol note into 1495 review (folded from pir-1495 lane)
- `3667cbc85` [PIR #1498] Scroll dial: 'Scroll · editor only' in canvas mode (both gestures inert), cross-ref #1501
- (plus plan/thread commits)

## Test Results

- `pnpm --filter @cluesmith/codev-streamdeck build`: ✓ pass
- `pnpm --filter @cluesmith/codev-streamdeck test`: ✓ pass (238 tests, 5 new/updated ScrollNav cases)
- `check-types` (tsc --noEmit): ✓ pass
- `validate` (manifest ↔ layouts): ✓ pass
- **Manual (dev-approval, hardware, VS Code extension 3.3.0 — all five checks attested):**
  1. Diff-phase builder → `Scroll · send` + `#id title`, no bar; rotation scrolls the editor.
  2. Forward mode: press forwards the selection to the builder immediately.
  3. Queue mode: line 1 flips to `Scroll · queue`; press stages into the review queue and the
     Send Fb badge increments (builder not interrupted).
  4. No builder selected: `No builder`; press inert.
  5. Canvas mode (spec/plan-phase builder): `Scroll · editor only`; both rotation and press inert.

  (All five ran under VS Code, whose 3.3.0 extension knows the `feedback-*` verbs, so the
  press-path checks 2–3 were genuinely exercised, not silently dropped.)

## Architecture Updates

No HOT (`arch-critical.md`) change: this is a single-dial legibility fix inside the already
established touchstrip pattern (subscription + `setFeedback` + `reviewMode`), not a new
system-shape invariant. No new module boundary, wire field, or state.

The durable architectural finding — the deck's two Scroll-dial gestures have **different
reach into VSCode** (rotation = any text editor via `editorScroll`; press = builder-diff
only via `feedback-selection`, whose `activeEntry()` requires a tracked diff entry) — is a
reference detail about the deck↔VSCode command surface, recorded below under Lessons
Learned (COLD) and in the finding section rather than promoted to an invariant.

## Lessons Learned Updates

Routed one COLD entry to `codev/resources/lessons-learned.md` (## Testing): the
default-fixture semantic trap (below). The two other findings reinforce an existing HOT
lesson ("Verify reviewer/plan claims against the actual file before acting") rather than
adding a new one — the press-also-dead result came precisely from reading the code against
a premise asserted as fact, so the existing rule already covers it.

### The four durable findings (the record for this lane)

**1. The two gestures have different scopes — this outlives the wording.** Rotation is
any-text-editor (`editorScroll`); the press is builder-**diff**-only (`feedback-selection`
→ `feedbackSelection` → `selectionAnchor()` → `activeEntry()`, which returns undefined
unless the focused document is a tracked builder-diff entry — `feedback.ts:48-90`). No
single mode label can describe both gestures in every state, which is *why* canvas mode
shows `editor only` rather than a delivery-mode qualifier: on a spec/plan (an
artifact-canvas webview) there is neither an active text editor nor a diff entry, so both
gestures are inert.

**2. The root cause was a declaration mismatch, not missing render code.** Five dials
declared `layouts/dial.json`; this one declared `layouts/label.json`, so it was built
title-only *by manifest* and the code merely agreed with it. That is why the fix has a
manifest half (new `title-value.json`, retire `label.json`) as well as a code half.

**3. A green test that was semantically wrong (the fixture trap).** The original
`Scroll · send`/`queue` assertions ran against `pir-1`, the fixture default — which is
blocked at plan-approval and therefore **canvas** mode. They passed while asserting a
delivery-mode label for a builder that should never show one. The bug was invisible until
canvas mode was given meaning, which flipped that builder's expected label to
`editor only`. The next person adding a mode-dependent assertion will reach for the same
default fixture; pick the builder whose phase matches the mode under test (`pir-2` is diff).

**4. #1501 scope correction.** A canvas viewport-scroll capability (#1501) would revive
**rotation** only; the press stays diff-bound by design. So `Scroll · editor only` will
*narrow* (rotation starts working) rather than *disappear* when #1501 lands — it does not
simply collapse back to `send`/`queue`.

- `ScrollNav.renderTo` (`actions.ts`): the three-way qualifier (`editor only` / `queue` /
  `send`) and its render-site comment. The press is deliberately **not** gated on
  `reviewMode` — it no-ops server-side in canvas mode already, and gating it deck-side would
  add state for no behavioural gain.
- The shared `selectedBuilderLine` helper: `ReviewNav` now calls it too. Its output is
  byte-identical to the inlined logic it replaced, which is what keeps `ReviewNav`'s existing
  string assertions green — worth confirming that equivalence held.
- The fixture trap (finding 3): the diff/queue tests were moved onto `pir-2`.

## How to Test Locally

For reviewers pulling the branch:

- **View diff**: VSCode sidebar → right-click builder `pir-1498` → **Review Diff**.
- **Run on hardware**: build the sdk then the plugin, `streamdeck link` the *worktree's*
  `com.cluesmith.codev.sdPlugin`, `streamdeck restart com.cluesmith.codev`, confirm with
  `streamdeck list`. A manifest/layout change may need a full quit+reopen of the Elgato app.
- **What to verify**: the five hardware checks above — both delivery modes, the no-builder
  state, and a spec/plan-phase builder for `Scroll · editor only`.

## Flaky Tests

None.


amrmelsayed:
**Integration review (streamdeck architect) — APPROVE.** Deck-only (five files under `apps/streamdeck`), reviewed continuously through this lane's plan, its owner-directed scope addition, and a stop-and-report that overturned a premise I had asserted.

**Verified at source rather than from the report:** `ScrollNav` now holds a `store.onChange` subscription and renders via `setFeedback`; line 1 reads the *live* `feedbackMode()` rather than a static string; `reviewMode(selectedBuilder())` drives the canvas branch; `selectedBuilderLine` is a genuine extraction that `ReviewNav` now shares rather than a copied helper; `label.json` is deleted and `title-value.json` added with the dial layout minus the bar.

**The root cause is worth restating** because it changes what this lane was: the dial was *declared* title-only. Five dials pointed at `dial.json`, this one at `label.json`, which has no value or bar field. Code and manifest agreed with each other and disagreed with every sibling, so this was never a missing render function — it had a manifest half nobody had found.

**On Claude's none-mode note — agreed with the builder's judgement, and here is why the two cases genuinely differ.** In canvas mode the strip showed a delivery-mode qualifier *beside a named builder*, which implies "this press will send that builder's selection" with nothing anywhere to correct it — false by implication. In none-mode, the qualifier states the workspace's actual delivery mode (true), and line 2 says `No builder` explicitly, so the absent target is stated rather than inferred. I also confirmed the builder's third point at source: `onDialRotate` relays `scroll` with only the workspace path, so **rotation genuinely works with nothing selected** — the dial is half-live there, and the `editor only` wording would be false.

There is still a small residual: a qualifier describing a press that cannot currently fire. The clean fix is to suppress the qualifier when no builder is selected (line 1 becomes a bare `Scroll`), mirroring exactly what the canvas branch does and for the same reason. That is two lines.

**Not folding it into this PR**, deliberately. Doing so would move the head after the gate opened — restarting all seven checks, which is the #1462 tax this lane's own folded note documents — and would invalidate the human's hardware attestation of the no-builder state, which is precisely the state that would change. Neither cost is worth two lines. It should ride the next lane touching `actions.ts`, in the shape #1454 established.

**Attestation, stated plainly**: all five hardware checks were exercised by the owner in VS Code (whose extension knows the `feedback-*` verbs) — diff-phase strip and rotation, forward-mode press, queue-mode label flip with the badge incrementing, the inert `No builder` state, and canvas-mode `Scroll · editor only` with both gestures dead. Checks 1 and 3 together are what justify the lane: a mode label proven to change with the mode.

Consultation: full 3-way, all APPROVE. The plan-deviation note was added *before* the gate opened rather than after, which is the correct handling of exactly the trap this lane documents.

PR #1507, opened 2026-08-18
Fix #1502: bound test.yml jobs with timeout-minutes

### LINKED ISSUE #1502: CI: Artifact-Canvas Browser Tests hangs indefinitely (intermittent) — blocks every merge and leaves main's Tests run unconcluded
## Problem

The `Artifact-Canvas Browser Tests` job in the `Tests` workflow intermittently hangs indefinitely
instead of completing or failing. Every sibling job in the same run finishes normally, so the run
never reaches a conclusion and the PR's check set never becomes complete.

Observed 2026-08-18, three concurrent instances on unrelated branches:

| Run | Branch | Job started | State when observed |
|---|---|---|---|
| `32085216485` | **main** (merge of #1493) | 00:37:25Z | still `in_progress` at ~03:03Z (**~2.5 hours**) |
| `32093117702` | `builder/pir-1497` | 02:49:07Z | still `in_progress` after ~14 min |
| `32091129032`-era | `builder/air-1474` | 02:47:12Z | `in_progress` |

Baseline for the same job is **~1m09s** (e.g. the run on PR #1421). So these are hangs, not slow
runs.

In every hung run, the other four jobs (`Unit Tests`, `CLI Integration Tests`, `Tower Integration
Tests`, `Package Install Verification`) completed `success` within ~1–2.5 minutes.

Not a straightforward concurrency-contention story: two near-simultaneous `air-1474` runs at
02:46:36 and 02:46:40 both completed successfully, while runs starting ~30 seconds later hung.

## Impact

1. **It blocks merges.** Branch protection plus the standing admin-bypass envelope require the *full*
   check set green on a stationary head. A permanently-pending job means the set is never complete,
   so a correctly-verified merge is impossible without either waiting out a hang that may never end
   or lowering the bar to "six of seven", which defeats the envelope.
2. **main has been silently incomplete for hours.** Run `32085216485` has never concluded, so main's
   `Tests` workflow has no result for the #1493 merge. A glance at the branch shows later green runs
   and reads as healthy.
3. **It is release-blocking in effect.** A release cut cannot be verified green while a required check
   hangs intermittently.

## What to investigate

- Whether the browser/harness process fails to exit (a held handle, an un-awaited server, a browser
  instance not torn down) rather than the tests themselves hanging.
- Whether the job has *any* timeout. A hung job that runs for 2.5 hours suggests either no
  `timeout-minutes` on the job or one set far too high. A tight timeout would convert this from a
  silent stall into a visible failure, which is the minimum acceptable behaviour even before the root
  cause is found.
- Whether it correlates with runner cold-start or browser download.

## Acceptance criteria

- [ ] Root cause identified for the non-termination.
- [ ] The job carries a `timeout-minutes` bound proportionate to its ~1m09s baseline, so a stall
      fails visibly instead of pending forever.
- [ ] No indefinite `in_progress` jobs observed across a sustained run of the suite.

amrmelsayed:
## Summary

The `Artifact-Canvas Browser Tests` job in the `Tests` workflow intermittently hung for hours
instead of concluding, leaving PR check sets permanently incomplete and blocking merges. This PR
**removes the stall at its source** by dropping `--with-deps` from the Playwright install, and
**keeps a per-job `timeout-minutes` backstop** so any future stall in any job fails fast instead
of pending for hours.

Fixes #1502

## Root Cause

Confirmed from the logs of two cancelled hung runs (`32093117702` on `builder/pir-1497`,
`32085216485` on `main`). Both stalled at the **identical point**: the `apt-get update` that
`pnpm exec playwright install --with-deps chromium` runs to install Chromium's system libraries.

```
Get:7 https://archive.ubuntu.com/ubuntu noble-security InRelease [126 kB]
<... then nothing for ~1h and ~3h respectively, until cancellation ...>
```

The runner's fast `azure.archive.ubuntu.com` mirror is `Ign`-ored (unreachable), forcing fallback
to public `archive.ubuntu.com`, whose index fetch then stalls indefinitely. apt has no default
network timeout on that fetch, and no job in `test.yml` had a `timeout-minutes`, so GitHub's
6-hour default let the stall pend for hours. It is intermittent because it depends on that
transient mirror/network condition. The tests never start, so the `playwright.config.ts`
webServer / `--strictPort 5199` hypothesis is ruled out — the webServer is never invoked.

## Fix

**1. Remove the cause — drop `--with-deps` (`test.yml`, the install step).**
The stalling apt phase exists only to install Chromium's system libraries, and those libraries are
already present on the `ubuntu-latest` image. Proof lives in our own repo: the sibling
`dashboard-e2e.yml` workflow installs Chromium with the same command **minus** `--with-deps`
(`playwright install chromium`) on the same runner image, then launches Chromium for its
Playwright tests — and it has succeeded on `main` on every scheduled run from 2026-08-10 through
2026-08-17 (8 consecutive). So `--with-deps` buys nothing here except the apt phase that stalls.
Dropping it removes the stall's source rather than bounding it.

**2. Keep the backstop — `timeout-minutes` on every job (`test.yml`).**
`canvas-browser` and `unit` get 15 (healthy ~2–4 min); `integration`, `cli`, `package` get 10
(healthy 1–3 min). This is **not** traded away against the fix above. #1502's real lesson is that
an *unbounded* job conceals its own failure — it pends silently instead of failing — so every job
stays bounded regardless of this specific root cause. If any job ever stalls again for any reason,
it now fails visibly in minutes rather than pending for GitHub's 6-hour default.

### Considered and not done

- **`--retries` on `playwright test`** (as `dashboard-e2e` uses): not added. Retries re-run failed
  *tests*, not the *install* step, so they would not have touched #1502's stall. And the canvas
  browser suite asserts deterministic CSS-fragmentation invariants (protection rects, tall-block
  caps, per-fragment prose rects) — a retry there would mask a real layout regression rather than
  absorb infra flake. That is a different test character from `dashboard-e2e`'s async Tower/`gh`
  flows, where retries are justified.
- **Caching `~/.cache/ms-playwright`**: would not help — `--with-deps` ran `apt-get` regardless of
  the binary cache, so it never skipped the stalling phase. Moot now that `--with-deps` is gone.

## Test Plan

- [x] Regression test: `packages/codev/src/__tests__/bugfix-1502-test-workflow-timeouts.test.ts`
      parses `test.yml` and pins **both** halves of the fix: every job declares a bounded numeric
      `timeout-minutes` in `(0, 360)` (parameterized over the job list, so future jobs are covered
      too), and the canvas-browser install step does **not** pass `--with-deps`. Verified it fails
      both when a timeout is removed and when `--with-deps` is reintroduced; passes on the fix.
- [x] Build passes (`porch check`)
- [x] All tests pass (`porch check`)
- [x] This PR's `Artifact-Canvas Browser Tests` job installed Chromium **without** `--with-deps`
      and the browser launched — the canvas suite ran green in **45s** (vs ~4 min with the apt
      phase). Confirms the runner image carries every library canvas needs.

## Caveat

`dashboard-e2e` exercises the dashboard; this job exercises canvas rendering. If canvas needs a
system library the dashboard does not, dropping `--with-deps` surfaces it as a **browser launch
failure** in this PR's own `canvas-browser` job — cheap, immediate verification. If that happens,
the fix is to restore `--with-deps` and rely on the timeout backstop alone; that result is worth
knowing and is stated here honestly rather than claimed away. As with any intermittent failure, a
single green run does not *prove* the apt stall can never recur through some other path — which is
exactly why the timeout backstop stays.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,6 +9,9 @@ jobs:
   unit:
     name: Unit Tests
     runs-on: ubuntu-latest
+    # Healthy runs finish in ~2-3 min; bound well above that so a stall fails
+    # visibly instead of pending for GitHub's 6-hour default (see #1502).
+    timeout-minutes: 15
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -115,6 +118,12 @@ jobs:
   canvas-browser:
     name: Artifact-Canvas Browser Tests
     runs-on: ubuntu-latest
+    # Root cause of #1502: `playwright install --with-deps` runs `apt-get update`,
+    # whose index fetch to archive.ubuntu.com intermittently stalls indefinitely
+    # (the fast azure.archive.ubuntu.com mirror is Ign-ored, and apt has no network
+    # timeout on the fallback fetch). Healthy runs finish in ~4 min; this bound turns
+    # a multi-hour silent stall into a fast, visible failure that a rerun clears.
+    timeout-minutes: 15
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -144,6 +153,8 @@ jobs:
   integration:
     name: Tower Integration Tests
     runs-on: ubuntu-latest
+    # Healthy runs finish in ~2 min; bound above that so a stall fails visibly (#1502).
+    timeout-minutes: 10
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -179,6 +190,8 @@ jobs:
   cli:
     name: CLI Integration Tests
     runs-on: ubuntu-latest
+    # Healthy runs finish in ~1 min; bound above that so a stall fails visibly (#1502).
+    timeout-minutes: 10
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
@@ -214,6 +227,8 @@ jobs:
   package:
     name: Package Install Verification
     runs-on: ubuntu-latest
+    # Healthy runs finish in ~2 min; bound above that so a stall fails visibly (#1502).
+    timeout-minutes: 10
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -118,11 +118,9 @@ jobs:
   canvas-browser:
     name: Artifact-Canvas Browser Tests
     runs-on: ubuntu-latest
-    # Root cause of #1502: `playwright install --with-deps` runs `apt-get update`,
-    # whose index fetch to archive.ubuntu.com intermittently stalls indefinitely
-    # (the fast azure.archive.ubuntu.com mirror is Ign-ored, and apt has no network
-    # timeout on the fallback fetch). Healthy runs finish in ~4 min; this bound turns
-    # a multi-hour silent stall into a fast, visible failure that a rerun clears.
+    # Backstop for #1502. The stall's root cause is removed below (see the install
+    # step), but every job stays bounded regardless: an unbounded job conceals its
+    # own failure, which is #1502's real lesson. Healthy runs finish in ~4 min.
     timeout-minutes: 15
     steps:
       - name: Checkout repository
@@ -140,9 +138,16 @@ jobs:
       - name: Install dependencies
         run: pnpm install
 
+      # Root cause of #1502: `--with-deps` runs `apt-get update` to install Chromium's
+      # system libraries, and that apt index fetch intermittently stalls indefinitely
+      # (the fast azure.archive.ubuntu.com mirror is Ign-ored, and apt has no network
+      # timeout on the fallback to archive.ubuntu.com). Those libraries are already
+      # present on the ubuntu-latest image: the sibling dashboard-e2e workflow installs
+      # Chromium the same way without `--with-deps` and launches it successfully on
+      # every scheduled run. Dropping the flag removes the stalling apt phase entirely.
       - name: Install Playwright Chromium
         working-directory: packages/artifact-canvas
-        run: pnpm exec playwright install --with-deps chromium
+        run: pnpm exec playwright install chromium
 
       # Real-browser fragmentation regression suite (spec 1380): jsdom cannot express CSS
       # multi-column fragmentation, so the spike-derived invariants run against Chromium here.
--- a/codev/projects/bugfix-1502-ci-artifact-canvas-browser-tes/status.yaml
+++ b/codev/projects/bugfix-1502-ci-artifact-canvas-browser-tes/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1502
+title: ci-artifact-canvas-browser-tes
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-08-18T03:53:32.119Z'
+updated_at: '2026-08-18T03:53:32.120Z'
--- a/codev/projects/bugfix-1502-ci-artifact-canvas-browser-tes/status.yaml
+++ b/codev/projects/bugfix-1502-ci-artifact-canvas-browser-tes/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-08-18T04:04:03.885Z'
 iteration: 1
 build_complete: false
 history: []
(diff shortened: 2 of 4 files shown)


amrmelsayed:
## Architect Integration Review — APPROVE

Low-risk tier (4 files, +163/-0, CI configuration plus a regression test), so reviewed directly rather than via a consultation panel. I verified the substantive claims against the artifacts rather than the summary.

**Root cause is genuinely log-confirmed, and the evidence is specific.** Both hung runs stalled at the identical point inside `Install Playwright Chromium`: the `apt-get update` that `--with-deps` performs, fetching `archive.ubuntu.com … InRelease`, then silence — ~64 minutes on `32093117702` and ~3 hours on `32085216485` before the authorized cancels. The signature (the fast `azure.archive.ubuntu.com` mirror `Ign`-ored, forcing fallback to the public archive, with no apt network timeout on that fetch) explains both the stall and its intermittence. That is a real diagnosis, not a plausible story fitted to a symptom.

**The claim scope is correct.** "The apt index-fetch phase of `--with-deps` stalls" is the narrowest statement the logs support, and the PR says it rather than the more satisfying "CI hang fixed". The workflow comment is likewise honest that the bound converts a silent multi-hour stall into a fast visible failure *that a rerun clears* — it does not claim to prevent the stall.

**One rejection of mine that I want on the record as correct.** The issue listed caching `~/.cache/ms-playwright` as a candidate fix. This PR declines it with the right reason: that cache skips the browser-binary download, not the apt phase, so it would not have touched this stall at all. Good catch — implementing it would have produced a change that looked like a fix and prevented nothing.

**Bounds are sized off observed healthy durations, not the bad figure.** An earlier note of mine quoted ~1m09s as the baseline for the canvas job; that was wrong, and a timeout sized from it would have fired on healthy runs. This PR uses ~4 min observed with a 15-minute bound (~3.7x headroom), and 10 minutes for the ~1–2 minute jobs. Correct input, correct headroom.

**Bounding every job rather than only the implicated one is the right call**, and I am glad it was argued rather than done silently. A workflow with zero timeouts is one runner hiccup away from the same failure in a different job.

**The regression test earns its place**: it parses the workflow and asserts every job carries a bound that is greater than zero and less than GitHub's 360-minute default, so a future job added without a timeout fails the suite rather than inheriting the six-hour silence.

### Not blocking

- Tonight's stall rate was 2 of 3 runs, which if it were the steady state would make a 15-minute bound plus a manual rerun expensive. I read tonight as an upstream mirror incident rather than the base rate, since this job has passed routinely for months. Worth revisiting with a step-level retry only if the rate persists — not worth pre-building now.
- The four other workflow files are unbounded. Filing separately rather than growing this PR.

Merge authority is the owner's at the pr gate; this review is a recommendation.

amrmelsayed:
## Architect review — addendum after the `--with-deps` removal

The PR changed materially since my first review, so this supersedes it. Re-verified against the artifacts.

**The removal is confirmed in the diff** (`- pnpm exec playwright install --with-deps chromium` → `+ pnpm exec playwright install chromium`), and the verification is stronger than the summary claimed.

On run `32099116626` (head `913035b37`), `Artifact-Canvas Browser Tests` completed **`success` in 40 seconds** — started 04:26:25Z, completed 04:27:05Z. Against a with-`--with-deps` healthy time of ~4 minutes, and against two runs earlier tonight that never finished at all.

That reframes the change. It does not only remove a hang: **the apt phase was roughly 85% of this job's runtime and 100% of its failure surface.** The job is now ~6x faster and has no network-fetch step to stall on.

Supporting evidence also checks out: `dashboard-e2e.yml` installs Chromium the same way and is green on **10 of its last 10 runs**, so this is the repo's established pattern rather than a new bet.

**Keeping the timeouts alongside the removal is the right shape**, and I want the reasoning recorded because a future reader may see them as now-redundant. They are not. Removing `--with-deps` fixes *this* stall; the bounds are what stop the *next* unbounded job from concealing its own failure for three hours, which is the durable lesson of #1502. One is a fix, the other is a class of protection.

**The `--retries` argument is correct and I am glad it was argued rather than copied.** Playwright's `--retries` retries *tests*; the failure here was in the *install step*, which no test-level retry can reach. Copying the sibling workflow's flag would have added a knob that could not have helped this failure, on a deterministic layout suite that should not need it.

**The assertion pinning `--with-deps` absence** is the right regression guard — it makes a well-meaning future re-addition of the flag fail the suite rather than silently reintroduce a multi-hour stall.

### Outstanding before merge

The head has moved to `dd1a7e952` (main merged in to validate the integrated state, which was the right call). All seven checks are pending on that head as I write. My recommendation is APPROVE **conditional on a full green against that stationary head** — not on the green I verified at `913035b37`, which belongs to a superseded commit.

Merge authority is the owner's at the pr gate.

Issue #1401, opened 2026-08-11
artifact-canvas: remote command channel for review navigation (Tower relay + sdk route)


amrmelsayed:
## Context

Split out of #1400 per sequencing agreement between main and the streamdeck architect: the bridge has no deck dependency, lands first, and is coordinated by main (it touches packages/artifact-canvas, Tower, and the sdk). #1400 remains the deck-actions issue and depends on this one.

The canvas's keyboard-first review work (#1237/PR #1344) and horizontal reading mode (#1380/PR #1398) defined a semantic action vocabulary: focus next/previous block, next/previous commented block, next/previous heading, document start/end, page column forward/back, open composer on focused block, toggle reading mode.

## Requirements

1. **Canvas command channel.** A way for Tower to deliver navigation/review commands to an open canvas page, most plausibly over the live-update channel the page already holds. **The command set mirrors the existing keyboard vocabulary exactly; no parallel vocabulary.** Composer open/submit/cancel are in scope; text entry is not (comment bodies are typed on the keyboard).
2. **Tower surface.** An endpoint or message route that accepts one command for a given file/workspace and relays it to the target canvas view.
3. **sdk route.** A client call for that surface. The streamdeck architect is a stakeholder on the call's surface shape as controller-subpath owner (same review arrangement as #1189) and reviews it at design time, before implementation.
4. **Target rule decided at spec time.** The no-canvas-open and multiple-canvases-open cases get an explicit, specified answer (defined target rule or explicit error; never a silent ambiguous no-op). This is a spec-phase decision, not an implementation-time one.

## Non-goals

- Deck actions themselves (#1400).
- Any change to the canvas's in-page keyboard behavior.

## Notes

The bridge is deliberately generic: any remote driver of the viewer (not only the deck) gets it for free.

amrmelsayed:
## sdk-surface review (streamdeck architect, controller-subpath stakeholder per #1189)

Reviewed spec draft at spir-1401 worktree, Desired State §3 plus the items deferred to this review. **Verdict: APPROVE the shape, with one requested addition (item 3) and one tightening (item 4).** Nothing here blocks the builder's consultation pass; the two changes are small and spec-level.

### 1. Generic-relay exposure: NO. Targeted route only.
[...]

PR #1510, opened 2026-08-18
Dashboard: make the held-mail counter clickable, showing the held messages (from → to)

### LINKED ISSUE #1450: Dashboard: make the held-mail counter clickable, showing the held messages (from → to)
## What

The dashboard shows a held-mail count in the corner (e.g. "2 held" — the mailbox-first counter from Spec 1313), but it's inert text. Finding out *what* is held currently requires dropping to the CLI (`afx inbox -w <workspace>`).

Make the counter:

1. **Look clickable** — underline it (or equivalent affordance) so it reads as interactive rather than as a static status label.
2. **Show the held messages on click** — a popover/panel listing each held message with **at least from → to** (e.g. `architect → cost`). Age, hold reason, and created-at would be nice to have — the CLI's `afx inbox` table (ID / AGE / REASON / FROM → TO / WORKSPACE) is a good reference for the fields that exist.

## Why

"2 held" with no way to inspect it sends the user hunting. The immediate question when you see the counter is "held from whom, to whom, and does it matter?" — the first two are cheap to surface inline.

## Notes

- Data is already available server-side (the same store `afx inbox` reads); this is a UI affordance + a small read endpoint/reuse of an existing one.
- Showing the full body inline is optional; from → to plus age/reason covers the triage need, and `afx inbox show <id>` remains the deep-dive path.

mohidmakhdoomi:
# PIR Review: Clickable held-mail counter with a held-messages popover

Fixes #1450

## Summary

The dashboard header's held-mail counter ("2 held") was inert text, so learning *what* was held
meant dropping to `afx inbox` in a terminal. It is now a disclosure button — dotted underline,
`aria-expanded`/`aria-controls` — that opens a panel listing each held message as `from → to`
with its age and why-held reason. Server-side this reuses the existing `handleInboxList` behind a
new workspace-scoped `GET /api/inbox` branch; no new handler and no new projection, so the
metadata-only redaction rule and CLI-only dismissal (Spec 1313 decision 8) are untouched.

The one genuinely subtle part is that **the badge count and the list disagree by design** — see
*Things to Look At*.

## Files Changed

- `apps/web/src/components/HeldCountBadge.tsx` (+219 / -…) — span → disclosure button, grouped popover, generation-guarded lazy fetch
- `apps/web/src/index.css` (+125 / -0) — underline affordance, popover panel, z-index tier
- `apps/web/src/lib/heldMail.ts` (+48 / -0) — new; `formatHeldAge` / `formatHeldDuration` / `isScheduled`
- `apps/web/src/lib/api.ts` (+19 / -0) — `fetchInbox`
- `apps/web/src/components/App.tsx` (+8 / -3) — wire `loadMessages`
- `packages/codev/src/agent-farm/servers/tower-routes.ts` (+30 / -3) — `workspaceOverride` param + workspace-scoped branch
- `packages/types/src/api.ts` (+46 / -0) — `HeldMessage`
- `packages/types/src/index.ts` (+1 / -0) — export it
- `apps/web/__tests__/HeldCountBadge.test.tsx` — 32 cases total: 5 originals kept unchanged, **27 new**
- `apps/web/__tests__/heldMail.test.ts` — new, **10 cases**
- `packages/codev/src/agent-farm/__tests__/inbox-routes.test.ts` — **12 new** route cases (14 → 26 in the file)
- `packages/codev/scripts/issue-1450-dashboard-evidence.mts` (+370 / -0) — new; real-browser evidence harness
- `codev/plans/1450-dashboard-make-the-held-mail-c.md`, `codev/state/pir-1450_thread.md`, `codev/resources/arch.md`, `codev/resources/lessons-learned.md` — artifacts and governance

## Commits

- `b6c15774e` [PIR #1450] Plan draft
- `934e9354e` [PIR #1450] Plan revised: Held/Scheduled split, a11y disclosure pattern, isolated-Tower Playwright
- `5a782c0b7` [PIR #1450] Plan: use AF_TEST_DB isolation seam for the evidence Tower (not a HOME redirect)
- `d70e3503f` [PIR #1450] feat: clickable held-mail counter with a held-messages popover
- `7483bcce9` [PIR #1450] Review + retrospective
- `2cb2b3623` [PIR #1450] Address 3-way consult + architect integration review

## Test Results

- `pnpm build`: ✓ pass
- `pnpm test` (`@cluesmith/codev`): ✓ 4917 passed, 48 skipped, **0 failures**
- `apps/web` (`pnpm --filter @cluesmith/codev-web test`): ✓ 33 files — **37 new web tests** (27 component + 10 formatter)
  - Note: the root `test` script runs only the `codev` package; web tests need the filtered command.
- **49 new tests total** across the three files (27 + 10 + 12).
- **Manual, real browser**: `packages/codev/scripts/issue-1450-dashboard-evidence.mts` — 23/23 checks
  in headless Chromium against an isolated Tower (worktree build, port 14700, `NODE_ENV=test` +
  `AF_TEST_DB`). Real workspace, real shellper PTYs painted with an occupied composer, real
  `POST /api/send` held by the render gate, real built SPA. Asserts the underline affordance,
  `aria-expanded` toggling, `from → to` rows, the Held/Scheduled grouping,
  `GET .../api/inbox` returning **200**, Held-group length === badge count, popover stacking over
  a live terminal, and Escape-closes-with-focus-return.
- **Human review at the `dev-approval` gate**: the reviewer exercised an interactive instance of
  that same environment personally (2 held + 1 scheduled fixture) and approved.

## Architecture Updates

Routed **COLD only** (`codev/resources/arch.md`). Neither hot file was touched: both are at their
10-entry cap, and nothing here is a cross-cutting invariant worth displacing an existing entry —
these are subsystem facts about Tower's routing and the mailbox, which is what the cold archive is
for.

Two additions:

1. **Agent Farm Internals → "Two route tables, and why a 'working' endpoint can still 404 for the
   dashboard."** `tower-routes.ts` dispatches through two independent tables — the Tower-level
   `ROUTES` map (`/api/<thing>`, used by the CLI) and `handleWorkspaceRoutes`
   (`/workspace/<b64>/api/<thing>`, used by the dashboard, whose `getApiBase()` returns `'./'`).
   Registering in one does not register in the other, which is exactly why this issue existed:
   `GET /api/inbox` had backed `afx inbox` since Spec 1313 while `./api/inbox` 404'd for the
   dashboard. Also records why workspace-scoped handlers take a `workspaceOverride` that **wins
   over** `?workspace=`, and that the exact-vs-prefix match is a security boundary.
2. **Mailbox section** — the count/list asymmetry, written as "do not 'fix' it" (details below),
   including the corollary that a scheduled-only state is invisible to the badge by design.

## Lessons Learned Updates

Routed **COLD only** (`codev/resources/lessons-learned.md` → Testing). Both entries are about
browser-test technique — useful, but not the always-injected kind, and the hot file is at cap.

1. **A browser assertion can pass while proving nothing — assert its precondition in the same
   test.** My z-index check confirmed the popover was on top at its centre point and was green
   and worthless: `querySelector('.xterm')` had returned the *left* pane's terminal, which can
   never overlap a top-right panel, so the actual hazard (xterm's WebGL canvas painting over an
   unlayered element) was never exercised. The fix asserts the setup — "a terminal genuinely
   overlaps this rect" — before asserting the property. Same shape as #1401's guard lesson.
2. **Two Playwright mechanics specific to this dashboard**: `waitUntil: 'networkidle'` can never
   fire (the SSE stream at `/api/events` stays open for the page's lifetime), and a lazily-fetched
   panel must be waited on by its *loaded* content or assertions read the "Loading…" state. Plus:
   assert the **status code** of the endpoint under test, because an auth failure renders as a
   tidy error state that looks like a working UI.

**1. The Held/Scheduled split — the part that took two attempts to get right.**
My first plan asserted that pre-due `--delay` rows "already inflate `heldCount`". That was
backwards, and the architect's review caught it. Verified against `db/mailbox.ts`:

- `heldSummaryForWorkspace` (`:215-227`) — the badge count — filters
  `not_before IS NULL OR not_before <= now`.
- `listHeld` (`:113-124`) — behind `GET /api/inbox` — has **no** such filter.

So `heldCount <= inbox.length`, always, deliberately: a scheduled send is "scheduled, not stuck"
and must not raise an attention indicator. A naive popover would say "2 held" and list 3 rows. The
panel therefore **groups**: `Held (N)` where N is exactly the badge count, above a secondary
`Scheduled (M)` with a countdown and explanatory copy. Groups render only when non-empty, so the
ordinary case looks like one plain list. Pinned by a unit test (1 due + 1 pre-due at `count={1}`)
and re-verified live in the browser run, which reproduced exactly that fixture.

**2. Accepted edge case, not an oversight.** With 0 due and 1 scheduled row, `heldCount` is 0, the
badge does not render, and that row is unreachable from the dashboard. That is the existing
contract — the badge is an *attention* indicator — and `afx inbox` remains the surface that sees
it. Surfacing it would mean rendering a badge whose count is 0. Documented on the component, on
the `HeldMessage` type, and in arch.md; changing it is a change to what the badge counts.

**3. The exact-match on `'inbox'` is load-bearing.** `inbox/:id` returns the message **body** and
`inbox/:id/dismiss` mutates. Both must stay unreachable under the workspace prefix for the
"metadata-only, read-only" claim to hold. Three tests pin that (both 404, and the row survives a
POST) rather than leaving it to inspection.

**4. `formatHeldAge`, not `formatDuration`.** `apps/web/src/lib/open-files-shells-utils.ts:2-10`
already exports a `formatDuration` with different semantics (minute granularity, `<1m` floor) and
existing callers. The new helper is second-granularity to match `afx inbox`, so it is separately
named rather than merged — two same-named formatters with different output in one `lib/` is a
trap. It is a deliberate ~10-line port, not an import: the web app must not cross the
server/client isolation boundary (#1189), and `codev-types` is a types-only devDependency.

**5. Fetch lifecycle.** Loads on open and on `count` change while open (the count is SSE-driven, so
a change means the mailbox moved). Each load carries a generation counter; stale responses are
discarded — React 19 would not warn about the open→close→open race. And while the panel is open
the component stays mounted even at `count <= 0`, because `useOverview` polls every 2.5s and the
last row being delivered would otherwise unmount the button and drop focus to `<body>`. The
closed-at-zero contract is unchanged, so the original zero-state tests pass untouched.

**6. Not a dialog.** Deliberately the WAI-ARIA *disclosure* pattern (`aria-expanded` +
`aria-controls` + a real `<ul>`), not `role="dialog"` — a dialog role without moving focus in is
announced inconsistently, and this panel should not steal focus from the terminals.

## How to Test Locally

- **View diff**: VSCode sidebar → right-click builder `pir-1450` → **Review Diff**
- **Run the browser evidence** (needs an out-of-tree Playwright; see the script header):
  ```bash
  pnpm build
  npm install playwright-core --prefix /tmp/pw
  PW_CORE=/tmp/pw/node_modules/playwright-core \
  PW_CHROMIUM=~/.cache/ms-playwright/chromium-*/chrome-linux64/chrome \
  node --experimental-strip-types packages/codev/scripts/issue-1450-dashboard-evidence.mts
  ```
  It stands up and tears down its own isolated Tower on port 14700; the live Tower on 4100 and the
  real `global.db` are never touched.
- **What to verify**:
  - The counter is underlined and reads as interactive; clicking opens the panel.
  - Rows show `from → to` (a null sender renders `?`, matching `afx inbox`), with age and reason.
  - With a `--delay` send in flight, `Held (N)` matches the badge and `Scheduled (M)` is separate.
  - Escape closes and returns focus to the counter; click-outside closes; a terminal behind the
    panel does not paint over it.
  - On a normally-running dashboard, the union of both groups matches `afx inbox -w <workspace>`
    row for row, and the Held group alone matches the badge count.

## Post-Consultation Fixes

The 3-way consult and the architect's integration review both landed as non-blocking
(APPROVE / COMMENT / APPROVE). Every finding was verified against the files before acting; all
were real and all are fixed in `2cb2b3623` (see the PR's commit list).

| Finding | Source | Disposition |
|---|---|---|
| `handleInboxList`'s docstring claimed an empty `workspaceOverride` stays scoped, but `rawWorkspace ? … : undefined` widened it to **all workspaces** | Codex + architect (3) | **Real.** Unreachable today (the dispatcher 400s a missing/relative prefix first), but the comment promised a guarantee the code did not make. Made the code true rather than weakening the comment: a scoped call with a blank override now scopes to `''`, which matches no rows. The safe failure for a scoped call is zero rows, never every row. |
| Review file's test counts were wrong | Codex | **Real.** Recounted from the merge-base: **27** new component + **10** formatter + **12** route = **49**, not the 37 originally claimed. Corrected above. |
| `count → 0` with scheduled rows left the panel with no "cleared" notice, and `Scheduled`'s "not counted above" had nothing above it | Codex | **Real.** The notice now keys on `heldRows.length === 0` rather than `messages.length === 0`, with copy that distinguishes "cleared" from "nothing held, rows below are scheduled". |
| arch.md edit swallowed the pre-existing pruning/cron sentences into the new count-vs-list paragraph | Claude + architect (2) | **Real** — content survived but read as part of the Held/Scheduled discussion. Paragraph break restored. (Second splice of this kind in this project; the first was caught in `packages/types/src/api.ts` before commit.) |
| `loadMessages` prop identity drove the refetch effect — an inline lambda from a future caller would loop | Claude + architect (4) | **Real footgun.** Latched in a ref; `load` now has empty deps, so behaviour no longer depends on a caller remembering to memoize. |
| Popover lacked `aria-live`, so asynchronously-loaded rows were never announced | Claude | **Real.** Added `aria-live="polite"` + `aria-busy`. |
| Footer said `afx inbox dismiss <id>` but no id is rendered anywhere | Claude | **Real.** Reworded to `Ids and dismissal: afx inbox`. Rendering a full uuid per row would dominate the row, and this surface never mutates. |
| Server projection was untyped — `HeldMessage` was client-side decoration only | architect (1) | **Real.** Annotated `const projected: HeldMessage[]`, so a drifting projection (dropped field, or a `body` slipping in) fails the server build. |
| Keep prior rows during refetch instead of blanking to "Loading…" | architect (5), optional | **Taken.** A refetch fires on every `count` change while open; flashing the list away is the worst moment to do it. `aria-busy` carries the in-flight state; only a cold open shows the spinner. An **error** still replaces the rows — once a refetch fails, the old list is no longer known to be current. |
| `HeldMessage` jsdoc typo `/api/inbox:id` | architect (6) | **Not reproduced.** The jsdoc already reads `GET /api/inbox/:id` (`packages/types/src/api.ts:597`). No change made. |

Re-verified after the fixes: `pnpm build` ✓, web suite ✓ (373 passed), route suite ✓ (26 passed),
and the browser evidence re-run ✓ **23/23**.

## Flaky Tests

None. No pre-existing failures were encountered in the full suite, so nothing was
skipped or quarantined.

## Scope Note

Out of scope and unchanged, per the plan: VSCode's `mailbox-indicators.ts` (the issue is titled
*Dashboard*); mobile (`MobileLayout` does not render the badge today); dismiss/show-body from the
UI (Spec 1313 decision 8 and the redaction rule); and `codev-skeleton/` (this is product code, not
framework files, so there is no skeleton twin to mirror).

### DIFF
--- a/apps/web/__tests__/HeldCountBadge.test.tsx
+++ b/apps/web/__tests__/HeldCountBadge.test.tsx
@@ -341,6 +341,21 @@ describe('HeldCountBadge', () => {
     expect((await screen.findByTestId('held-empty')).textContent).toContain('Held mail cleared');
   });
 
+  it('says nothing is held when the held rows drain but a scheduled row remains', async () => {
+    // count -> 0 with a pre-due row still listed: the panel is NOT empty, but there is
+    // nothing HELD, and the Scheduled group's "not counted above" needs a line above it.
+    const now = Date.now();
+    const loadMessages = () => Promise.resolve([row({ id: 'sched', notBefore: now + 15_000 })]);
+    const { rerender } = render(<HeldCountBadge count={1} escalated={false} loadMessages={loadMessages} />);
+    await open();
+    rerender(<HeldCountBadge count={0} escalated={false} loadMessages={loadMessages} />);
+
+    const note = await screen.findByTestId('held-empty');
+    expect(note.textContent).toContain('scheduled');
+    expect(screen.queryByTestId('held-group-held')).toBeNull();
+    expect(screen.getByTestId('held-group-scheduled')).toBeTruthy();
+  });
+
   it('unmounts once the user closes a panel whose count already dropped to 0', async () => {
     const { rerender, container } = render(
       <HeldCountBadge count={1} escalated={false} loadMessages={noMessages} />,
--- a/apps/web/__tests__/HeldCountBadge.test.tsx
+++ b/apps/web/__tests__/HeldCountBadge.test.tsx
@@ -1,38 +1,410 @@
-import { describe, it, expect, afterEach } from 'vitest';
-import { render, screen, cleanup } from '@testing-library/react';
+import { describe, it, expect, afterEach, vi } from 'vitest';
+import { render, screen, cleanup, fireEvent, waitFor, act } from '@testing-library/react';
+import type { HeldMessage } from '@cluesmith/codev-types';
 import { HeldCountBadge } from '../src/components/HeldCountBadge.js';
 
 afterEach(cleanup);
 
+/** A held row with sensible defaults; override only what a test cares about. */
+function row(over: Partial<HeldMessage> = {}): HeldMessage {
+  return {
+    id: 'row-1',
+    workspacePath: '/ws',
+    toAgent: 'cost',
+    fromAgent: 'architect',
+    reason: 'busy',
+    escalated: false,
+    createdAt: Date.now() - 90_000, // 1m ago
+    notBefore: null,
+    ...over,
+  };
+}
+
+const noMessages = () => Promise.resolve<HeldMessage[]>([]);
+
+/** Open the panel and wait for its first load to settle. */
+async function open(): Promise<HTMLElement> {
+  fireEvent.click(screen.getByTestId('held-badge'));
+  return screen.findByTestId('held-popover');
+}
+
 describe('HeldCountBadge', () => {
+  // ---------------------------------------------------------------- Spec 1313 contract
+  // These five predate Issue 1450 and must keep passing unchanged: the badge's
+  // count-only / zero-state / attention behaviour is not what the popover changed.
+
   it('renders nothing when the count is 0', () => {
-    const { container } = render(<HeldCountBadge count={0} escalated={false} />);
+    const { container } = render(<HeldCountBadge count={0} escalated={false} loadMessages={noMessages} />);
     expect(container.firstChild).toBeNull();
     expect(screen.queryByTestId('held-badge')).toBeNull();
   });
 
   it('renders nothing for a negative count (defensive)', () => {
-    const { container } = render(<HeldCountBadge count={-1} escalated={false} />);
+    const { container } = render(<HeldCountBadge count={-1} escalated={false} loadMessages={noMessages} />);
     expect(container.firstChild).toBeNull();
   });
 
   it('shows the held count when greater than 0', () => {
-    render(<HeldCountBadge count={3} escalated={false} />);
+    render(<HeldCountBadge count={3} escalated={false} loadMessages={noMessages} />);
     expect(screen.getByTestId('held-badge')).toBeTruthy();
     expect(screen.getByText('3 held')).toBeTruthy();
   });
 
   it('is not in the attention state when not escalated', () => {
-    render(<HeldCountBadge count={2} escalated={false} />);
+    render(<HeldCountBadge count={2} escalated={false} loadMessages={noMessages} />);
     const badge = screen.getByTestId('held-badge');
     expect(badge.className).not.toContain('held-badge--attention');
     expect(badge.querySelector('.held-dot--attention')).toBeNull();
   });
 
   it('enters the attention state (pulsing dot) when escalated', () => {
-    render(<HeldCountBadge count={1} escalated={true} />);
+    render(<HeldCountBadge count={1} escalated={true} loadMessages={noMessages} />);
     const badge = screen.getByTestId('held-badge');
     expect(badge.className).toContain('held-badge--attention');
     expect(badge.querySelector('.held-dot--attention')).toBeTruthy();
+  });
+
+  // ---------------------------------------------------------------- Issue 1450: affordance
+
+  it('is a button wired as a disclosure, collapsed by default', () => {
+    render(<HeldCountBadge count={1} escalated={false} loadMessages={noMessages} />);
+    const badge = screen.getByTestId('held-badge');
+    // A real <button> is what buys keyboard activation (Enter/Space) and focusability for
+    // free — the affordance is not just the underline.
+    expect(badge.tagName).toBe('BUTTON');
+    expect(badge.getAttribute('aria-expanded')).toBe('false');
+    expect(badge.getAttribute('aria-controls')).toBeTruthy();
+    // Disclosure pattern, not a dialog — see the component docstring.
+    expect(badge.getAttribute('aria-haspopup')).toBeNull();
+    expect(screen.queryByTestId('held-popover')).toBeNull();
+  });
+
+  it('points aria-controls at the panel it actually opens', async () => {
+    render(<HeldCountBadge count={1} escalated={false} loadMessages={noMessages} />);
+    const badge = screen.getByTestId('held-badge');
+    const panel = await open();
+    expect(panel.id).toBe(badge.getAttribute('aria-controls'));
+  });
+
+  it('opens the panel on click and loads the messages once', async () => {
+    const loadMessages = vi.fn(() => Promise.resolve([row()]));
+    render(<HeldCountBadge count={1} escalated={false} loadMessages={loadMessages} />);
+
+    await open();
+
+    expect(screen.getByTestId('held-badge').getAttribute('aria-expanded')).toBe('true');
+    expect(loadMessages).toHaveBeenCalledTimes(1);
+  });
+
+  it('does not fetch until the panel is opened', () => {
+    const loadMessages = vi.fn(noMessages);
+    render(<HeldCountBadge count={1} escalated={false} loadMessages={loadMessages} />);
+    expect(loadMessages).not.toHaveBeenCalled();
+  });
+
+  it('closes again on a second click', async () => {
+    render(<HeldCountBadge count={1} escalated={false} loadMessages={() => Promise.resolve([row()])} />);
+    const badge = screen.getByTestId('held-badge');
+
+    await open();
+    fireEvent.click(badge);
+
+    expect(screen.queryByTestId('held-popover')).toBeNull();
+    expect(badge.getAttribute('aria-expanded')).toBe('false');
+  });
+
+  // ---------------------------------------------------------------- Issue 1450: row rendering
+
+  it('renders each message as from → to', async () => {
+    render(
+      <HeldCountBadge
+        count={1}
+        escalated={false}
+        loadMessages={() => Promise.resolve([row({ fromAgent: 'architect', toAgent: 'cost' })])}
+      />,
+    );
+    await open();
+    expect(await screen.findByText('architect → cost')).toBeTruthy();
+  });
+
+  it('renders a missing sender as "?" (matching afx inbox)', async () => {
+    render(
+      <HeldCountBadge
+        count={1}
+        escalated={false}
+        loadMessages={() => Promise.resolve([row({ fromAgent: null, toAgent: 'cost' })])}
+      />,
+    );
+    await open();
+    expect(await screen.findByText('? → cost')).toBeTruthy();
+  });
+
+  it('renders age and reason, and marks an escalated row', async () => {
+    const now = Date.now();
+    render(
+      <HeldCountBadge
+        count={1}
+        escalated={true}
+        loadMessages={() =>
+          Promise.resolve([row({ createdAt: now - 120_000, reason: 'no-live-pty', escalated: true })])
+        }
+      />,
+    );
+    await open();
+
+    const rowEl = await screen.findByTestId('held-row');
+    expect(rowEl.textContent).toContain('2m');
+    expect(rowEl.textContent).toContain('no-live-pty');
+    expect(rowEl.textContent).toContain('!');
+    expect(rowEl.className).toContain('held-row--attention');
+  });
+
+  it('falls back to "held" when the row carries no reason', async () => {
+    render(
+      <HeldCountBadge count={1} escalated={false} loadMessages={() => Promise.resolve([row({ reason: null })])} />,
+    );
+    await open();
+    expect((await screen.findByTestId('held-row')).textContent).toContain('held');
+  });
+
+  // ------------------------------------------------- Issue 1450: the Held / Scheduled split
+  // The blocking finding from plan review: `heldCount` excludes pre-due `--delay` rows
+  // (heldSummaryForWorkspace filters not_before) while GET /api/inbox lists them
+  // (listHeld does not). The panel must render that difference, not hide it.
+
+  it('groups a pre-due row separately so the Held group matches the badge count', async () => {
+    const now = Date.now();
+    const due = row({ id: 'due', fromAgent: 'architect', toAgent: 'cost', notBefore: null });
+    const preDue = row({ id: 'pre', fromAgent: 'architect', toAgent: 'docs', notBefore: now + 15_000 });
+
+    // count=1 is what the server reports for these two rows: only `due` is eligible.
+    render(<HeldCountBadge count={1} escalated={false} loadMessages={() => Promise.resolve([due, preDue])} />);
+    await open();
+
+    const heldGroup = await screen.findByTestId('held-group-held');
+    const scheduledGroup = screen.getByTestId('held-group-scheduled');
+
+    // THE assertion: the group the user reads as "the held mail" is exactly the badge count.
+    expect(heldGroup.querySelectorAll('li')).toHaveLength(1);
+    expect(heldGroup.textContent).toContain('architect → cost');
+    expect(scheduledGroup.querySelectorAll('li')).toHaveLength(1);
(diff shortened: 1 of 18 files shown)


mohidmakhdoomi:
## Architect integration review (claude lane, Medium tier)


## Architectural Fit — strong

The server change is the right seam and follows the established pattern exactly. `handleWorkspaceRoutes` gets an `apiPath === 'inbox'` branch that passes `workspacePath` as `workspaceOverride`, mirroring `handleOverview` (`:2712`) and `handleAnalytics` (`:2736`) verbatim. No new handler, no second projection, no new dependency. Auth is inherited from the single pre-dispatch choke point (`isRequestAllowed`, `:256`), so this adds no unauthenticated surface.

The exact-match on `'inbox'` really is load-bearing, and the three non-reachability tests (`:455`, `:464`, `:473`) are the right way to pin it — the body-bearing `inbox/:id` and mutating `inbox/:id/dismiss` regexes anchor on `^/api/inbox/`, so the workspace prefix can't reach them, and the tests prove the 404 rather than leaving it to inspection.

Client side, `HeldCountBadge` stays presentational with an injected loader (mirrors `CloudStatus`), and the rendering matches `afx inbox` row-for-row — I checked `inbox.ts:130-142`: `?` for a null sender, `→Ns` countdown for pre-due, `scheduled` reason, `!` on escalation. The Held/Scheduled split is correct and the route test at `:412` proves the asymmetry directly (`listHeld` returns 2, `heldSummaryForWorkspace(...).total` is 1).

## Findings

**1. `handleInboxList`'s projection isn't typed as `HeldMessage` — this repo's convention says it should be.** The PR adds `HeldMessage` to `packages/types` but the producer at `tower-routes.ts:2166` is a bare `const projected = rows.map((r) => ({...}))`. The established pattern is to annotate: `const state: DashboardState = {…}` (`:2798`), `const result: OverviewData = {…}` (`overview.ts:1065`). As written the new type is client-side decoration only — rename a field on the server and nothing fails until a human notices. One-word fix: `const projected: HeldMessage[] = …`. Since this PR is what introduces the type, wiring it into the producer is in scope.

**2. `codev/resources/arch.md` — the new paragraph swallowed two unrelated sentences.** The Escalation-&-visibility edit appended "Terminal rows (delivered/superseded/dismissed) are pruned after `mailbox.retentionDays`… Cron delivers through the same gate via `deliverCronMessage`…" onto the end of the new bolded count-vs-list paragraph. Those sentences are about pruning and cron and now read as part of the asymmetry argument. Move them back to their own paragraph.

**3. The `workspaceOverride` comment overstates a security guarantee.** `tower-routes.ts:2137-2140` says the override "wins… so even an empty-string override is honored rather than falling through." But `:2151` then does `rawWorkspace ? normalize(...) : undefined`, so an empty-string override yields `undefined` → **lists every workspace**, i.e. the widest scope, not the narrowest. It's unreachable today (`handleWorkspaceRoutes:2484` rejects a decoded path that doesn't start with `/` or a drive letter), so no live bug — but either drop the claim or make it true (`if (workspaceOverride !== undefined) { workspace = normalize(workspaceOverride) }`).

**4. `loadMessages` prop identity is load-bearing and undocumented as such.** `load` is `useCallback([loadMessages])` and the fetch effect depends on it. `App` passes the module-level `fetchInbox`, so it's stable today. A future caller passing an inline arrow — a mobile layout, a VSCode webview — would refetch on every parent render, i.e. every 2.5s `useOverview` poll, for as long as the panel is open. Either latch the loader in a ref or add "must be referentially stable" to the prop's doc comment. It's an exported public props interface now.

**5. Minor UX: refetch blanks the list.** On a `count` change while open, `load()` sets `{kind:'loading'}`, so the rows the user is reading vanish for the round-trip. Showing "Loading…" only on first load (keeping prior rows during refresh) is a few lines and avoids a flicker at exactly the moment the mailbox moved.

**6. Doc typo:** `HeldMessage` jsdoc says `GET /api/inbox:id` — missing slash.

**7. Evidence-script accretion.** `packages/codev/scripts/issue-1450-dashboard-evidence.mts` (370 lines) is committed, not in CI, needs out-of-tree `playwright-core` and a hardcoded port. It doesn't ship (`package.json` `files` covers only `scripts/forge` + `postinstall.mjs`) and there's precedent (`measure-prompt-behavior.ts`), so it isn't wrong. But the isolated-Tower dance (`NODE_ENV=test` + `AF_TEST_DB` + private port) has now been written three times per the script's own header (send-integration e2e, pir-1365, this) — that's the reusable part, and nothing will tell you when a per-issue copy rots.

---
VERDICT: COMMENT
SUMMARY: Reuses the existing handler behind the correct workspace-scoped seam with real security tests; two cheap fixes (type the server projection, un-splice the arch.md paragraph) and some doc/robustness nits, none merge-blocking.
CONFIDENCE: HIGH
---

KEY_ISSUES:
- `handleInboxList`'s projection (`tower-routes.ts:2166`) is not annotated `HeldMessage[]`, breaking the repo's producer-typed-against-shared-type convention (`DashboardState` at `:2798`, `OverviewData` at `overview.ts:1065`) — the new type is currently client-only, so server/client can drift silently
- `arch.md` edit spliced the pre-existing pruning/cron sentences onto the end of the new count-vs-list paragraph
- The `workspaceOverride` comment (`:2137-2140`) claims an empty-string override is honored; `:2151`'s truthiness check would actually widen scope to all workspaces. Unreachable today, but the comment asserts a guarantee the code doesn't make
- `loadMessages` must be referentially stable or the open panel refetches every 2.5s; undocumented on a now-public props interface

INTEGRATION_NOTES:
- Auth, redaction, and CLI-only dismissal are genuinely preserved — the tests prove `inbox/:id` and `inbox/:id/dismiss` 404 under the workspace prefix and that the row survives a POST, rather than asserting it in prose
- Follow-up candidate: VSCode's `mailbox-indicators.ts` shows the same `N held` in the status bar with no path to the list. The server route it needs now exists; the same Held/Scheduled grouping rule would apply
- Follow-up candidate: `MobileLayout` renders no badge at all, so mobile has neither the count nor the list
- If a second consumer ever needs this rendering, `formatHeldAge`/`isScheduled` belong in `@cluesmith/codev-sdk` (already an `apps/web` runtime dep, environment-agnostic) rather than a third copy. The CLI↔web duplication is unavoidable under #1189 and the current placement is right for one consumer
- Follow-up candidate: extract the isolated-Tower fixture (`AF_TEST_DB` + private port + teardown) into a shared helper so evidence scripts stop reimplementing it — that, not the per-issue assertions, is the part worth keeping

waleedkadous (approved):
Lovely work — this is exactly the affordance the held-mail counter needed, and the evidence discipline is a model: I ran your `issue-1450-dashboard-evidence.mts` myself against an isolated Tower with a real Chromium (23/23, console clean, live global.db untouched) and checked the screenshots by eye — the underlined counter, the Held/Scheduled split with the "not counted above" explainer, the `afx inbox` footer, and the popover painting over a live terminal tab all look right. Both integration lanes (codex, claude) verified the workspace-scoped route keeps the auth and no-body boundaries intact. Approving.

A few things you might consider, none blocking — happy for them to be a follow-up:
- `tower-routes.ts` ~2170: the comment says `const projected: HeldMessage[]` prevents an extra `body` from compiling. The two lanes disagreed on whether contextual typing through `.map()` actually gives you excess-property checking there — cheapest resolution is `satisfies HeldMessage` on each projected object, which makes the guarantee real regardless.
- Three of the post-consult fixes have no test pinning them (keep-rows-during-refetch, `aria-live`/`aria-busy`, and the ref-latched loader — the `does not refetch on an unrelated rerender` test reuses the same function reference, so it would pass against the old code too). A future refactor can undo all three silently.
- `handleInboxList` is now exported purely as a test seam for a branch that's unreachable through `handleRequest`; the other route tests all go through the front door, and dropping the seam costs nothing.
- Docs: the popover refreshes only on `heldCount` changes (matches the plan — just avoid implying every mailbox movement refreshes it), and the new arch.md table line ~672 generalizes "omitted workspace = first known" while `/api/inbox` intentionally lists all — worth qualifying as handler-dependent.
- The evidence script sits outside `tsconfig` include and vitest collection, and it's the third hand-rolled isolated-Tower fixture — extracting that fixture into a shared helper would be a nice small follow-up.

Also noted for the follow-up list: VS Code's `mailbox-indicators.ts` shows the same `N held` with no path to the list, and mobile renders no badge at all — the route you added is what both would need.

PR #1550, opened 2026-08-24
vscode: forward/feedback-hunk press resolves against a fresh, degrading hunk model (#1534)

### LINKED ISSUE #1534: vscode: forward-hunk press errors 'place the cursor in a changed hunk' inside a visibly changed hunk (frozen hunk ranges + rotation/press model split)
## Problem

Forwarding the change under the cursor intermittently fails with the status-bar message **"Codev: place the cursor in a changed hunk"** while the cursor is visibly inside a changed (green) region of a builder diff. Reported 2026-08-20 from live review use (deck press and in-editor use both affected); the reporter also observed "Forward to Builder (lines N-M)" lenses rendering at positions that look wrong.

## Diagnosis (verified against source, not runtime)

The error can only come from the cursor-based forward paths — `codev.forwardCurrentHunkToBuilder` (`apps/vscode/src/extension.ts:1290`) and the feedback mirror `hunkAnchor` (`apps/vscode/src/review-queue/feedback.ts:61`) — both of which test the cursor line against `entry.hunks`. A lens click never checks the cursor (it carries pre-built args), so the defect is in how `entry.hunks` is produced and consumed. Two independent mechanisms produce the symptom:

**1. Rotation and press consult different hunk models (split brain).**
The relay maps `diff-next-hunk` / `diff-prev-hunk` to VS Code's built-in `workbench.action.compareEditor.nextChange` / `previousChange` (`apps/vscode/src/command-relay.ts`) — the diff editor's own live change model. But the press verbs (`forward-hunk`, `feedback-hunk`) validate the cursor against `entry.hunks`, our own parse of `git diff -M --unified=3 <merge-base>` (`apps/vscode/src/diff-inject-ref.ts:parseHunkRanges`). The two models are not the same:
- **Deletion-only changes**: VS Code's navigation stops on them, but `parseHunkRanges` yields no range for a pure deletion (no new-side lines — documented in the function's own comment). Rotate-to-a-deletion then press = guaranteed error, right where the navigation landed the user.
- Algorithmic/anchoring differences between VS Code's diff and git's can shift boundaries even for modifications.

**2. `entry.hunks` is frozen at diff-open time and never refreshed (staleness).**
`viewDiff` and `registerFileInjectSession` compute hunks once when the diff opens (`apps/vscode/src/commands/view-diff.ts:388-405, 467-478`). No `onDidChangeTextDocument` / `onDidSaveTextDocument` hook re-runs the parse (verified by grep). When the builder keeps committing/editing after the reviewer opens the diff, every line shifts: VS Code recomputes its green regions live, our frozen ranges do not. Cursor visibly in green, `hunks.find` misses, error. The same drift misplaces the per-hunk lenses.

**Related gaps surfaced by the same trace:**
- The keyboard path `Cmd/Ctrl+K H` (`resolveCursorRef`, #1073) falls back symbol → hunk → file and never errors; the press verbs check hunks only. So the deck press is strictly narrower than both the lenses (which also advertise symbol ranges) and the keyboard equivalent.
- Lens position note: a "(lines N-M)" lens above a comment banner may be the *symbol* lens behaving per API — `DocumentSymbol.range` includes leading comments — so not every "wrong spot" report is the staleness defect. Worth keeping in mind when verifying the fix.
- The error message is itself misleading (an instrument answering a narrower question than it names, per the #1445/#1448/#1456 principle): the true condition is "the open-time git ranges no longer contain this line", not "you are not in a changed hunk".

## Fix directions (for whoever takes it)

1. **Re-resolve hunks at press time** — a single-file `git diff <baseRef> -- <relPath>` in the press handler (or a debounced refresh on save/change) removes the staleness class. Cheap: one file, one git call.
2. **Give the press verbs the keyboard's fallback** — route `forward-hunk`/`feedback-hunk` through `resolveCursorRef` (symbol → hunk → file) so a press inside a symbol still forwards something sensible instead of erroring.
3. **Align rotation and press on one hunk model** — either drive `diff-next-hunk` from `entry.hunks` or validate the press against the compare editor's change model. Larger; the first two may make it unnecessary.
4. Whatever ships, fix the message: say what is actually known (e.g. "no changed lines recorded at the cursor — the diff may be stale; reopen it"), not an instruction the user has already satisfied.

## Out of scope

The Stream Deck plugin — it sends the correct canonical verbs; all hunk logic is host-side. No plugin release is needed for any of the fixes above.

amrmelsayed:
# PIR Review: forward-hunk / feedback-hunk press erroring inside a visibly changed hunk

Fixes #1534

## Summary

The VS Code reviewer's "forward the change under the cursor" press verbs (`forward-hunk` /
`feedback-hunk`, deck-driven, plus the in-editor commands) intermittently failed with
**"Codev: place the cursor in a changed hunk"** while the cursor was visibly inside a green
changed region. The press validated the cursor against `entry.hunks` — a `git diff` parsed once
when the diff opened and never refreshed — which drifts from what is actually on screen (the
builder keeps committing) and cannot represent a deletion-only change (no new-side line), even
though the dial rotation that *placed* the cursor navigates VS Code's live compare-editor model.
The fix routes both press paths and the Cmd/Ctrl+K H keyboard path through one shared resolver
that re-parses the file's diff **fresh at press time** and **degrades symbol → hunk → file**
instead of erroring, and replaces the misleading message with an honest one.

## Files Changed

`git diff --stat` against the merge-base (source only; the same commit also adds this review and
a lessons-learned entry):

- `apps/vscode/src/commands/press-cursor-ref.ts` (+83 / -0) — **new** shared resolver
- `apps/vscode/src/__tests__/press-cursor-ref.test.ts` (+83 / -0) — **new** unit tests
- `apps/vscode/src/extension.ts` (+24 / -20) — both cursor handlers route through the resolver
- `apps/vscode/src/review-queue/feedback.ts` (+16 / -8) — `hunkAnchor` async via the resolver
- `apps/vscode/src/commands/view-diff.ts` (+12 / -2) — 3 entry-construction sites pass the new fields
- `apps/vscode/src/diff-inject-codelens.ts` (+5 / -0) — entry gains `baseRef` + `worktreePath`
- `apps/vscode/src/__tests__/feedback.test.ts` (+34 / -2) — staleness + whole-file-fallback cases
- `apps/vscode/src/__tests__/{builder-review-ranges,diff-codelens-mode,diff-inject-context-key,submit-review}.test.ts` — fixture updates for the two new required fields

## Commits

`git log main..HEAD --oneline` (excluding porch bookkeeping commits):

- `6c5fc1346` [PIR #1534][Phase: implement] fix: resolve forward/feedback-hunk press against a fresh, degrading hunk model
- `1b4f7afdc` [PIR #1534] Plan draft

## Test Results

- `check-types` (tsc, both tsconfigs): ✓ pass (clean, after building the workspace deps —
  see Lessons)
- `porch check 1534`: ✓ build, ✓ tests
- `pnpm test:unit` (vitest): ✓ 882 passed; **6 new** (`press-cursor-ref.test.ts` ×4:
  fresh-parse-catches-stale, deletion-only→symbol, no-coverage→file, git-fail→frozen-fallback;
  `feedback.test.ts` ×2: fresh-parse-over-stale, whole-file fallback never emits the old error).
- Verification honesty: this was verified by **unit tests + source tracing**. The physical
  Stream Deck dial-press run was scripted and handed to the deck owner; no deck-run result has
  been relayed back as of this writing. The `dev-approval` gate was approved by the human on his
  word, with that provenance understood — **not** on a deck-verified result. If a deck run is
  performed later, its result can be appended here.

## Architecture Updates

No arch changes. This is a bug fix inside the VS Code extension's existing diff-inject seam; it
adds no module boundary, no cross-package contract, and no system-shape fact. `entry.hunks`
consumption stayed local to `apps/vscode`; `DiffInjectSessionEntry` gained two fields but remains
an apps/vscode-local type (it is not in `@cluesmith/codev-types`), so no HOT `arch-critical.md`
fact and no COLD `arch.md` reference detail is warranted.

## Lessons Learned Updates

Routed one COLD entry to `codev/resources/lessons-learned.md` (Debugging and Root Cause Analysis
section) — a reusable pattern, not a per-spec changelog, so it does not belong in the hot,
capped `lessons-critical.md`:

> An error that tells the user to satisfy a condition they *visibly already satisfy* is almost
> always validating against a stale or parallel model of the world, not the user's action —
> treat the message as a lead, not a fact. Includes the two-models-drift shape (a snapshot vs a
> live model), the structural deletion-only mismatch that survives a fresh parse, the
> resolve-fresh-and-degrade fix pattern, and the API ceiling on true single-model unification
> (`textEditorDiffInformation` is a **proposed** API — clean only under an IDE-bundled build
> with a `product.json` grant).

## Things to Look At / Follow-ups

- **Residual (deliberately out of scope):** rotation (VS Code's compare-editor model) and press
  (our git parse) remain two models. After this fix the disagreement can no longer error or
  mis-scope — a deletion the rotation stops on degrades to the enclosing symbol/file — but they
  are not literally unified. True single-model navigation (issue direction "c") would let the
  press read the editor's own computed change regions and delete our parser. That is blocked
  today: those regions are exposed only via the **proposed** `textEditorDiffInformation` API
  (sole consumer: the built-in git extension), which a marketplace extension cannot enable — it
  becomes clean only in an IDE-bundled build with a `product.json` API grant. Worth a separate
  follow-up issue rather than a rider on this fix.
- **Lens position report (not this defect):** a "(lines N-M)" lens sitting above a comment
  banner is usually the *symbol* lens behaving per API (`DocumentSymbol.range` includes leading
  comments), not the frozen-hunk staleness — verified unchanged, not "corrected."
- **No timeout on the press-time git call** (`press-cursor-ref.ts:44`): a healthy worktree
  returns in single-digit ms, and any failure already falls back to the frozen hunks, but a
  pathologically hung `git` (e.g. an index lock) would leave the press pending. Adding
  `{ timeout: 3000 }` to the `execFile` options would degrade a hang straight to the fallback —
  cheap, purely defensive, deferred as it is not part of the observed failure.
- **Scope confirmed clean:** no `apps/streamdeck` changes (the deck sends correct canonical
  verbs; all hunk logic is host-side), no Tower, no `@cluesmith/codev-types`.


amrmelsayed:

**Risk tier: low-to-medium.** The `+671/-37 across 16 files` headline is mostly docs — plan (201), review (95), thread (86), status (27), lessons (2) ≈ 411 lines. Code is **~260 lines across 7 files**, all inside `apps/vscode`. No Tower, no `packages/types`, no `apps/streamdeck` — verified by file list, matching the plan's scope fences exactly.

### Verification chain, stated with its provenance

- **Code verified at `8f4d5c272` pre-PR, and byte-identical at the PR head `6b5d812b8`** — the rebase (10 commits of main drift, zero file overlap) and the review-phase commits changed no source. So the pre-PR verification carries: the old error string is gone from both sites, and all three paths (`forwardCurrentHunkToBuilder`, `hunkAnchor`/`feedbackHunk`, and `Cmd+K H`) route through the single shared `resolvePressCursorRef`.
- **The worst case is provably today's behaviour**: on git failure at press time the resolver falls back to the frozen `entry.hunks`, so no input can make the press *worse* than before the fix.
- **The keyboard path silently had the same staleness** (its file-fallback hid it); it now gets the fresh parse for free by sharing the helper. That is scope the issue didn't ask for but the defect did.
- **Evidence record, honestly bounded:** unit (882 green, incl. fresh-parse-catches-stale / deletion-only / no-coverage / git-fail cases) + independent source verification by two architects. The physical-deck script was handed to the owner; **no deck run result was ever relayed**, and the dev gate was approved on the owner's word. The review artifact says exactly that and claims no more — checked, it doesn't say "deck-verified" anywhere.

### Design decisions carried correctly

- Degrade-over-error (whole-file forward + honest note) was put to the owner at the plan gate with the stricter variant offered, and approved as planned.
- Direction (c) — single unified change model — is recorded as a follow-up **with its real constraint attached**: `textEditorDiffInformation` is a proposed API (verified in the bundled workbench source; its only consumer is the built-in git extension), so a marketplace extension can command compare-editor navigation but cannot enumerate its change model. (c) becomes clean only under an IDE-bundled build with a `product.json` grant. That constraint is what makes the follow-up actionable rather than aspirational.
- `baseRef` + `worktreePath` are **persisted on the entry** rather than re-derived from paths — the owner's established preference for real identifiers over fragile derivation.
- The lessons entry is correctly routed COLD with the tier justification stated.

**No blocking findings.** CMAP verdicts to follow from the builder; merge decision rests with the owner.

Issue #1574, opened 2026-09-01
tower: self-attesting message frames — recipient in every delivered wrapper + reply hint at the point of need (absorbs #1543, closes the real #1530 defect)


waleedkadous:
## Problem

Delivered-message wrappers are not self-attesting, in two ways that have each already cost an incident:

1. **No recipient in the header.** The injected frame is `### [BUILDER <sender> MESSAGE | <time>] ###` — sender and timestamp, no `to:`. Consequence (both halves demonstrated by the #1543 incident): no pane-reader can verify a message on its screen was addressed to it — which is exactly what let an agent confidently (and falsely) assert builder-addressed messages had arrived on its terminal, mobilizing a multi-agent incident response before forensics refuted it; and a *genuine* misdelivery would be invisible to its victim, the one party positioned to detect it.
2. **Nothing in the frame says how to reply.** The #1530 "task-lane replies silently lost" report was refuted by forensics (2026-08-30): the builder never invoked `afx send` — instructed to "reply," it typed its replies as plain assistant text into its own terminal, which goes nowhere by construction, with zero feedback. Protocol-lane builders avoid this because porch prompts drill `afx send architect` repeatedly; a task-lane builder gets a minimal spawn prompt, and after a context refresh even that is gone. The framework hole is that the reply mechanism is not stated at the point of need.

## Fix (prescribed)

All at the formatter/prompt surface — no plumbing, locks, gate, or routing changes.

1. **Recipient in every delivered wrapper** (`utils/message-format.ts`): `### [BUILDER <sender> MESSAGE → <to_agent> | <time>] ###` (same pattern for the architect/user variants). The mailbox already persists `to_agent` per row — thread it to the formatter at enqueue/delivery. Apply to all wrapper variants so every delivery is self-attesting.
2. **Reply hint on architect→builder frames**: append one trailing line to the formatted frame, e.g. `(reply: afx send architect "…")`. Keep it to a single short line; builders' porch prompts remain the primary teaching surface — this is the point-of-need reminder that survives `/clear`.
3. **Task-lane spawn prompt** (`commands/spawn.ts`, `spawnTask` template): add one line naming `afx send architect` as the reply channel, and the same line in the self-refresh re-entry frame if trivially reachable.
4. **Rescope housekeeping**: this issue implements the retitled scope of #1543 (recipient field) and the real defect behind #1530 (reply-channel discoverability); close both when this merges, with a pointer.

## Constraints

[...]

Issue #1564, opened 2026-08-27
afx send: long architect→builder messages truncated or delivered empty-body


amrmelsayed:
Two delivery failures in one session (2026-08-27), same workspace, architect → builder direction, both on longer messages:

1. **Truncation**: a ~1,900-char single `afx send` arrived at the builder as only its tail fragment ("...of the synthesis doc." — the final ~30 chars). Sender saw `[ok] Message delivered`.
2. **Empty body**: a ~900-char send arrived as just the message-header delimiter block with no body at all ("your last message arrived empty (just a '###...' delimiter, no body — and nothing in my inbox)"). Sender again saw `[ok] Message delivered`.

Short messages (≤ ~600 chars) delivered fine throughout the same session, including immediately after each failure — resending the same content split into 2 parts worked.

**Impact**: silent content loss with a success result at the sender — the worst failure shape for gate-decision traffic. In case 2 the builder was correctly paranoid and held at a porch gate rather than treating an empty message as approval, but that safety came from agent discipline, not the transport.

**Suggestions**: (a) size limit enforced loudly at send time (reject or auto-chunk, never silently truncate); (b) delivered-length echo in the send result so sender can detect mismatch; (c) if a render-gate/PTY-write constraint is the culprit, route oversize bodies through the durable mailbox (`afx inbox`) automatically with a pointer injected instead.

Env: afx/codev 3.3.1, Tower 3.3.1, macOS 26, architect and builder terminals in the same workspace.

waleedkadous:
Root cause found (2026-08-30 code investigation) and fix lane opened. Mechanism: the write window is length-proportional — bodies ≥4 lines go out one line per 10ms with no bracketed paste and no mid- or post-write verification, so a ~1,900-char send holds a ~400ms exposure window vs ~100ms for short sends (matching the ~600-char threshold and why splitting the resend worked); any composer disruption mid-window discards the accumulated draft and the trailing Enter submits only the tail (tail-only shape) or nothing (delimiter-only shape). '[ok] Message delivered' is a socket-level claim — no echo/ACK exists before markDelivered. The biggest disruptor (a Tower-side interrupt/escape landing mid-write) is now structurally impossible since PR #1492 merged; the residuals (receiver-side disruption + the false success receipt) are prescribed in #1573 — echo-verification before 'delivered' (verify-or-retry), settle-before-write, and this issue's own suggested loud size limit + delivered-length echo. Keeping this open until #1573 lands and the field shape stops recurring.

[...]

Issue #1521, opened 2026-08-18
tower: mailbox delivery race eats leading bytes when the write lands mid-composer-settle


amrmelsayed:
## Problem

Mailbox delivery (Spec 1313) can intermittently lose the **leading bytes** of a delivered message when the write lands while the receiving terminal's composer is still settling after a turn. The render-gate classifies the prompt as clean, then writes — but if the classification races the composer's settling window, the terminal eats the front of the burst.

## Observed (2026-08-18, two independent sightings)

- The owner's dev-approval relay for pir-1494 arrived in the vscode architect's channel with its header truncated: `ER via VS Code | 2026-08-18T10:52:43.461Z` instead of `[USER via VS Code] | …`. Witnessed first-hand by the receiving architect.
- The pir-1494 builder saw the same corruption intermittently while testing the gate-relay message and diagnosed it: **not** the message formatter (the string is correct at the source and test-pinned), **not** the #1494 change (it does not touch the write path). The loss happens at delivery.

## Why it matters more now

Truncation severity depends on what the leading bytes carry. Until now they were mostly prose. Under the gate-approval relay shipped by #1494, the leading bytes are the `[USER via VS Code]` header — the marker that tells the receiving agent *a human's decision is being relayed*. Losing exactly that prefix downgrades an authorization-bearing message into what reads like ordinary agent traffic. The relay path also fires on-demand (button click), which appears to hit the settling window more often than queued sends.

## Suspected mechanism

Race between the render-gate's clean-prompt classification and the composer finishing its previous turn: classify-then-write is not atomic, so a write that lands mid-settle loses its head. Intermittent; timing-dependent.

## Fix directions (for whoever takes it)

- Settle-before-write: require the prompt to be stable for a short interval after classification before the burst is written, or re-verify cleanliness immediately before the write.
- Sacrificial leading byte(s): prefix the burst with expendable bytes (e.g. a newline) so any swallowed head costs nothing meaningful.
- Either must preserve the mailbox-first contract: persist first, deliver only onto a verified-empty prompt, respond `delivered` | `held`.

## Out of scope

- Any change to #1494's relay content or routing — the message is correct at source; this is a delivery-layer defect.

amrmelsayed:
Two additions from the surface owner, one reassuring and one that raises the stakes.

[...]

PR #1577, opened 2026-09-01
Fix #1573: verify delivery reached the terminal before reporting it delivered

### LINKED ISSUE #1573: tower: delivery write edge — verify-or-retry 'delivered', settle-before-write, loud size limit (silent-loss residuals of #1564/#1521)
## Problem

Two field bugs share one disease — **silent message loss with a success receipt at the sender** — on the channel that carries gate decisions:

- #1564: a ~1,900-char architect→builder send arrived as its final ~30 chars; a ~900-char send arrived as header-delimiter only, no body. Short sends fine; sender saw `[ok] Message delivered` both times.
- #1521: leading bytes eaten when a delivery lands while the receiving composer is still settling after a turn (`[USER via VS Code]` → `ER via VS Code`).

PIR #1365 (PR #1492, merged 2026-09-01) closed the biggest trigger — a Tower-side interrupt/escape landing mid-write — and ended false `delivered` from lock-taking writers. This issue closes the residuals: the **receiving terminal itself** can still eat bytes, and Tower cannot tell.

## Root cause (proven from code, 2026-08-30 investigation)

- `[ok] Message delivered` means "frames queued on a connected socket": `writeMessagePaced` resolves true when every `session.write()` returned true (`message-write.ts`), and `ShellperClient.write` returns true iff the socket object is connected (`shellper-client.ts`) — no echo/ACK of any kind before `markDelivered` (`mailbox-delivery.ts`).
- The render gate's "clean prompt" has **no stability requirement**: a screen that repainted 1ms ago passes identically to one idle a minute. `session.lastDataAt` (`pty-session.ts:823`) is never consulted before the write. The quiescence drain trigger has an accidental 500ms settle window; the on-demand send path (`handleSend` → immediate delivery attempt) and the `'submit'` fast trigger have none — matching #1521's "on-demand relays hit it more."
- The write window is length-proportional (`(lines−1)×10ms + 80ms`, no bracketed paste, per-line writes) with zero mid-write or post-write verification — matching #1564's ~600-char threshold and tail-only/delimiter-only shapes.
- **No size limit exists anywhere** between the CLI flag and the PTY bytes (only the generic 1MiB HTTP body cap and the 48KB `--file` cap).

## Fix (prescribed — trimmed scope, owner-approved 2026-09-01)

Three changes, in leverage order. All in the post-#1492 converged write edge.

1. **Settle-before-write** (~one line + constant + test): in `deliverAgentMail`, immediately before the write, require `now − session.lastDataAt ≥ SETTLE_BEFORE_WRITE_MS` (~250ms, named constant) else `hold('busy')` — the backstop drainer retries on its existing schedule. This gives the request-time and submit-trigger paths the same accidental safety the quiescence path already has.
2. **Loud size limit + delivered-length echo**: reject message bodies over **48KB** (match the existing `--file` bound; one shared constant) at `handleSend` with a clear error naming the limit and suggesting `--file` or splitting; mirror the check client-side in `commands/send.ts`. Add `bodyLength` to the send response and print it in the CLI result. Never silently truncate.
3. **Echo-verification before `markDelivered`** (the medium change — **strictly scoped**): after `writeMessagePaced` resolves and before `markDelivered`, wait a short bounded interval and check the session's rendered mirror (the same buffer the render gate classifies) for the message's **header line only**. Present → `delivered`. Absent → do NOT mark delivered; hold the row (existing hold/redeliver machinery; rows already escalate for visibility, so a persistent failure is loud, not a loop). Direction of error becomes duplicate-delivery, never silent loss.
   - Scope guard: header-line presence only. No screen diffing, no repair logic, no per-harness heuristics. If this cannot be done narrowly and reliably (e.g. the mirror scrolls the header away for very long bodies — note the 48KB cap bounds this), implement 1+2, cut 3, and say so in the PR.
4. *(Optional, only if ≤ a couple of lines)*: sacrificial leading newline before the header so residual head-eating costs nothing. Drop at the first sign of harness weirdness.

**Explicitly out of scope**: bracketed-paste conversion of the body (structural fix for bare-`\n` Enter risk — needs per-harness verification; file separately), any change to gate classification, locks (#1492 territory), or registry/routing (#1368's lane).

## Testing

- Control test reproducing today's false-`delivered`: a fake session whose mirror never shows the header still gets `markDelivered` on current code; fails post-fix.
- Settle: a write attempted <250ms after mirror data → held; after → proceeds.
- Size: over-limit rejected loudly at route and CLI; at-limit passes; `bodyLength` echoed.
- Echo-verify: header present → delivered; absent → held and redelivered by backstop; the #1492 suite stays green (same files).

Refs: #1564, #1521, #1365 / PR #1492, tracking #1483.

waleedkadous:
## Summary

`[ok] Message delivered` meant only "frames queued on a connected socket" — Tower had no evidence the receiving terminal absorbed anything. This adds the three checks that make the receipt honest: the screen must have settled before we write, the body must be within a stated size limit, and the message's header must actually appear on the terminal — as a result of *this* write — before the row is marked delivered.

Fixes #1573

## Root Cause

Confirmed from the post-#1492 code, not inferred:

- **No arrival evidence.** `submitMessagePaced` resolves `written` when every `session.write()` returned true, and `ShellperClient.write` returns true iff the socket object is connected. `deliverAgentMail` called `markDelivered` on that alone — no echo, no ACK. A composer that ate the leading bytes (#1564) was indistinguishable from a clean delivery.
- **No stability requirement in the gate.** The render gate proves a composer is *empty*, never that it has finished being *painted*. `session.lastDataAt` was never consulted. The quiescence drain trigger had an accidental 500 ms settle; the request path and the `submit` fast trigger had none — matching #1521's "on-demand relays hit it more".
- **No size limit anywhere** between the CLI flag and the PTY bytes, only the generic 1 MiB HTTP cap and the 48KB `--file` cap.

## Fix

**1. Settle-before-write.** `DeliverySession` gains `lastDataAt`. Delivery requires `now − lastDataAt ≥ SETTLE_BEFORE_WRITE_MS` (250) both before the per-terminal lock and again inside the precheck, mirroring how the other three write-instant conditions are already double-checked. Written as a positive `>=` so a session with no usable timestamp (NaN) reads as *not* settled and holds — a fail-open there would look like a guarantee while being none. One existing test double did in fact lack the field.

**2. Loud 48KB limit.** `MAX_MESSAGE_BYTES` + a shared error string in `utils/message-format.ts`, enforced at the CLI (local, immediate) and at `POST /api/send` (public route). `MAX_FILE_SIZE` in both `commands/send.ts` and `commands/reset.ts` is now *defined as* that constant, so the three former copies of `48 * 1024` cannot drift, and the CLI checks after the `--file` append because attachment content travels in the same body. `bodyLength` rides every send response that carried a body, through the SDK to `afx send`'s success line. Never a silent truncation.

**3. Echo verification.** A new required `watchEcho` port. It is opened immediately **before** the write and samples how many times the message's header currently appears on the session's `gateScreen` — the same mirror the gate classifies; the returned `EchoWatch.verify()` then polls (50 ms, up to 600 ms, full retained buffer including scrollback) until that count is strictly **greater**.

New evidence, not mere presence, and that distinction is load-bearing: a held attempt leaves its own echo in the same scrollback, so a presence check would match the copy the *first* attempt left behind and certify a retry whose bytes were swallowed — the false receipt postponed by one redelivery rather than removed. Returning an `EchoWatch` rather than exposing sample and verify as two ports makes the ordering impossible to get wrong.

A negative means "could not confirm", so the row is **held** for the existing redelivery machinery: the direction of error becomes a duplicate the agent can see, never a silent loss the sender was told was a success.

### Why the needle is normalized

I measured this against live PTYs rather than guessing, because the issue's scope guard asks for change 3 to be cut if it cannot be done reliably.

| harness | body | buffer | header found |
|---|---|---|---|
| claude 2.1.252 | 20 lines | normal | exact form **NO**, normalized YES, from +200 ms |
| claude | 300 lines / 24.9KB | normal + scrollback | exact YES, buffer line 6, every sample |
| codex 0.146.0 | 20 lines | normal | exact YES, every sample |
| agy | 12 lines | **alternate** | unmeasured (unauthenticated here) |

The load-bearing surprise: **claude markdown-renders the header on submit.** The composer echoes `### [ARCHITECT INSTRUCTION | <ts>] ###` verbatim; the transcript then shows `[ARCHITECT INSTRUCTION | <ts>]`, the fences consumed as an H3. An exact-line match would fail on every short claude delivery — the common case. Reducing both sides to their alphanumeric skeleton fixes that and incidentally absorbs quote prefixes, indentation, and wrapping. It stays header-only and harness-agnostic: no screen diffing, no repair, no per-harness branches. The tail is deliberately not used — #1564's message arrived as its *final* ~30 characters, so a footer needle would have certified the exact corruption this check exists to catch.

## Deliberate omissions and residuals

- **The issue's optional item 4 (sacrificial leading newline) is not included.** `writeMessageToSession` would send a bare `\n` as its own write, and whether a harness treats that as "insert newline" or "submit" is exactly the per-harness behaviour I could not measure for codex and agy. An empty submit ahead of every message is a worse failure than the head-eating it guards against, which change 1 already addresses. The issue marks it optional and says to drop it at the first sign of harness weirdness.
- **Messages whose normalized first line is under 12 characters skip verification** and keep the pre-#1573 behaviour. A two-character raw send would match incidental screen text, so confirming it would be a rubber stamp — better to leave it unguarded than to certify it falsely. Verification is therefore not universal: it covers every formatted send (all `afx send` traffic that carries a header) and raw sends with a distinctive first line.
- **agy uses the alternate screen buffer**, which has no scrollback, so a message longer than one viewport could scroll its header out of reach and fail to confirm. Consequence is a redelivery, never a dropped message. Documented at the binding; unmeasured because agy is unauthenticated in this environment.
- **Nothing bounds redelivery when verification never confirms** — filed as #1578. Escalation makes such a row visible (`markEscalated` sets a flag and fires an SSE event) but does not stop the drainer, so the rewrite loop itself is unbounded: one duplicate per clean gate for as long as the harness refuses to show the header. Deliberately not bounded here — every way to bound it means either a new mailbox column or accepting a false delivery after N attempts, and that trade is a decision rather than a bugfix.
- **A count false-negative is possible on very long writes**: if the write evicts the pre-write copy of the header from the 1000-line mirror, the count comes back equal rather than greater and a genuine delivery reads as unconfirmed. Safe direction (a redelivery), and it compounds the previous point.
- **An unconfirmed delivery re-reads and re-normalizes the retained buffer once per poll** while the `afx send` request waits. Bounded by the 600 ms timeout and off the happy path entirely — the measured case confirms on the first read.
- **Verification passes on the composer echo as well as the transcript**, so a *swallowed Enter* (typed but never submitted) still verifies. Distinguishing composer from transcript is the classifier's job and is out of scope; the dirty composer holds all following mail, so it surfaces.
- **A session that emits output more often than every 250 ms while showing a clean composer would never settle** and its mail would starve. Unlikely in practice — output normally makes the composer classify busy, and the existing quiescence trigger already assumes a 500 ms quiet window — and the starvation notice covers visibility.
- **Behaviour change worth naming:** a `--file` attachment at or near 48KB plus any message text now fails where it previously went through — for `afx send` and, because its `--file` addendum rides inside the generated prompt, for `afx refresh` too. That is the shared ceiling working as intended, but it is a real change. Both commands check the total body locally (shared `messageLimitError`), so the refusal is immediate and identically worded rather than a 400 from the route.

## Test Plan

- [x] Regression test added — `bugfix-1573-delivery-verification.test.ts`, 18 tests: the settle window (inside / at the boundary / re-checked in-lock / NaN), the **control test** (completed write + header never shown → held, not delivered, no broadcast), confirmed delivery with the `watch → write → verify` ordering pinned, redelivery of a held row, short-needle skip, needle normalization against the three measured rendered forms, and `watchEchoOnScreen` against a real `SessionScreen` (composer form, markdown-stripped form, scrolled-into-scrollback, **stale header from an earlier attempt**, a redelivery that does land, absent, early return). Plus route tests for over-limit / at-limit / `bodyLength`, and CLI tests for the local refusal, the `--file` interaction, and the byte-count echo.
- [x] Verified the tests are real: with the settle and verify branches disabled, 6 of 16 fail; with verification reduced to presence-only, the stale-header test fails. Restored, all pass.
- [x] Build passes
- [x] All tests pass — 5344 passed, 48 skipped, 0 failures

## Review

**CMAP round 1** — gemini APPROVE, claude APPROVE, **codex REQUEST_CHANGES**: the stale-evidence hole above, which was real and is fixed in `d611bb7`. Claude's smaller findings (a stray `eslint-disable` left by a removed debug line, and `reset.ts`'s third `48 * 1024` literal) are fixed in the same commit.

**CMAP round 2** — gemini APPROVE, **codex APPROVE** (flipped, no remaining issues), claude COMMENT with five findings, all addressed: the unbounded-redelivery residual is filed as #1578; `afx refresh`'s route-400 regression is fixed with a local precheck; the stale `{@link DeliveryPorts.verifyEcho}` doc link and the test count in this body are corrected; the count false-negative and the per-poll re-normalization are documented as residuals here and at the binding.

`tower-routes.test.ts`'s `gateSession` double now echoes writes into its own mirror, because a real terminal does and the delivery path now depends on it. The #1492 suite stays green.

## Rebased onto #1575 (self-attesting frames, #1574)

#1575 landed first and touches `message-format.ts` and `tower-routes.ts`. Both conflicts were additive-vs-additive and keep both sides.

The new frame **composes with echo verification rather than breaking it**, and strengthens it. Headers now read `### [ARCHITECT INSTRUCTION → <toAgent> | <ts>] ###`; the needle is derived from the row's actual `formatted_message`, so it picked the new shape up for free. The arrow normalizes away as punctuation while the recipient **name** stays in the needle — so verification is now recipient-specific: a frame delivered to the wrong agent cannot satisfy the right agent's check. The frame is still 3 lines, so #1574's own constraint about `PACED_WRITE_LINE_THRESHOLD` holds and the write path this PR guards is byte-identical.

Fixtures now **build** the frame with `formatArchitectToBuilderMessage` rather than hand-writing it, plus an explicit assertion that the recipient segment reaches the needle. That is the durable lesson from the rebase: a hand-copied header would have gone on passing while the frame it modelled drifted away from what the delivery path actually writes.

Post-merge: **5360 passed, 48 skipped, 0 failures** (including #1574's 16 new tests).


waleedkadous:
Architect review (posted as comment — GitHub blocks a formal review from the PR author's own account): APPROVED and admin-merged on the owner's explicit word. Matches the #1573 prescription with three reviewed-and-accepted deviations (sacrificial newline dropped per the issue's escape hatch; unbounded-redelivery residual escalated as #1578; shared 48KB ceiling's --file behavior change accepted). The occurrence-count echo-verification (new evidence, not presence) is the right answer to codex's stale-echo finding, and measuring real harness rendering before shipping the matcher was exactly the discipline this write edge needed. CI green on 416eaf32 post-rebase.

Issue #1583, opened 2026-09-02
Tower/afx: one sent message re-injected into a builder prompt dozens of times (byte-identical re-delivery loop), plus same-day truncation and stuck-composer incidents


amrmelsayed:
Owner-directed filing (2026-09-02, shannon workspace, voice lane). Three incident classes hit one architect↔builder pair in a single day; the first is the headline defect and plausibly causes the second-order symptoms.

## 1. Re-delivery loop (core defect)

A single wrapped afx message — sent by the architect via `afx send builder-spir-4679 --file <path> '<note>'` at **2026-09-02T09:45:46.866Z** (the 4679 verify-evidence payload) — was surfaced into the recipient builder's prompt **dozens of times, byte-identical** (same timestamp, same body), as if one delivered/held message were re-injected repeatedly rather than once. It forced a response loop on the recipient; the builder recognized the duplication and held rather than re-acting. Evidence: the repeated identical 09:45:46.866Z blocks in builder-spir-4679's transcript (shannon workspace).

Possibly aggravating: during the same window the sender ran `afx interrupt builder-spir-4679` twice (09:50, 09:53) on the af-mailbox watcher's advice, then an `afx send --interrupt` bypass (09:53) — if interrupt reprocesses the held queue without consuming, that may be the replay engine.

## 2. Stuck composer / held-mail wedge (same pair, same window)

af-mailbox notices at 09:48 and 09:53: messages held (reason: busy) for 3–7m because the recipient's 'composer never classifies as a ready prompt'; cron nudges did not deliver; two ESC+Enter interrupts did NOT clear it (the 09:53 watcher notice fired after both). Held ids observed: cb24419d-9bb9-4623-a560-03bfceeabc79, 8c10f62e-baf0-4069-b5c1-b0fb57c39286 (later dismissed by sender), c060d2bf-94c1-405b-a195-2a19cfc0373e. If item 1's replay loop was occupying the composer, these are downstream symptoms.

## 3. Mid-message truncation (same day, 3×, length-correlated)

[...]

Issue #1578, opened 2026-09-01
tower: bound redelivery when echo verification never confirms (#1573 residual)


waleedkadous:
## Problem

#1573 (PR #1577) made mailbox delivery hold a row when the message's header never appears on the receiving terminal, so a lost message is retried instead of being falsely reported delivered. That is the right direction of error — but nothing bounds the retrying.

If a harness is *systematically* unverifiable, every clean-gate pass writes the same message again, forever. The agent sees the instruction re-typed once per turn, and the row never resolves.

The known way to get there: a harness rendering on the **alternate screen buffer** (agy boots into it) has no scrollback, so a message longer than one viewport scrolls its own header out of reach and can never be confirmed. A second, narrower path compounds it: a very long write can evict the pre-write copy from the 1000-line mirror, so the occurrence count comes back equal rather than greater and a genuine delivery reads as unconfirmed.

#1573 assumed this would be loud rather than looping, because held rows escalate. Escalation is **visibility only** — `markEscalated` sets a flag and fires an SSE event; it does not stop the drainer from re-delivering. Two CMAP reviewers flagged the gap on PR #1577 and it is recorded there as a known residual.

## Suggested fix

Bound the attempts. The obvious shapes, cheapest first:

1. An in-process attempt counter in the delivery module (a `Map` keyed by row id, pruned with the held set). After N unconfirmed writes, stop rewriting and leave the row held + escalated, or mark it delivered with a loud log. No schema change; lost across a Tower restart, which is acceptable for a bound.
2. An `attempts` column on the `mailbox` table — durable, but a migration.

Either way the choice to make explicitly is what happens at the bound: **stop retrying and stay held** (the agent may never get the message, but nothing is spammed) versus **accept and mark delivered** (at most N duplicates, then today's pre-#1573 behaviour, loudly logged). The sender is not misled either way — the `afx send` response already reports `held` on the first unconfirmed attempt.

Worth measuring agy's actual behaviour first: it was not measurable during #1573 (unauthenticated in that environment), so the headline trigger is inferred from its alternate-screen boot, not observed.

Refs #1573, PR #1577.

waleedkadous:
[...]

PR #1585, opened 2026-09-02
Fix #1584: never re-write a mailbox row whose write completed

### LINKED ISSUE #1584: tower: HOTFIX — echo-verify re-write loop re-injects a delivered message unboundedly (3.3.2 regression from #1573; zero re-writes after a completed write)
## Problem (3.3.2 regression — hotfix)

#1583 (shannon, 2026-09-02): one `afx send --file` message was re-injected into a builder's prompt **dozens of times, byte-identical**; a separate architect→architect message was delivered twice. Both on 3.3.2. Root cause confirmed at current main:

- `mailbox-delivery.ts:694`: after the gated write has **completed** (every byte + Enter accepted by the PTY), a failed echo-verification returns `hold('busy')`.
- The next clean-prompt pass re-runs the row with a **full re-write**. No per-row attempt cap exists anywhere in the module.
- `busy` holds are deliberately excluded from escalation streaks (`isClassifierStuck`), so the loop is silent to the owner.
- Field trigger is the common case, not the agy corner #1578 predicted: a recipient that **starts responding immediately** scrolls the header out of the sampled screen (or a long write evicts the pre-write copy from the mirror), so the occurrence count does not increase, verification fails, the message is re-injected, the recipient responds again — self-sustaining. `afx interrupt` worsens it (each interrupt = fresh clean prompt = another pass).

Verify-then-rewrite was the dangerous half of #1573. The honest half (settle-before-write, 48KB cap, "tell the sender when we could not verify") stays.

## Fix (prescribed — owner-approved design 2026-09-02: ZERO re-writes)

A row whose write completed is at-least-once delivered and **must never be written again**.

1. At the verify site (`mailbox-delivery.ts` ~694): on verification failure after a completed write, do **not** hold. Allow one bounded additional verify window (no bytes written — e.g. one more `watch`-style sample after a short delay, ≤ ~1.5s total) to accommodate slow renderers; if still unconfirmed, **mark the row `delivered`** and:
   - log `WARN` (`delivered-unverified`, row id, to_agent, terminal id, header needle length);
   - set the row's `escalated` flag (existing column) so the dashboard/SSE surface it;
   - return a result that carries `verified: false`.
2. Thread `verified` (boolean) additively onto the immediate-delivery send response (`tower-routes.ts` handleSend, `sdk/tower-client.ts` typing, `commands/send.ts` prints e.g. `[ok] Message delivered (unverified — header not seen on the terminal)`). Additive field; older clients unaffected.
3. Remove/neutralize any path by which a completed-write row can be re-written: audit every `hold(...)` reachable after `submitMessagePaced` resolves successfully; only *pre-write* prechecks may hold.
4. `isClassifierStuck` / `recordStreak`: no change needed once (1) lands (no busy-hold after a completed write exists anymore) — confirm with a test rather than assume.

Out of scope: smarter verification (scrolled-header tolerance), attempt columns/migrations, #1567 truncation class, agy measurement. Those remain under #1578 (retitle it to "smarter verification" follow-up, keep open).

## Testing

- **Control test (must fail on current main, pass after)**: fake session whose mirror never shows the header → the message is written **exactly once**; the row ends `delivered` + `escalated`, never `held`; a second drainer tick writes nothing.
- Immediate-responder simulation: mirror shows the header, then scrolls it out before verification samples → written once, delivered (verified or unverified, but never re-written).
- Verified path unchanged: header count increases → `delivered`, `verified: true`, not escalated.
- Send response / CLI: `verified:false` surfaces as the unverified wording; `verified:true`/absent prints today's line.
- Keep the #1573 and #1365 suites green; update any #1573 test that asserted the hold-and-redeliver behavior to assert the new contract instead of deleting it.
- Field verification before close (architect-driven, post-release): long `--file` send to a builder that responds immediately → exactly one copy in its transcript.

## Release

Hotfix release 3.3.3 to `@latest` immediately on merge. Note in the PR that this supersedes the residual recorded in #1578.

Refs: #1583 (field report), #1578 (residual this replaces), #1573 / PR #1577 (regression source), #1564/#1521 (the verification's original purpose).

waleedkadous:
## Summary

Hotfix for a 3.3.2 regression from #1573 (PR #1577). A failed echo-verification *after* a completed PTY write returned `hold('busy')`, so the drainer re-wrote the whole message on every later clean-prompt pass — with no attempt cap. #1583 saw one `afx send --file` re-injected into a builder's prompt dozens of times, byte-identical, silently.

A row whose write completed is at-least-once delivered and is now **never written again**.

Fixes #1584

## Root Cause

`mailbox-delivery.ts:694` (pre-fix). After `ports.writeMessage(...)` returned `{ status: 'written' }` — every byte and the trailing Enter accepted by the session — a `false` from `echo.verify()` re-held the row:

1. The row stayed `held`, so `MailboxDrainer.tick()` (1.5s backstop) and `scheduleDrain` re-ran `deliverAgentMail` for that agent, which performed a **full re-write** of `formatted_message`. No attempt counter exists anywhere in the module and the `mailbox` table has no attempts column.
2. `isClassifierStuck()` deliberately excludes `busy`, so the loop never reached liveness telemetry — it was invisible to the owner.
3. The field trigger is the *common* case, not the agy corner #1578 predicted: a recipient that starts responding immediately scrolls the header out of the 1000-line `SessionScreen` mirror (or a long write evicts the pre-write copy), so `countEchoOnScreen` never sees `count > before`. Verification fails → the message is re-injected → the recipient responds again. Self-sustaining. `afx interrupt` makes it worse: each interrupt is a fresh clean prompt, i.e. another pass.

The in-tree proof: the #1573 suite's "a row held by a failed verification is redelivered by the next pass" asserted `writes == [FORMATTED, FORMATTED]` — the loop was pinned as intended behaviour.

## Fix

Per the owner-approved design in the issue (zero re-writes). Verify-then-rewrite was the dangerous half of #1573; the honest half — settle-before-write, the 48KB cap, telling the sender when we could not verify — stays.

**The delivery is committed before it is verified.** `markDelivered` now runs immediately after `status === 'written'`, before any further `await` or fallible call. That ordering is the fix: while the row still reads `held` it is re-writable by the next drainer tick, so verifying first would leave a ~1.2s window in which a Tower crash/restart — or a `verify()` that *rejects* rather than answering — reopens the loop. `better-sqlite3` is synchronous, so committing here leaves no window at all.

Echo verification still runs — it remains the only end-to-end evidence the bytes reached the terminal — but it is now pure **reporting**, downstream of a committed delivery:

- one extra bounded `verify()` window (a second *look*, zero bytes written; ~1.2s total since `ECHO_VERIFY_TIMEOUT_MS` is 600) to accommodate a slow renderer;
- a rejection is caught and treated as unconfirmed, never propagated;
- still unconfirmed → `markEscalatedDelivered`, a `WARN` log tagged `delivered-unverified` carrying row id / to_agent / terminal id / needle length, and `verified: false` on the outcome;
- **plus a human-visible notice** via a new optional `DeliveryPorts.onUnverifiedDelivery` port. A *delivered* row is invisible to every held-scoped surface (`afx inbox`, `heldSummaryForWorkspace`, the held-count indicator), so the `escalated` flag alone reaches nobody — fine for an interactive `afx send`, which gets `verified: false`, and not fine for a cron or backstop delivery with no sender waiting. Its live binding rides the same generic `notification` SSE channel `surfaceLiveness` uses, with truthful wording; it is deliberately **not** `onEscalation`, whose title says "held past escalation age" and would be a false statement about a delivered row.

`db/mailbox.ts` gains a delivered-only `markEscalatedDelivered`: the commit-first ordering means the verdict is known only after the row has left `held`, where the existing held-only `markEscalated` would silently no-op. A separate function rather than a relaxed guard — held-only is what makes the drainer's escalation pass safe to call blindly.

`verified` is threaded additively onto the immediate-delivery response through `tower-routes.ts handleSend` → `sdk/tower-client.ts` → `commands/send.ts`, which prints `Message delivered to X (N bytes) (unverified — header not seen on the terminal)`. `verified: true` or absent prints today's line, so older clients are unaffected. It is reported only when *this request's own pass* delivered *this* row (a pass picks the agent's oldest held row).

**Audit of every `hold(...)` reachable after `submitMessagePaced` (issue item 3):** the verify site was the only one. `dropped` lost bytes mid-pace to a dead socket; `preempted` may have been cleared or truncated by an unserialized operator write; `contended`/`aborted` wrote nothing. None is a completed write, so all four may still hold — documented at the point-of-no-return comment. `cron-delivery.ts` shares `deliverAgentMailSerialized`, so there is no second verify site.

## Test Plan

- [x] Regression test added — `bugfix-1584-no-rewrite-after-write.test.ts`, **12 tests**.
  - **Control test**: a terminal that never shows the header → written **exactly once**, row ends `delivered` + `escalated` (never `held`), and two real `MailboxDrainer.tick()`s write nothing.
  - **Durability**: the row's DB status read at the instant `verify()` is first consulted is already `delivered`; and a `verify()` that **throws** yields `verified: false` with no second write.
  - **Immediate-responder simulation** against a REAL `SessionScreen`: the header is echoed, then buried under 2000 lines of the recipient's own response → written once, delivered, not re-injected.
  - **Issue item 4, confirmed rather than assumed**: twelve unverified deliveries in a row produce twelve writes and **zero** `onLiveness` calls (`LIVENESS_STREAK_THRESHOLD` is 10).
  - Plus the notice's contents, no-notice-when-confirmed, the WARN log, at-most-two verify windows, the delivery broadcast, the unchanged verified path, and a pre-write hold that must still hold.
  - Six of the original seven fail against the unmodified module at HEAD; three of the five added after review fail against the pre-review commit. Both verified by swapping the module back in and re-running.
- [x] The two #1573 tests that pinned hold-and-redeliver are **rewritten to the new contract, not deleted** — the scenario is unchanged, only the answer differs.
- [x] Route (`tower-routes.test.ts`) and CLI (`send.test.ts`) coverage for `verified`: `false` surfaces the unverified wording; `true` and absent print today's line.
- [x] Build passes (`pnpm build` in `packages/codev`, tsc clean).
- [x] All tests pass — **5377 passed, 0 failed, 48 skipped** across 273 files. (Note for reviewers: a fresh worktree fails 12 files until `pnpm install` + `pnpm build` materializes `packages/codev/skeleton/`; I confirmed those failures are identical at HEAD before building.)
- [x] CI green — 7/7 checks.
- [ ] Field verification before close (architect-driven, post-release): a long `--file` send to a builder that responds immediately → exactly one copy in its transcript.

## Review

Two CMAP rounds. Round 1: gemini=APPROVE, codex=REQUEST_CHANGES, claude=COMMENT — every finding was real and fixed rather than rebutted. Codex caught that the point of no return was not *durable* (the row stayed held across verification; my item-3 audit had looked for `hold(...)` calls and missed the throw path) and that the `escalated` flag reached no human surface. Claude caught three doc blocks that still asserted the **inverted** safety property — `watchEcho`, `EchoWatch.verify` and `watchEchoOnScreen` all still said a negative means "hold for redelivery, a duplicate rather than a silent loss", when those very residuals are the #1583 trigger.

Round 2 after the fixes: gemini=APPROVE, codex=COMMENT (stale PR body + two stale comments — this body is the rewrite, and both comments are fixed), claude pending at time of writing.

## Release

Per the issue: hotfix release **3.3.3 to `@latest`** immediately on merge. This **supersedes the residual recorded in #1578**, which stays open as the "smarter verification" follow-up (scrolled-header tolerance, agy measurement).

https://claude.ai/code/session_01VS5Pk5bRniMHMJreVZYvsN

### DIFF
--- a/codev/projects/bugfix-1584-tower-hotfix-echo-verify-re-wr/status.yaml
+++ b/codev/projects/bugfix-1584-tower-hotfix-echo-verify-re-wr/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1584
+title: tower-hotfix-echo-verify-re-wr
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-09-02T14:02:15.761Z'
+updated_at: '2026-09-02T14:02:15.762Z'
--- a/codev/projects/bugfix-1584-tower-hotfix-echo-verify-re-wr/status.yaml
+++ b/codev/projects/bugfix-1584-tower-hotfix-echo-verify-re-wr/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-09-02T14:33:45.499Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-09-02T14:02:15.761Z'
-updated_at: '2026-09-02T14:16:20.185Z'
+updated_at: '2026-09-02T14:33:45.500Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-1584-tower-hotfix-echo-verify-re-wr/status.yaml
+++ b/codev/projects/bugfix-1584-tower-hotfix-echo-verify-re-wr/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1584
 title: tower-hotfix-echo-verify-re-wr
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-09-02T14:02:15.761Z'
-updated_at: '2026-09-02T14:02:15.762Z'
+updated_at: '2026-09-02T14:05:43.590Z'
--- a/codev/projects/bugfix-1584-tower-hotfix-echo-verify-re-wr/status.yaml
+++ b/codev/projects/bugfix-1584-tower-hotfix-echo-verify-re-wr/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1584
 title: tower-hotfix-echo-verify-re-wr
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-09-02T14:02:15.761Z'
-updated_at: '2026-09-02T14:05:43.590Z'
+updated_at: '2026-09-02T14:16:20.185Z'
--- a/codev/state/bugfix-1584_thread.md
+++ b/codev/state/bugfix-1584_thread.md
@@ -84,3 +84,22 @@ in and re-running); the 7th is the pre-write-hold guard, which should pass on bo
 
 Updated (not deleted) the two #1573 tests that pinned hold-and-redeliver, plus route
 (`tower-routes.test.ts`) and CLI (`send.test.ts`) coverage for `verified`.
+
+## 2026-09-02 — PR
+
+PR #1585 opened (`Fixes #1584`). Three commits: the fix, the tests + this thread, and a
+follow-up tightening (report `verified` only for the row *this* request's pass delivered — a
+pass picks the agent's OLDEST held row, so without an id check a concurrent drainer delivery
+of our row could be reported with another message's verification result).
+
+**Environment note for whoever picks this up:** the worktree arrived with no `node_modules`
+and no `packages/codev/skeleton/`. 12 test files (adopt, update, hot/cold-tier, protocol-drift,
+consult, consolidate, spawn-retirement, session-manager, …) fail until `pnpm install` +
+`pnpm --filter "@cluesmith/codev^..." build` + `pnpm build` in `packages/codev`. I confirmed
+the identical per-file failure counts at HEAD before building, so none were mine. After the
+build: **5372 passed, 0 failed, 48 skipped** across 273 files.
+
+`consult` did NOT auto-detect the project from builder context — it listed every project and
+exited 0 without reviewing. `--project-id bugfix-1584` fixes it. Worth an issue if it recurs.
+
+CMAP verdicts: gemini=APPROVE (no issues, 12.8s). codex + claude pending.
--- a/codev/state/bugfix-1584_thread.md
+++ b/codev/state/bugfix-1584_thread.md
@@ -103,3 +103,40 @@ build: **5372 passed, 0 failed, 48 skipped** across 273 files.
 exited 0 without reviewing. `--project-id bugfix-1584` fixes it. Worth an issue if it recurs.
 
 CMAP verdicts: gemini=APPROVE (no issues, 12.8s). codex + claude pending.
+
+### CMAP round 1 — gemini=APPROVE, codex=REQUEST_CHANGES, claude=COMMENT
+
+Both non-approvals were correct. Fixed rather than rebutted.
+
+**codex #1 (critical, and I had missed it):** the point of no return was not DURABLE. The row
+stayed `held` in the database through up to ~1.2 s of verification, so a Tower crash/restart in
+that window — or an `echo.verify()` that *rejects* rather than answering — left it held and
+re-writable by the next tick. My item-3 audit had looked only for `hold(...)` calls and missed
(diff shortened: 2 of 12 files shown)


waleedkadous:
Architect review (as comment — GitHub blocks a formal review from the PR author's own account): APPROVED and admin-merged on the owner's word. Verified against the #1584 prescription: commit-before-verify ordering (codex round-1 was right — a throwing verify() or a Tower restart inside the old window would have reopened the loop; synchronous SQLite commit closes it), zero post-write holds after the audit, the control test written-exactly-once with two real drainer ticks writing nothing, the immediate-responder simulation against a real screen mirror, and the unverified-delivery notice port so a delivered-but-unverified cron/backstop row reaches a human. CI 7/7 on 4f692a773. Ships as 3.3.3 hotfix; field verification (long --file send to a responding builder → exactly one copy) is architect-driven post-release before #1583 closes.

Issue #1515, opened 2026-08-18
Test-spawned Towers inherit real ~/.agent-farm cloud identity — suite runs deregister the user's production Tower from Codev Cloud


waleedkadous:
## Impact (observed in production, 2026-08-17/18)

Running the codev test suite while a cloud-connected Tower is up **disconnects the user's real cloud tunnel and deregisters their Tower**, silently, until a human manually reconnects. Last night's sequence, from `~/.agent-farm/tower.log`:

- `23:49:43Z` — during a builder's full-suite run: `Tunnel connect requested (via=local ua="node")` ×3 + one `Tunnel disconnect requested`, then **`Server-side deregister succeeded for tower 01KZ…`**, then the production Tower logs **`Cloud config removed or invalid, disconnecting tunnel`**.
- Tunnel stays down **6.5 hours** until the owner manually reconnects at `06:14Z`. Downstream consumers of the cloud connection (e.g. the Stream Deck integration) broke for the whole window.
- Further identical bursts at `03:02–03:13Z` during another builder's suite runs (config already gone by then, so no further state transitions).

This is a recurrence of the "disconnect saga" (#1370/#1372 era) — and the `via=` attribution logging shipped for it in 3.3.0 is what finally named the actor: **our own test suite**.

## Root cause

`packages/codev/src/agent-farm/__tests__/helpers/tower-test-utils.ts` spawns a real `tower-server` child with:

```ts
env: { ...process.env, NODE_ENV: 'test', AF_TEST_DB: `test-${actualPort}.db`, SHELLPER_SOCKET_DIR: socketDir, ... }
```

The DB and shellper sockets are isolated; **`~/.agent-farm` is not**. The spawned Tower therefore:

1. **Reads the production cloud config** — real tower ID and token. A test exercising `/api/tunnel/connect`/`disconnect` acts *as the user's registered Tower*: disconnect performs a **server-side deregister of the production tower ID** and **deletes the production cloud config file**.
2. The real Tower's config watcher notices the deletion and drops its live tunnel.
3. **Writes to the shared `~/.agent-farm/tower.log`**, interleaving test-tower lines with production lines — which is why the earlier forensics (this saga's "unidentified actor") never converged: the evidence was two processes writing one file.

## Prescribed fix (BUGFIX-shaped)

[...]

Issue #985, opened 2026-06-04
consult -m claude bills metered Opus API instead of the Claude subscription (ANTHROPIC_API_KEY shadows CLAUDE_CODE_OAUTH_TOKEN)


waleedkadous:
**BUGFIX.** `consult -m claude` runs CMAP/review traffic on the **metered Opus API** instead of the Claude **subscription**, because it forwards `ANTHROPIC_API_KEY` into the Agent SDK env, which shadows `CLAUDE_CODE_OAUTH_TOKEN` in the SDK's auth priority. Reported by a downstream Codev adopter (~$150/day on their org key on a heavy dev day).

## Root cause (verified in source on current main)
`packages/codev/src/commands/consult/index.ts`, the `claude` path:
- Lines ~504-509 build the `env` passed to `claudeQuery` by copying **all of `process.env`** — which includes `ANTHROPIC_API_KEY` (and `ANTHROPIC_AUTH_TOKEN`) when present in the architect's environment.
- The Agent SDK prioritizes `ANTHROPIC_API_KEY` over `CLAUDE_CODE_OAUTH_TOKEN`, so the subprocess authenticates as a **metered API key** rather than the subscription/OAuth token.
- the reporter verified: with `ANTHROPIC_API_KEY` + `CLAUDECODE` unset, an OAuth subscription token authenticates the Agent SDK fine.
- Adjacent (NOT the focus): line ~519 hardcodes `model: 'claude-opus-4-6'`. Note it, but keep this fix to the auth issue unless trivial.

## Prescribed fix (scoped to consult's claude subprocess ONLY)
- When `CLAUDE_CODE_OAUTH_TOKEN` is set, **delete `ANTHROPIC_API_KEY` and `ANTHROPIC_AUTH_TOKEN` from the local `env` object** built at ~504 before passing it to `claudeQuery` — **NOT** from global `process.env`. Other callers that need the API key (persona, dev:local) are unaffected because the deletion is scoped to consult's subprocess env.
- When `CLAUDE_CODE_OAUTH_TOKEN` is NOT set, leave behavior unchanged (fall back to the API key) — so CI / key-only environments still work.
- Regression test: with `CLAUDE_CODE_OAUTH_TOKEN` set, the env passed to `claudeQuery` must NOT contain `ANTHROPIC_API_KEY`/`ANTHROPIC_AUTH_TOKEN`; with it unset, the key is preserved.

## Effect & caveat
[...]

PR #1626, opened 2026-09-05
Fix #1219: sanitize inherited CLAUDE_CODE_* env at every Tower spawn

### LINKED ISSUE #1219: tower: sanitize inherited CLAUDE_CODE_* env — restarting Tower from a Claude Code session silently disables transcript saving (and resume) for every spawned agent
# tower: sanitize inherited CLAUDE_CODE_* env — restarting Tower from a Claude Code session silently disables transcript saving (and resume) for every spawned agent

## Mechanism (verified end-to-end on a live install)

Claude Code plants `CLAUDE_CODE_CHILD_SESSION` (and `CLAUDE_CODE_SESSION_ID`, `_ENTRYPOINT`, `_EXECPATH`) into the environment of every subprocess it spawns, so a nested `claude` can detect it was launched from inside another session and disable transcript saving.

Tower is a daemon — but it inherits the environment of whoever starts it. When Tower is started/restarted from a shell inside a Claude Code session (an extremely common operation: `afx tower start`, `pnpm -w run local-install`, crash recovery driven by an architect), the marker bakes into Tower's env and cascades:

```
claude session → Bash tool → afx tower start → Tower → shellper → agent claude
```

Every architect/builder claude session Tower spawns then believes it is a nested child session and shows:

> ⚠ Transcript saving is off — inherited CLAUDE_CODE_CHILD_SESSION marker · restart with CLAUDE_CODE_FORCE_SESSION_P…

Verified with `ps eww`: the running Tower daemon carried the marker, and multiple shellper-spawned agent claude processes carried it (inherited at spawn). After restarting Tower via `env -u CLAUDE_CODE_CHILD_SESSION -u CLAUDE_CODE_SESSION_ID -u CLAUDE_CODE_ENTRYPOINT -u CLAUDE_CODE_EXECPATH afx tower start`, Tower's env is clean and new spawns are unaffected.

## Why this is worse than a warning banner

Sessions with transcript saving disabled **cannot be resumed**. Tower's whole crash-recovery story (#1145, #1149, session_id-based resume) assumes agent sessions persist transcripts. A Tower that was ever restarted from inside a Claude session quietly produces agents that are unrecoverable after the next crash — the failure only manifests at recovery time, when it's too late. The leak also persists across Tower restarts *via shellpers*: daemonized shellper processes outlive Tower and keep their spawn-time env, so agents they restart keep the marker even after Tower itself is cleaned.

## Proposed fix

1. **Tower sanitizes at spawn**: strip `CLAUDE_CODE_*` (or at minimum `CLAUDE_CODE_CHILD_SESSION`) from the env it passes to shellper/agent processes. This is the robust fix — it protects regardless of how Tower was started, including the shellper-outlives-Tower path for newly restarted claude processes.
2. **`afx tower start` scrubs its own env** before daemonizing, and (optionally) prints a note when it detects it was launched from inside a Claude session.
3. **`codev doctor` check** (cheap): warn if the running Tower's env carries `CLAUDE_CODE_CHILD_SESSION` — detectable via `ps eww` on the daemon PID.

## Notes

- Any agent terminals spawned while a contaminated Tower was up have the marker baked in until their claude process restarts; the fix is forward-looking.
- The transcript-off warning banner appears to be recent Claude Code CLI behavior; the leak itself may long predate its visibility — which is exactly why the doctor check (3) is worth having.

Area: `area/tower`.

waleedkadous:
Fixes #1219

## Summary

A Tower started from inside a Claude Code session inherits that session's identity markers and hands them to every agent it spawns. Each agent then believes it is a nested child session, disables transcript saving, and becomes **unresumable** — silently, until crash recovery fails. This routes every Tower-descendant spawn through one sanitizer that strips per-session identity, and adds a `codev doctor` check for daemons that are already contaminated.

## Root Cause

Claude Code plants session markers — `CLAUDE_CODE_CHILD_SESSION`, `CLAUDE_CODE_SESSION_ID`, `CLAUDE_CODE_ENTRYPOINT`, `CLAUDE_CODE_EXECPATH`, and friends — into every subprocess it creates, so a nested `claude` can tell it was launched from inside another session. Tower is a daemon, but it inherits the environment of whoever started it, and starting Tower from inside a Claude Code session is routine (`afx tower start`, `pnpm -w run local-install`, an architect recovering from a crash). The markers cascade:

```
claude session → Bash → afx tower start → Tower → shellper → agent claude
```

Confirmed on a live install rather than assumed: `ps eww` on two running Tower daemons showed the full marker set, inherited from the shell that started them.

Three code shapes, none of which handled `CLAUDE_CODE_*`:

1. `towerStart()` daemonized with `env: process.env` (`agent-farm/commands/tower.ts`).
2. Six Tower spawn sites built `{ ...process.env }` and deleted **only** `CLAUDECODE` — `tower-instances.ts` ×2, `tower-terminals.ts` ×2, `tower-routes.ts` ×2.
3. `createPtySession()` sends no `env`, so builder terminals took the `env || process.env` branch in `handleTerminalCreate` and received Tower's whole environment.

`shellper-main` passes its config `env` straight to node-pty, which replaces the child environment wholesale — so the env Tower hands to `SessionManager.createSession` *is* the agent's env. That makes (2)+(3) load-bearing and (1) the upstream source.

## Fix

New `packages/codev/src/lib/agent-env.ts` with a single `sanitizeAgentEnv()`: **strip what is per-session identity, keep everything else.**

Identity is a small, stable, nameable set — four prefix families (`CLAUDE_CODE_SESSION_`, `CLAUDE_CODE_REMOTE_SESSION_`, `CLAUDE_CODE_BRIDGE_`, `CLAUDE_CODE_MESSAGING_`) plus ten individual names, including the `CHILD_SESSION` / `ENTRYPOINT` / `EXECPATH` nesting triple this issue is about. Configuration is *not* a nameable set, so it is not enumerated.

Every Tower-descendant spawn routes through it — the daemon spawn, both architect-launch sites, both architect-reconnect sites, terminal-create and shell-create (plus their non-persistent fallbacks), the shellper daemon spawn, and cron task execution. The six hand-rolled `delete cleanEnv['CLAUDECODE']` copies collapse into the one helper. `afx tower start` logs what it scrubbed, and `codev doctor` grows a **Tower Environment** section that reads the running daemon's env via `ps eww` — a daemon started before this fix stays contaminated until restarted, and that was previously invisible.

A supporting commit moves `getProcessesOnPort` to a leaf `agent-farm/utils/port.ts` (re-exported, no consumer moves): importing it from the command module dragged `utils/shell.ts`'s module-scope `promisify(exec)` into the doctor tests, whose `node:child_process` mock has no `exec` export.

### Why not deny the whole `CLAUDE_CODE_*` namespace

The first version did exactly that, with a short allowlist, on the theory that a missed *marker* is worse than a missed *config var*. The codex CMAP lane rejected it and was right. Checking the shipped `claude` binary (2.1.261) settled it: it reads **594** distinct `CLAUDE_CODE_*` variables, overwhelmingly configuration. The allowlist would have silently dropped `USE_FOUNDRY`, `USE_MANTLE`, `USE_ANTHROPIC_AWS`, `USE_GATEWAY`, `SKIP_FOUNDRY_AUTH`, `SKIP_MANTLE_AUTH`, `OAUTH_REFRESH_TOKEN`, `OAUTH_SCOPES`, `OAUTH_CLIENT_ID`, `API_BASE_URL`, `PROXY_URL`, `HTTPS_PROXY`, `CLIENT_CERT`, `MANAGED_SETTINGS_PATH` and more.

The original reasoning was wrong about the failure mode: dropping a config var is not harmless. It points an agent at the wrong provider, or strips its subscription credential and reroutes CMAP to the metered API — the #985 scar. No hand-maintained allowlist survives 594 variables.

### `CLAUDE_PID` disposition

Checked against the shipped binary the same way as the 594: **`CLAUDE_PID` is stripped.**

It is not read for nested-session detection — its only reader is the `pkill` guard in the Bash tool's shell prelude, which refuses a pattern matching the Claude CLI's own pid. But it is per-session identity by construction, planted in the same object literal as the marker this issue is about:

```js
function qOe(e){let t={CLAUDECODE:"1", CLAUDE_CODE_SESSION_ID:e.sessionId,
                       CLAUDE_CODE_CHILD_SESSION:"1", CLAUDE_PID:String(process.pid)};
  if(e.source==="agent") t.AI_AGENT=…;
  if(e.effortLevel!==void 0) t.CLAUDE_EFFORT=e.effortLevel; …}
```

`String(process.pid)` of one specific process is identity by any reading, and an inherited stale value points that `pkill` guard at a dead or unrelated pid. So it is in the denylist despite sitting outside the `CLAUDE_CODE_` namespace.

`CLAUDE_EFFORT` is planted by that *same function* and is deliberately **not** stripped — which is the useful part of the check: "Claude Code sets it" is not the test. It carries a setting, not an identity, so it is configuration and passes through.

## Test Plan

- [x] Regression test added — `bugfix-1219-claude-env-leak.test.ts`, 31 cases. **9 assertions fail** with the spawn-site changes reverted to the pre-fix commit and pass with them; reverting just *one* of `tower-routes.ts`'s four sites fails 2.
- [x] Build passes
- [x] All tests pass — 5739 passing, 48 skipped

Covered by the regression test:

- Every marker observed on a real contaminated Tower is stripped.
- Whole session families are stripped, not just the names seen so far.
- A `MUST_SURVIVE` fixture of 24 provider / auth / routing / policy variables — each verified present in the shipped binary, including the Foundry, Mantle, Anthropic-AWS and refresh-token cases — is preserved intact.
- An unrecognised `CLAUDE_CODE_*` variable is **kept**, not dropped.
- Every listed spawn site calls the sanitizer **the expected number of times**, and carries none of the raw-environment shapes a reverted or copy-pasted site would take.
- `codev doctor` warns when Tower is clean but shellpers it spawned earlier still carry markers, and never claims a plain Tower restart is sufficient on its own.

Verified beyond the suite:

- **Doctor check against live processes**: a real contaminated Tower → `warn` naming all 7 markers; a dead pid → `skipped`. On this machine the session check found the exact case it exists for — 71 shellpers running, 5 still carrying markers, Tower itself clean, which the daemon-only version called `ok`.
- **End to end**: a Tower started from a shell carrying 7 markers logged `Scrubbed inherited Claude Code session markers: …` and came up with a clean environment. **That run is the incident described below** — it is reported there rather than claimed here as a clean result. Field verification will be redone after merge via `local-install` on `main`.

## Incident during verification

While verifying this fix I took down every terminal on the development machine. Recording it here because it is the kind of failure this PR's own subject matter — processes quietly attaching to state they should not touch — is about.

I started a test Tower on port 14733 to confirm the daemon scrub end to end, and exported `AGENT_FARM_DIR` to isolate it. **That is not the override.** The real one is `CODEV_AGENT_FARM_DIR` (`packages/core/src/constants.ts:18`, added by #1515). So my test Tower opened the **production** `~/.agent-farm/global.db` and appended to the production `tower.log`.

Its startup reconcile then found the 56 live terminal rows and connected to every production shellper socket. A shellper holds one client, so the real Tower on 4100 lost all of them; the two Towers fought over reconnects, and the reconcile's stale-row path **deleted 53 of the 56 rows**. When I killed the test Tower those sessions were orphaned: 60 unregistered shellpers, and every architect terminal across 12 workspaces gone from the dashboard — including the architect's own. The architect restored the rows by hand from the checkpointed DB file and restarted Tower.

The log corroborates it: 56 `reconcile-adopt` lines at `18:26:59.4xx`, then 54 `Session … removed but shellper pid=… is alive; preserving socket` lines by `18:27:02.5xx`.

My gate message described this as having "briefly contended with the live Tower for the cloud tunnel. It recovered." That was wrong — I reported the tunnel flapping I had noticed and not the session loss I had not looked for, which made a fleet-wide outage read as a transient blip.

**#1629** is filed for the systemic defects, and they are the real lesson: a second Tower must not be able to open a live `global.db` and adopt rows owned by a running Tower (owner lock, or refuse to adopt), and an isolation env var whose misspelling silently means "use production" is a trap rather than a safeguard. My mistake was the trigger; neither of those should be reachable by one wrong variable name.

No Tower has been started from this lane since. Field verification of the daemon scrub is deferred to `local-install` on `main` after merge.

### Known follow-ups (filed, not in this PR)

- **#1627** — `afx architect` spawns `claude` unsanitized. Pre-existing and outside Tower (it never stripped `CLAUDECODE` either), so an architect started from inside a Claude session still nests.
- **#1628** — `codev doctor` inspects only `DEFAULT_TOWER_PORT`, and the session scan costs ~2 `ps` spawns per shellper (~142 on a 71-shellper machine); both fixable, neither worth widening this PR for.

One limit worth stating rather than filing: `codev doctor` classifies with the same list the sanitizer uses, so it **cannot** backstop a marker Claude Code adds that this list has not caught up with. Claude Code's own "transcript saving is off" banner remains the only signal for that case — it is how this issue was found.

https://claude.ai/code/session_01ApsmgthG7VWgMM2JaizWD6

### DIFF
--- a/codev/projects/bugfix-1219-tower-sanitize-inherited-claud/status.yaml
+++ b/codev/projects/bugfix-1219-tower-sanitize-inherited-claud/status.yaml
@@ -7,8 +7,10 @@ current_plan_phase: null
 gates:
   pr:
     status: pending
+    requested_at: '2026-09-05T19:00:58.823Z'
 iteration: 1
 build_complete: false
 history: []
 started_at: '2026-09-05T18:12:23.945Z'
-updated_at: '2026-09-05T18:28:43.887Z'
+updated_at: '2026-09-05T19:00:58.824Z'
+pr_ready_for_human: true
--- a/codev/projects/bugfix-1219-tower-sanitize-inherited-claud/status.yaml
+++ b/codev/projects/bugfix-1219-tower-sanitize-inherited-claud/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1219
 title: tower-sanitize-inherited-claud
 protocol: bugfix
-phase: fix
+phase: pr
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-09-05T18:12:23.945Z'
-updated_at: '2026-09-05T18:16:27.797Z'
+updated_at: '2026-09-05T18:28:43.887Z'
--- a/codev/projects/bugfix-1219-tower-sanitize-inherited-claud/status.yaml
+++ b/codev/projects/bugfix-1219-tower-sanitize-inherited-claud/status.yaml
@@ -1,7 +1,7 @@
 id: bugfix-1219
 title: tower-sanitize-inherited-claud
 protocol: bugfix
-phase: investigate
+phase: fix
 plan_phases: []
 current_plan_phase: null
 gates:
@@ -11,4 +11,4 @@ iteration: 1
 build_complete: false
 history: []
 started_at: '2026-09-05T18:12:23.945Z'
-updated_at: '2026-09-05T18:12:23.946Z'
+updated_at: '2026-09-05T18:16:27.797Z'
--- a/codev/projects/bugfix-1219-tower-sanitize-inherited-claud/status.yaml
+++ b/codev/projects/bugfix-1219-tower-sanitize-inherited-claud/status.yaml
@@ -1,0 +1,14 @@
+id: bugfix-1219
+title: tower-sanitize-inherited-claud
+protocol: bugfix
+phase: investigate
+plan_phases: []
+current_plan_phase: null
+gates:
+  pr:
+    status: pending
+iteration: 1
+build_complete: false
+history: []
+started_at: '2026-09-05T18:12:23.945Z'
+updated_at: '2026-09-05T18:12:23.946Z'
--- a/codev/state/bugfix-1219_thread.md
+++ b/codev/state/bugfix-1219_thread.md
@@ -128,3 +128,69 @@ shape repeated six times, so that shape is what is pinned. The *behaviour* —
 what `sanitizeAgentEnv` strips and keeps — is pinned properly by unit tests.
 
 <signal>PHASE_COMPLETE</signal>
+
+## PR + CMAP
+
+PR #1626. Four CMAP rounds, because the first design was wrong and the review
+caught it. Verdict history:
+
+| Round | gemini | codex | claude |
+|---|---|---|---|
+| 1 | APPROVE | **REQUEST_CHANGES** | APPROVE |
+| 2 | APPROVE | **REQUEST_CHANGES** | APPROVE |
+| 3 | — | COMMENT | — |
+| 4 | APPROVE | **REQUEST_CHANGES** (scope only) | APPROVE |
+
+### What the review actually changed
+
+**codex R1 — the allowlist was wrong, and this was the big one.** I had denied
+`CLAUDE_CODE_*` wholesale with an allowlist, arguing a missed marker (silent,
+unresumable) beats a missed config var (loud, harmless). codex said the second
+half of that was false. I checked the shipped binary instead of arguing: **594**
+distinct `CLAUDE_CODE_*` variables, overwhelmingly configuration. Every name
+codex cited is real — `USE_FOUNDRY`, `USE_MANTLE`, `USE_ANTHROPIC_AWS`,
+`SKIP_FOUNDRY_AUTH`, `OAUTH_REFRESH_TOKEN`, `OAUTH_SCOPES` — plus `API_BASE_URL`,
+`PROXY_URL`, `CLIENT_CERT`, `MANAGED_SETTINGS_PATH` that I'd have dropped too.
+Dropping a provider selector or a credential is not harmless; it is the #985
+metered-billing scar. Inverted to a session-identity denylist (4 prefix families
++ 10 names).
+
+**codex R2 — the doctor check had a false negative.** `towerStop` leaves
+shellpers running by design, so after a plain restart the daemon reads clean
+while its earlier agents keep the markers. My check read only the daemon and the
+hint said "restart Tower" as if that were sufficient. Now scans shellper envs
+too, with two honest remediation strings. **This machine was already in that
+state**: 71 shellpers, 5 contaminated, Tower clean — old check said `ok`.
+
+**claude R1 — `tower-cron.ts` spawned with raw `process.env`.** A genuine missed
+site. Fixed and added to SPAWN_SITES.
+
+**claude/codex R3-4 — spawn assertions were file-granular.** One of
+tower-routes' four sites could revert with the test still green. Now pins a call
+count per file plus forbidden raw-env shapes; verified by actually reverting one
+of four and watching 2 assertions fail.
+
+**claude R4 — two false-clean bugs of mine.** `readProcessEnv` returned `{}`
+(reads as "clean") when ps shows no env; and "Tower is clean" printed with no
+Tower running. Both fixed, verified live against pid 1.
+
+### Open disagreement for the architect
+
+codex R4 is REQUEST_CHANGES on **scope only** — 321 net production LOC vs
+BUGFIX's ~300. Its arithmetic is right. The breakdown: 366 added = 174 comment +
+21 blank + 171 code, and 24 of those code lines are the `getProcessesOnPort`
+move, so ~147 lines are genuinely new. codex itself says "the implementation is
+focused; no correctness or security defect was found," and gemini R4 reads the
+same diff as "without scope creep or unnecessary abstractions."
+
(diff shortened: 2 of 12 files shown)


waleedkadous:
## Architect integration review

Merging on the human's word after reading the fix and verifying the load-bearing claims against the branch:

- All seven Tower spawn paths route through the one `sanitizeAgentEnv` helper (session-manager, tower start, cron, instances x2, routes x4, terminals x2). No spawn site left that reads `process.env` directly.
- Denylist over a prefix wipe is the right shape: a `CLAUDE_CODE_*` wipe would have dropped auth and billing config (the #985 scar). `CLAUDE_PID` added after checking the shipped binary; `CLAUDE_EFFORT` kept as config.
- Doctor check covers both the daemon and already-running shellpers, which survive a Tower restart by design. Its known limit (same list as the sanitizer, cannot catch a marker the list has not learned) is stated in the body.
- Scope objection from codex (about 320 production lines vs the bugfix guideline) accepted as architect's call: all three deliverables were prescribed in the issue.
- The incident section stays as written. It is the corrected record for #1629 and the reason the end-to-end claim came out of the test plan.

Field verification of the daemon scrub happens next via `local-install` on `main`, restarting Tower from inside a Claude session **without** a manual scrub and inspecting the daemon's environment.

Issue #1567, opened 2026-08-31
afx send: delivery-side silent truncation of long bodies (head-loss can invert meaning) — needs integrity check before Enter


amrmelsayed:
Reported by the shannon workspace's main architect (Amr-directed forward, 2026-08-31, three-part relay received complete); reviewed and filed by codev:main.

## Defect

`afx send` can silently truncate long message bodies **at delivery** (the mailbox row stores the full body; what lands on the recipient's prompt is incomplete). The sender receives `delivered` with no error; the recipient has no marker that anything is missing. Both variants observed:

- **Head-truncation**: only the tail arrives; the recipient loses all framing. Worst confirmed case: a ruling arrived INVERTED in meaning (the part that negated the instruction was in the lost head), caught only because the sender re-read their own sent text.
- **Probable tail-clip**: body ends early.

## Occurrence ledger (shannon workspace, six confirmed, three senders)

- ~2026-08-26T22:24Z, shapps architect → builder-pir-4596: two consecutive long rulings head-truncated.
- 2026-08-27 ~09:08Z, main architect → builder-air-3816: a 5-part instruction arrived as "item 5 + a fragment"; the builder's hold-and-ask discipline caught it.
- 2026-08-27, brand architect: three occurrences (receipts at shannon `codev/state/.brand.bak.md` lines 240/244/246): two clipped relays to builder-pir-4623 plus the inverted ruling to builder-pir-4625.

Every catch was behavioral (hold-and-ask, sender re-read). **No mechanical detection exists.** Raw receipts available from shannon:main on request.

## Relation to known defects (same channel, distinct integrity failures)

- #1543 tracks sender/recipient labeling in the wrapper (now 8+ occurrences per shannon, including two architect-to-architect label misroutes on 08-30). That defect corrupts attribution; this one corrupts CONTENT. Together they mean neither who-said-it nor what-was-said is currently trustworthy for long inter-agent messages.
- Sender-side clipping via shell substitution (inline backticks) is a separate, known operator-error class with the `--file` discipline as its mitigation; the ledger cases above are delivery-side (mailbox row complete, delivered text incomplete), which `--file` does not protect against.

## Investigation leads (from a source skim, not conclusions)

- The delivery path has a documented race window after the render-gate sample: `mailbox-delivery.ts:392` ("if the screen advanced since we sampled `tokenBefore`, a draft may have started"). Longer bodies take longer to write, widening any post-verification race — consistent with the length correlation.
[...]

PR #1666, opened 2026-09-09
Fix #1664: wait on the composer, not on the recipient's output

### LINKED ISSUE #1664: mailbox gate: a recipient that is producing output holds mail until its turn ends — only a human draft should hold (Claude Code queues input mid-turn)
## Owner ruling (2026-09-09)

> If the AI is producing output it shouldn't wait. It should only wait if the human is typing, because of the way queueing works.

Claude Code (and codex) accept input during a turn and queue it for the next one. Writing a message onto an **empty** composer while the agent is streaming is therefore safe and is the intended behaviour. The gate must hold only when a human is at the line: a draft on the composer (`busy:user-text`) or a keystroke in flight (`busy:recent-input`, #1473).

## What happens today

Two mechanisms in `packages/codev/src/agent-farm/servers/mailbox-delivery.ts` hold on the recipient's **output**, independent of the human:

1. **Settle-before-write** (`SETTLE_BEFORE_WRITE_MS = 250`, from #1573): `if (!settled(ports, session)) return hold('busy')` — the write is refused unless the session has been free of *any* output byte for 250 ms. A streaming turn with a spinner repaint rarely goes quiet that long, so the row waits for the turn to end, re-tried on the 1.5 s backstop.
2. **Change-token re-hold**: `ringToken` folds `bytesWritten` in, so any repaint between the token sample and the classify re-holds as a detail-less `busy` (both in the gate and in the in-lock precheck).

Both produce the plain `busy` with no detail — the one verdict that is invisible to the diagnostic axis, so nothing says "waiting on the *agent*, not a human".

**Cost, measured from global.db, last 24 h, all workspaces:** 215 deliveries waited more than 2 s; mean wait 54 s, max **561 s**; 59 crossed the escalation age and lit the badge. In the adopter workspace the escalated mean was 139 s. Every one of those rows eventually delivered on its own when the turn ended — i.e. the recipient was never unsafe to write to, just busy talking.

## Why the settle exists, and what must survive

#1573 added the settle because of #1521: a delivery that landed while the composer was **being redrawn at the end of a turn** lost its leading bytes (`[USER via VS Code]` → `ER via VS Code`). That hazard is real, but it is about the *composer region* being repainted, not about output anywhere on the screen. The spinner and the transcript stream above the composer do not eat input bytes; the TUI reads stdin regardless and queues it.

## Fix (prescribed, BUGFIX)

Replace the whole-screen output settle with a **composer-region stability** check, and keep every human-side hold exactly as it is:

- The classifier already isolates the composer region (marker row → lower rule). Hash the rendered rows of that region. Deliver when the region is classified empty **and** its hash is unchanged across two samples ≥ `SETTLE_BEFORE_WRITE_MS` apart. Output that changes only rows *outside* the region (spinner, streaming transcript) must not hold.
- The change-token re-check keeps its **input** half (`inputSeq`) unchanged. For the output half, compare the composer-region hash rather than `bytesWritten`; a repaint that leaves the region byte-identical is not a reason to re-hold.
- Do the same in the in-lock precheck.
- Give the remaining output-side hold a detail (`busy:composer-redraw` or similar) so it is visible on `afx inbox` and the send response, instead of the null-detail `busy`; add it to the shared vocabulary (`MailboxGateDetail`, `api.ts`, `sse.ts`, `hold-verdict.ts`, `inbox.ts describeDetail`) as #1473 did for `recent-input`. It is self-clearing, so it stays out of `isUnverifiableVerdict`.

### Acceptance (must be measured, not asserted)

1. The #1567 repro harness (`bugfix-1567` evidence) re-run against a recipient that is **streaming a long reply with the spinner active**: 20/20 deliveries land intact, verified by echo, and are submitted as one message when the turn ends. Same harness against a recipient with a **draft** on the line: 20/20 held as `busy:user-text`.
2. The #1521 turn-end case: a delivery timed to land during the composer redraw is held, then lands intact. 0/20 head-loss.
3. Before/after from global.db on the maintainer's machine over a comparable window: the count of rows that waited > 2 s with a null-detail `busy` drops to ~0; `user-text`/`recent-input` holds unchanged.
4. Unit tests pin: streaming output outside the composer region does not hold; a region repaint does; the new detail flows through every surface.

Refs: Spec 1313 (mailbox-first; "delivery happens on output quiescence" was a delivery *moment*, not a precondition), #1573/#1521/#1564 (the settle's origin), #1473 (input-side settle — unchanged), #1567 (bracketed paste; harness).

waleedkadous:
## Summary

The mailbox gate held mail whenever the recipient was *producing output* — so a message to a working agent queued behind the entire turn (24 h of `global.db`: 227 deliveries waited past 2 s, mean 54 s, max **561 s**) even though its composer was empty and safe to write to the whole time. The gate now waits on the **composer region**, not on the screen.

Fixes #1664

## Root Cause

`mailbox-delivery.ts` proved "the composer has stopped being painted" with two whole-screen signals:

1. `settled()` — required `now - lastDataAt >= 250 ms`, i.e. **no output byte anywhere** on the screen.
2. `ringToken()` folded `bytesWritten`, so any repaint re-held at the post-classify re-validation and in the in-lock precheck.

Both returned `hold('busy')` with a **null detail** — the one verdict invisible to #1482's diagnostic axis, so nothing said "waiting on the agent, not a human".

Measured by replaying a real streaming claude PTY capture at its true repaint boundaries (`ESC[?2026l`):

| signal | result |
|---|---|
| repaints while streaming | median **8/sec**, mean interval **122 ms**; only **3 of 728** seconds offered a 250 ms gap |
| whole-screen hash changes, 1500 consecutive frames | **1242** |
| composer-region hash changes, same frames | **4** |
| region-unchanged runs, wall clock | 107 s, 60 s, 304 s, 81 s, 149 s |

So the precondition was unobtainable for the length of a turn, while the thing it was a proxy for sat byte-identical for minutes.

## Fix

- **`render-gate.ts`** — `locateComposerRegion` extracted so the classifier and the new fingerprint can never disagree about which rows are the composer, and `composerRegionFingerprint()` added: the region's rendered text, its row span, the cursor, and the geometry. `null` when no region can be located.
- **`session-screen.ts`** — `peek()`, the synchronous counterpart of `read()`, because the in-lock precheck cannot await.
- **`mailbox-delivery.ts`** — a required `composerFingerprint` port and a per-session record of the last observed region. A write requires the region unchanged across two samples ≥ `SETTLE_BEFORE_WRITE_MS` apart, **or** the whole screen quiet (kept: strictly stronger, and it lets an idle recipient deliver on its first pass). `MAX_COMPOSER_SAMPLE_GAP_MS` restarts that clock across an observation gap, so a stale sample can never authorise a write onto a composer repainted while nobody was looking. The `bytesWritten` half of the TOCTOU token is gone; **the `inputSeq` half is untouched — a human keystroke still holds.** Same rule in the in-lock precheck.
- **New detail `composer-redraw`** — self-clearing, through `MailboxGateDetail`, `api.ts`, `sse.ts`, `hold-verdict.ts`, `afx inbox`, the dashboard badge and the VS Code toast, with a `retryAfterMs` sharing #1473's one-shot re-drain timer (the input-hold streak diagnostic is gated to `recent-input`, so a repainting composer is never described as a typist).

#1521's hazard survives: a composer mid-redraw *does* move the fingerprint, and the in-lock precheck re-checks it at the write instant.

117 non-comment lines of production change.

## Test Plan

- [x] **Regression test added** — `bugfix-1664-composer-stability-gate.test.ts`, 19 cases. Verified it **fails without the fix**: with the old gating temporarily restored, 4 cases fail including "THE REGRESSION: delivers to an agent that is streaming".
- [x] **Build passes** — `pnpm build` clean.
- [x] **All tests pass** — 6030 passed, 48 skipped.
- [x] **Measured acceptance against a real `claude 2.1.266`**, driven through the production `deliverAgentMail` (real sqlite mailbox, live gate bindings, `submitMessagePaced` — the harness re-implements no gate). 20 trials each; evidence in `codev/evidence/1664-streaming-delivery/`:

| run | gate | delivered | head-loss | median | max | waits >2 s |
|---|---|---|---|---|---|---|
| streaming | current | **20/20 intact** | **0/20** | 394 ms | 396 ms | **0/20** |
| streaming (*same scenario*) | **retired** whole-screen settle | 20/20 intact | 0/20 | 2505 ms | 6432 ms | **11/20** |
| human draft on the line | current | **0/20 — 20/20 held `busy:user-text`** | — | — | — | — |
| across a turn end (#1521's window) | current | **20/20 intact** | **0/20** | 393 ms | 404 ms | **0/20** |

**All 20** streaming deliveries were written at an instant where `now - lastDataAt < 250 ms` — writes the retired gate could not have made. The draft run had a turn concurrently running in all 20 trials: **streaming does not hold mail; a person does.** The streaming scenario was re-measured after every CMAP round that changed delivery semantics; across all of them plus the turn-end run — **100 delivering trials — head-loss is 0/100.**

### Notes for the reviewer

- **Acceptance criterion 3** (before/after over a comparable window of the maintainer's own `global.db`) needs the fix actually running in Tower, so it belongs to the post-merge verify phase. The controlled same-TUI comparison above is what can be measured pre-merge; the "before" field numbers are recorded in the thread.
- The legacy comparison **understates** the field cost: it uses `--model haiku`, whose turns are short.
- Found and closed mid-flight: the first implementation bounded nothing on the age of the previous composer observation, so a sample taken minutes earlier could authorise an immediate write onto a just-repainted composer — #1521's hazard, reopened. Caught because consecutive harness trials delivered on their very first pass; pinned by a test.
- **`claude 2.1.266` no longer renders `esc to interrupt`**, which the #1567 harness matches to detect a running turn — on this version it reports a working agent as idle. Not fixed here (out of scope), but recorded in the evidence README.

## CMAP: 4 rounds, 3 real defects found

Final verdicts — **gemini APPROVE · codex COMMENT · claude APPROVE** (no blockers). Three rounds of `REQUEST_CHANGES` caught three genuine defects, all of which I had missed:

1. **A delivery's own write moved the composer, but the stability sample survived it** (codex r1). A second delivery 250–2000 ms later could reuse the pre-write observation and write into the redraw following our own submit — #1521's window, reopened by our own hand, and inside the gap bound so nothing else could catch it.
2. **A `!verdict.clean` pass is evidence of movement, and it was skipped** (claude r1), so a later clean pass with a coincidentally identical fingerprint claimed stability across the interval we watched the composer change.
3. **`peek()` could hand back a grid missing bytes that were fed but not yet parsed** (codex r2), so a redraw could hide in the parse queue and the in-lock check would compare equal. I had *documented* this as an accepted one-parse-turn residual; it was cheaper to close. `feed()` now numbers each chunk and `peek()` refuses to answer while the grid is stale.
4. **The owner starvation notice told humans to `afx interrupt` a self-clearing hold** (codex r3, claude r3). Routing on `if (info.detail || …)` put every detail but `user-text` into the "gate CANNOT VERIFY / will never deliver / run `afx interrupt`" arm. For `composer-redraw` that kills the turn of an agent that is merely working — and **the same branch already misfires for `recent-input` on `main` today**, telling an owner to interrupt someone who is simply typing. That is precisely the mistake the function's own comment says aggravated #1583. Fixed for both by routing on the shared `isUnverifiableVerdict` predicate; the `recent-input` half is a deliberate one-detail widening beyond this issue, called out here rather than left to ship.

Every fix is pinned by a test verified to fail with that fix reverted.

### Follow-ups deliberately not done here

- A `null` fingerprint reports the self-clearing `composer-redraw` rather than the classifier-stuck family, so a *systematic* classify/peek divergence would hold without classifier-stuck telemetry (age-based escalation still fires). Distinguishing them needs a richer port return type for a case that requires classify and peek to disagree persistently.
- The post-merge verify should include a **codex** recipient — the live acceptance drove claude only, though codex's and agy's composer shapes now have fixture-level fingerprint coverage.
- `MAX_COMPOSER_SAMPLE_GAP_MS` (2 s) sits only 500 ms above the 1.5 s backstop — latency-only, self-correcting via the retry timer.


waleedkadous:
## Architect integration review — 3-way

**Verdicts: gemini APPROVE (HIGH) · claude COMMENT (HIGH) · codex REQUEST_CHANGES (HIGH).** CI 7/7 green.

This is excellent work — the measurement that opened the investigation (8 repaints/sec, 3 quiet seconds in 728, region byte-identical for minutes) is the kind of evidence that ends an argument, and `locateComposerRegion` is the right abstraction: classifier and fingerprint structurally cannot disagree about which rows are the composer. The owner-notice routing through `isUnverifiableVerdict` fixes a wrong remedy that was already shipping. I read the gate block and the precheck on the branch; the human-side holds are untouched.

Three things before merge, all verified against the branch:

1. **The premise "the recipient queues mid-turn input" is now a correctness premise, and it is measured only on claude** (codex + claude, independently). A TUI that no-ops Enter mid-turn would leave the body on its composer, verify as delivered (the header needle is present), and block every later message as `user-text` — worse than the latency we are removing. `AGY_PROFILE` takes the generic path today. Please add a per-profile capability (say `queuesInputMidTurn`) on `GateProfile`: `true` for claude (measured), `true` for codex **once your harness has run the streaming scenario against a live codex** (make that a gating item, and commit the evidence beside the claude runs), and `false` for agy, which keeps the whole-screen settle until someone measures it. That keeps the owner's ruling exactly where it is proven and fails safe where it is not.
2. **Fingerprint vs. classifier attribute gap in the in-lock precheck** (codex). The fingerprint deliberately omits SGR attributes and relies on the classifier re-answering emptiness — but the synchronous precheck compares fingerprints only and does not reclassify, so a dim→normal flip of identical characters between classify and write would pass on a stale clean verdict. Implausible in practice, cheap to close: fold a per-cell attribute digest of the region (dim / inverse / fg-palette — the three the classifier reads) into the fingerprint, and pin it with a placeholder→user-text transition test between classify and precheck.
3. **One comment overclaims** (claude): around `mailbox-delivery.ts:1040` the "turn ends during the paced write" residual is described as detected because the echo watch reports `verified: false`. `watchEchoOnScreen` counts the header needle or paste placard only, so a mid/tail-body split verifies TRUE. Correct the comment and name the gap as a follow-up under #1578 — prevention is not requested.

Two cheap non-blocking asks while you are in there: the `afx send` bullet in `arch-critical.md` is now a 1,029-character line — the hot-tier cap counts lines and facts, not characters, so this is invisible to the test that exists to keep the always-on tier affordable; demote the measurement figures to `arch.md` (which already carries them). And "Only a HUMAN at the line holds mail" will be quoted as the rule — a five-word qualifier naming the self-clearing and unverifiable classes keeps it honest.

Acceptance item 3 (before/after from the maintainer's global.db) is correctly deferred to post-merge verification — agreed.

